PR #3, opened 2021-02-04
PF-338 Add notebook create command

wchamber:
This works in a fresh dev WM created workspace.

TODO: Consider creating a NotebooksManager.


src/main/java/bio/terra/cli/command/notebook/Create.java
```diff
@@ -0,0 +1,110 @@
...
+      names = "machine-type",
+      defaultValue = "n1-standard-4",
+      description = "The Compute Engine machine type of this instance.")
+  private String machineType;
+
+  @CommandLine.Option(
+      names = "vm-image-project",
+      defaultValue = "deeplearning-platform-release",
+      description = "The ID of the Google Cloud project that this VM image belongs to.")
+  private String vmImageProject;
```
marikomedlock, line 42: How is this different than the workspace project id -- is it just where the GCR repository lives?

src/main/java/bio/terra/cli/command/notebook/Create.java
```diff
@@ -0,0 +1,110 @@
...
+    // each zone.
+    envVars.put(
+        "SUBNET",
+        "projects/"
+            + projectId
+            + "/regions/"
+            + zone
+            + "/subnetworks/subnetwork");
+
+    // TODO: Is there anyway to stream back the output?
```
marikomedlock, line 99: Not yet, but I have a ticket in this sprint to address this: https://broadworkbench.atlassian.net/browse/PF-424

src/main/java/bio/terra/cli/command/notebook/Create.java
```diff
@@ -0,0 +1,110 @@
...
+
+  @Override
+  public Integer call() {
+    GlobalContext globalContext = GlobalContext.readFromFile();
+    WorkspaceContext workspaceContext = WorkspaceContext.readFromFile();
+
+    AuthenticationManager authenticationManager =
+        new AuthenticationManager(globalContext, workspaceContext);
+    authenticationManager.loginTerraUser();
+    TerraUser user = globalContext.requireCurrentTerraUser();
```
marikomedlock, line 61: I think it would be good move the bulk of the logic here (~ lines 61-103) into a new class (Notebooks or NotebooksManager maybe?) under the app package. I may be doing some reorganization in that package soon, but regardless I think it's it good to keep the command classes high-level and mostly focused on user input/output.

src/main/java/bio/terra/cli/command/notebook/Create.java
wchamber: The project where the GCE image lives. I think but am not sure if it's in GCR. The defaults are the AI Platform folks default images. https://cloud.google.com/ai-platform/notebooks/docs/reference/rest/v1/VmImage

src/main/java/bio/terra/cli/command/notebook/Create.java
wchamber: I think I generally agree, but want to push back and figure out where the boundary is. This command is close to a pass-through to a gcloud command. There's a tiny bit of formatting logic (though we could push that back up to the user), but otherwise just grabbing the terra defaults. With another class level, we'd copy all the inputs down a level, but what are we gaining from the abstraction? With something like a remote API, I see it cause there's stuff about setting up the client and formatting request/response. Is it still worthwhile for a delegated CLI call? (I think I'm ok with the answer being, yes, let's do it to have a clear distinction between CLI I/O & execution.)
I think some terminology could also help here. `NotebooksManager` doesn't tell me much. I think we're trying to distinguish between the CLI input/output layer that are these commands and the execution layer.

src/main/java/bio/terra/cli/command/notebooks/Create.java
```diff
@@ -0,0 +1,108 @@
...
+import picocli.CommandLine;
+
+/** This class corresponds to the third-level "terra notebook create" command. */
+@CommandLine.Command(
+    name = "create",
+    description = "Create a new AI Notebook instance within your workspace.")
+public class Create implements Callable<Integer> {
+
+  @CommandLine.Parameters(
+      index = "0",
```
gjuggler, line 20: I got a little caught up by this when testing locally, as it was easy to miss the <instanceName> entry in the CLI help screen, since it gets tacked on after all the optional arguments:

Screen Shot 2021-02-08 at 10 48 36 AM

Form a user perspective, I don't think there's a super compelling reason to want to name my own notebook every time. I'd be happy if Terra chose a reasonable name (maybe a sanitized version of my email address?) by default, but let me override.

Easy follow-up PR if you agree with that direction.

gjuggler (approved):
Looks good! Tested locally and it all worked as expected.

src/main/java/bio/terra/cli/command/notebooks/Create.java
```diff
@@ -0,0 +1,108 @@
...
+            + "--machine-type=$MACHINE_TYPE "
+            + "--vm-image-project=$VM_IMAGE_PROJECT "
+            + "--vm-image-family=$VM_IMAGE_FAMILY "
+            + "--network=$NETWORK "
+            + "--subnet=$SUBNET "
+            + "--metadata=^:^enable-oslogin=TRUE:installed-extensions=jupyterlab_bigquery-latest.tar.gz,jupyterlab_gcsfilebrowser-latest.tar.gz,jupyterlab_gcpscheduler-latest.tar.gz";
+    Map<String, String> envVars = new HashMap<>();
+    envVars.put("INSTANCE_NAME", instanceName);
+    envVars.put("LOCATION", location);
+    envVars.put("USER_EMAIL", user.terraUserName);
```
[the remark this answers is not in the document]
marikomedlock: That field maps to the user email field in SAM, so seems reasonable to change this field in the CLI.TerraUser class also. I'll either make a new ticket or just address this in my current branch.

src/main/java/bio/terra/cli/command/notebook/Create.java
marikomedlock: Yes, I'd like to distinguish between the input/output layer and the command logic. Hopefully that will become clearer after [PF-423](https://broadworkbench.atlassian.net/browse/PF-423).

Are you planning to change this from a gcloud command to using a Java client library? If so, I'd be fine punting this code reorg until then.

src/main/java/bio/terra/cli/command/Notebooks.java
```diff
@@ -0,0 +1,14 @@
+package bio.terra.cli.command;
+
+import bio.terra.cli.command.notebooks.Create;
+import picocli.CommandLine;
+
+/**
+ * This class corresponds to the second-level "terra notebook" command. This command is not valid by
```
marikomedlock, line 7: Nit: notebooks

src/main/java/bio/terra/cli/command/notebooks/Create.java
```diff
@@ -0,0 +1,108 @@
...
+import bio.terra.cli.app.DockerToolsManager;
+import bio.terra.cli.model.GlobalContext;
+import bio.terra.cli.model.TerraUser;
+import bio.terra.cli.model.WorkspaceContext;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.Callable;
+import picocli.CommandLine;
+
+/** This class corresponds to the third-level "terra notebook create" command. */
```
marikomedlock, line 13: Nit: notebooks

marikomedlock (approved):

src/main/java/bio/terra/cli/command/notebooks/Create.java
```diff
@@ -0,0 +1,108 @@
...
+    WorkspaceContext workspaceContext = WorkspaceContext.readFromFile();
+    workspaceContext.requireCurrentWorkspace();
+
+    AuthenticationManager authenticationManager =
+        new AuthenticationManager(globalContext, workspaceContext);
+    authenticationManager.loginTerraUser();
+    TerraUser user = globalContext.requireCurrentTerraUser();
+    String projectId = workspaceContext.getGoogleProject();
+
+    String command =
```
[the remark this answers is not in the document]
wchamber: I added a link to the gcloud ref. Longer term, I'd like to move this into the Notebook service and document it there. Good enough for now?
cloud console from what was created with the GUI. (I removed the oslogin, not entirely sure why I needed it at one point, but working fine without it now.)

src/main/java/bio/terra/cli/command/notebooks/Create.java
wchamber: It's a good question. I'll check in with Nicole to get her take.

src/main/java/bio/terra/cli/command/notebooks/Create.java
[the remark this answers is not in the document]
gjuggler: sgtm – good point that this is not the final resting place for these details anyway :)

src/main/java/bio/terra/cli/command/notebook/Create.java
wchamber: Yes to using a Java client library, but medium term once we have a notebook service up and running. So not for Alpha 1, but yes for Alpha 2. (For the not-yet-created notebooks stop/start commands, those will stay as gcloud commands.) Is that too far off? If yes, could I do it in a follow on to this PR?

Is the correct pattern to create a `NotebooksManager` with a `createNotebook` function with 5 nullable string arguments? (Are defaults command logic or manager logic?)

marikomedlock: Alpha 2 is fine for me. Main benefit to doing it sooner would be to have all the notebooks related code in one place, but it seems like there's not much shared now anyway.

Yes, I think a Notebooks or NotebooksManager class with a create/Notebook function sounds good. Good question about defaults. I don't have a strong preference, but would probably define them in the notebooks/manager class myself.

marikomedlock (approved):

src/main/java/bio/terra/cli/command/Notebooks.java
```diff
@@ -0,0 +1,14 @@
...
+ * itself; it is just a grouping keyword for it sub-commands.
+ */
+@CommandLine.Command(
```
aimiebell, line 10: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

src/main/java/bio/terra/cli/command/notebooks/Create.java
```diff
@@ -0,0 +1,108 @@
...
+import bio.terra.cli.model.GlobalContext;
+import bio.terra.cli.model.TerraUser;
+import bio.terra.cli.model.WorkspaceContext;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.Callable;
+import picocli.CommandLine;
+
+/** This class corresponds to the third-level "terra notebook create" command. */
+@CommandLine.Command(
```
aimiebell, line 14: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

src/main/java/bio/terra/cli/command/Notebooks.java
aimiebell, line 10: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

src/main/java/bio/terra/cli/command/notebooks/Create.java
aimiebell, line 14: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Method names should start with a lower case letter (NM_METHOD_NAMING_CONVENTION)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-method-names-should-start-with-a-lower-case-letter-nm-method-naming-convention).
Methods should be verbs, in mixed case with the first letter lowercase, with the first letter of each internal word capitalized.

PR #6, opened 2021-02-08
PF-339 Add describe, list, start, and stop initial notebooks commands.

wchamber:
Longer term, I expect the describe and list commands to talk to the Notebooks Service. As-is, no formatting is undertaken, but I'd like to get these commands in for an initial semi-usable milestone to pass off to Solutions folks.

Also tweak location description in other commands.

### DIFF
--- a/src/main/java/bio/terra/cli/command/notebooks/Create.java
+++ b/src/main/java/bio/terra/cli/command/notebooks/Create.java
@@ -29,7 +29,7 @@ public class Create implements Callable<Integer> {
       names = "location",
       defaultValue = "us-central1-a",
       description =
-          "The Google Cloud location to create the instance within, by default '${DEFAULT-VALUE}'.")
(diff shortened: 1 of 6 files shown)


src/main/java/bio/terra/cli/command/notebooks/Create.java
```diff
@@ -28,7 +28,8 @@
   @CommandLine.Option(
       names = "location",
       defaultValue = "us-central1-a",
-      description = "The Google Cloud location to create the instance within, e.g. us-central1-a")
+      description =
+          "The Google Cloud location to create the instance within, by default '${DEFAULT-VALUE}'.")
```
marikomedlock, line 32: Nit: Slightly awkward wording. Maybe the GC location of the instance?

marikomedlock (approved):

gjuggler (approved):

PR #7, opened 2021-02-08
PF-439: Update create workspace to use latest WSM client library.

marikomedlock:
Updated the "terra workspace create" command to call the new WSM endpoints for creating a cloud context and polling for job completion.

Note: This change is required for the "terra workspace create" command to work again. WSM has already updated their (dev) server, so the CLI is broken until this fix goes in.

### DIFF
--- a/src/main/java/bio/terra/cli/utils/WorkspaceManagerUtils.java
+++ b/src/main/java/bio/terra/cli/utils/WorkspaceManagerUtils.java
@@ -8,6 +8,7 @@ import bio.terra.workspace.api.WorkspaceApi;
 import bio.terra.workspace.client.ApiClient;
 import bio.terra.workspace.model.CloudContext;
 import bio.terra.workspace.model.CreateCloudContextRequest;
+import bio.terra.workspace.model.CreateCloudContextResult;
 import bio.terra.workspace.model.CreateWorkspaceRequestBody;
 import bio.terra.workspace.model.GrantRoleRequestBody;
 import bio.terra.workspace.model.IamRole;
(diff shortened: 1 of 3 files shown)


src/main/java/bio/terra/cli/utils/WorkspaceManagerUtils.java
```diff
@@ -98,30 +103,38 @@ public static WorkspaceDescription createWorkspace(ApiClient apiClient) {
...
             numJobPollingTries,
-            workspaceWithContext.getId(),
-            googleProjectId);
+            workspaceId,
+            jobId);
+        jobReport = jobsApi.retrieveJob(jobId.toString());
+        logger.debug("create workspace jobReport: {}", jobReport);
         numJobPollingTries++;
-        if (googleProjectId == null) {
+        if (jobReport.getStatus().equals(JobReport.StatusEnum.RUNNING)) {
```
zloery, line 124: With the new API you could instead poll on `workspaceApi.createCloudContextResult`, which has a `jobReport` property. 
The biggest difference would be that you get an ErrorReport if something goes wrong during creation, though I don't know whether you want to surface that for the CLI or not. It also provides the actual result, although that doesn't matter since you're returning a whole WorkspaceDescription and not just the context.

zloery (approved):
LGTM on WM client usage, though I don't have much context on the CLI codebase.
One alternative in comments, but it depends on how you want to surface errors in CLI.

src/main/java/bio/terra/cli/utils/WorkspaceManagerUtils.java
marikomedlock: Good point about the ErrorReport. I added it to the debug logging.

src/main/java/bio/terra/cli/utils/WorkspaceManagerUtils.java
```diff
@@ -86,6 +91,7 @@ public static SystemStatus getStatus(ApiClient apiClient) {
    */
   public static WorkspaceDescription createWorkspace(ApiClient apiClient) {
     WorkspaceApi workspaceApi = new WorkspaceApi(apiClient);
+    JobsApi jobsApi = new JobsApi(apiClient);
```
zloery, line 94: nit: no longer needed

zloery (approved):

wchamber (approved):

PR #9, opened 2021-02-09
PF-424: Write Docker container output to stdout.

marikomedlock:
- Execute the Docker logs command before waiting for the container to exit.
- Set the FollowStream boolean on the logs command to stream the output while the container is running.
- Print the log frame to stdout each time the ResultCallback helper class is called.


src/main/java/bio/terra/cli/app/DockerToolsManager.java
```diff
@@ -250,55 +257,74 @@ private Integer waitForDockerContainerToExit(String containerId) {
...
+   * Read the Docker container logs and write them to standard out.
+   *
+   * @param containerId id of the container
+   */
+  private void outputLogsForDockerContainer(String containerId) {
+    dockerClient
+        .logContainerCmd(containerId)
+        .withStdOut(true)
+        .withStdErr(true)
+        // .withTimestamps(true)
```
wchamber, line 279: remove?

src/main/java/bio/terra/cli/app/DockerToolsManager.java
```diff
@@ -250,55 +257,74 @@ private Integer waitForDockerContainerToExit(String containerId) {
...
   /** Helper class for reading Docker container logs into a string. */
   private static class LogContainerTestCallback extends ResultCallback.Adapter<Frame> {
 
     protected final StringBuffer log = new StringBuffer();
-    List<Frame> collectedFrames = new ArrayList<>();
-    boolean collectFrames = false;
+    List<Frame> framesList = new ArrayList<>();
+
+    boolean buildSingleStringOutput;
+    boolean buildFramesList;
```
wchamber, line 292: mind a couple comments on what these are?

src/main/java/bio/terra/cli/app/DockerToolsManager.java
```diff
@@ -250,55 +257,74 @@ private Integer waitForDockerContainerToExit(String containerId) {
...
     @Override
     public void onNext(Frame frame) {
-      if (collectFrames) collectedFrames.add(frame);
-      log.append(new String(frame.getPayload(), Charset.forName("UTF-8")));
+      String logStr = new String(frame.getPayload(), Charset.forName("UTF-8"));
+
+      // TODO: PF-423. Calling sysout.println here breaks the model of printing user-facing output
+      // from the command classes only.
+      // Revisit this once we have better model for centralizing output across all commands/rest of
+      // the codebase.
```
wchamber, line 310: no action on this pr: one thing we could do is have our own interface to wrap system.out.println and have that be in charge with centralizing output. we would still have to allow printing output from outside of the command classes.

wchamber (approved):

src/main/java/bio/terra/cli/app/DockerToolsManager.java
marikomedlock: Yes, that's along the lines of what I'm thinking for the follow-on ticket.

gjuggler (approved):

PR #11, opened 2021-02-11
PF-449: Change default server to wchamber-dev, update documentation.

marikomedlock:
Temporarily changed the default server from `terra-dev` to `wchamber-dev`, because the Terra dev services are behind a firewall that requires users to be on the Broad VPN. There is an open ticket with Broad DevOps to address this. In the meantime, personal development deployments are apparently not behind the firewall, so we're using one of those as a temporary workaround.

Added documentation from today's meeting with Willy/Will/Mariko to get Willy started on adding dsub to the CLI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,8 +26,9 @@ terra
 #### Authentication
 1. Register a Terra account with the [dev instance](https://bvdp-saturn-dev.appspot.com/#) 
 to go through authentication. I think this needs to be a non-Google/Verily corporate account.
-2. Keep going through the warning login screen ("! Google hasn't verified this app" -> click 
-"Advanced" -> "Go to ... (unsafe)") that shows up since the CLI is not a Google-verified app yet.
+2. `terra auth login` launches an OAuth flow that pops out a browser window with a warning login
+page ("! Google hasn't verified this app"). This shows up because the CLI is not yet a Google-verified
+app. Click through the warnings ("Advanced" -> "Go to ... (unsafe)") to complete the login.
 
 Note: The default server has been temporarily changed from `terra-dev` to `wchamber-dev`.
 This is because the Terra dev services are behind a firewall that requires users to be
--- a/README.md
+++ b/README.md
@@ -2,12 +2,13 @@
 
 1. [Setup and run](#setup-and-run)
 2. [Requirements](#requirements)
+    * [Authentication](#authentication)
+    * [External data](#external-data)
 3. [Example usage](#example-usage)
 4. [Commands description](#commands-description)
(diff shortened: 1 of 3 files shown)


src/main/java/bio/terra/cli/service/ServerManager.java
```diff
@@ -74,7 +74,7 @@ public boolean pingServerStatus() {
   }
 
   // This variable defines the server that the CLI points to by default.
-  private static final String DEFAULT_SERVER_FILENAME = "terra-dev.json";
+  private static final String DEFAULT_SERVER_FILENAME = "wchamber-dev.json";
```
gjuggler, line 77: Aha – I'm guessing this is intended to (1) not require being on the VPN, and also (2) provide a more stable version of WSM to prevent breakages due to just-merged PRs or whatnot. (Depends somewhat on Will not doing too much with his personal WSM, which I guess is safe at least for awhile?)

Nice approach!

gjuggler (approved):

src/main/java/bio/terra/cli/service/ServerManager.java
marikomedlock: The intent was just (1) -- so that Willy and Nicole don't need to be on the VPN to test the CLI. I'd prefer to use WSM dev long-term so that we don't have to worry about maintaining a separate instance.

Credit to @wchamber for the workaround here -- I didn't know personal dev environments weren't behind the firewall!

wchamber: Yep, 2) is actually a downside. I don't expect my personal environment to be more stable and will need to ad-hoc fix or monitor it.

README.md
```diff
@@ -20,11 +21,25 @@ terra
...
+2. Docker 20.10.2 (Must be running)
 
-* Other Docker versions may also work, but this is the only one tested so far.
-
-Make sure Docker is running.
+#### Authentication
+1. Register a Terra account with the [dev instance](https://bvdp-saturn-dev.appspot.com/#) 
+to go through authentication. I think this needs to be a non-Google/Verily corporate account.
+2. Keep going through the warning login screen ("! Google hasn't verified this app" -> click 
+"Advanced" -> "Go to ... (unsafe)") that shows up since the CLI is not a Google-verified app yet.
```
wchamber, line 30: Is this after running `terra auth login`? little unclear following the instructions for registering with the app terra dev instance.

wchamber (approved):

README.md
marikomedlock: Yes. Re-worded a little.

PR #12, opened 2021-02-15
PF-340 Resource commands. Pass in resolved references to Docker container.

marikomedlock:
1. Resources commands
- First pass at code structure to persist cloud resources and data references on disk only. (This will later have to be changed to use the WSM endpoints, when they are ready.)
- Call GCS directly to create bucket resource.  For this ticket, handle only GCS buckets in the workspace.

2. Resources references
- Pass in resolved resource references to Docker container as environment variables.
- Allow specifying the resource references in the command (e.g. nextflow run -work-dir {TERRA_MY_BUCKET}

3. Nextflow
- Updated Nextflow install in Dockerfile to set the mode for Google Life Sciences.

### DIFF
--- a/src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
+++ b/src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
@@ -81,6 +81,9 @@ public class DockerAppsRunner {
   /**
    * Run a command inside the Docker container for external tools.
    *
+   * <p>This method substitutes any Terra references in the command and also adds the Terra
+   * references as environment variables in the container.
+   *
    * @param command the full string command to execute in a bash shell (bash -c ..cmd..)
    */
   public void runToolCommand(String command) {
@@ -91,6 +94,9 @@ public class DockerAppsRunner {
    * Run a command inside the Docker container for external tools. Allows adding environment
    * variables and bind mounts beyond what the terra_init script requires. The environment variables
    * and bind mounts expected by the terra_init script will be added to those passed in.
+   *
+   * <p>This method substitutes any Terra references in the command and also adds the Terra
+   * references as environment variables in the container.
    *
    * @param command the full string command to execute in a bash shell (bash -c ..cmd..)
    * @param workingDir the directory where the commmand will be executed
@@ -361,7 +367,7 @@ public class DockerAppsRunner {
    *
    * <p>e.g. TERRA_MY_BUCKET -> gs://terra-wsm-test-9b7511ab-my-bucket
    *
-   * @return
+   * @return a map of Terra references (name -> cloud id)
    */
   private Map<String, String> buildMapOfTerraReferences() {
     // build a map of reference string -> resolved value
--- a/src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
+++ b/src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
@@ -172,7 +172,7 @@ public class DockerAppsRunner {
     Map<String, String> terraInitEnvVars = new HashMap<>();
     String googleProjectId = workspaceContext.getGoogleProject();
     terraInitEnvVars.put(
-        "TERRA_PET_KEY_FILE",
+        "GOOGLE_APPLICATION_CREDENTIALS",
         PET_KEYS_MOUNT_POINT + "/" + currentUser.getPetKeyFile(googleProjectId).getName());
     terraInitEnvVars.put("TERRA_GOOGLE_PROJECT_ID", googleProjectId);
     for (Map.Entry<String, String> terraInitEnvVar : terraInitEnvVars.entrySet()) {
--- a/src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
+++ b/src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
@@ -107,6 +107,12 @@ public class DockerAppsRunner {
     // check that the current workspace is defined
     workspaceContext.requireCurrentWorkspace();
 
+    // substitute any Terra references in the command
+    // also add the Terra references as environment variables in the container
+    Map<String, String> terraReferences = buildMapOfTerraReferences();
+    command = replaceTerraReferences(terraReferences, command);
+    envVars.putAll(terraReferences);
+
     buildDockerClient();
 
     // create and start the docker container. run the terra_init script first, then the given
@@ -166,9 +172,9 @@ public class DockerAppsRunner {
     Map<String, String> terraInitEnvVars = new HashMap<>();
     String googleProjectId = workspaceContext.getGoogleProject();
     terraInitEnvVars.put(
-        "PET_KEY_FILE",
+        "TERRA_PET_KEY_FILE",
         PET_KEYS_MOUNT_POINT + "/" + currentUser.getPetKeyFile(googleProjectId).getName());
-    terraInitEnvVars.put("GOOGLE_PROJECT_ID", googleProjectId);
+    terraInitEnvVars.put("TERRA_GOOGLE_PROJECT_ID", googleProjectId);
     for (Map.Entry<String, String> terraInitEnvVar : terraInitEnvVars.entrySet()) {
       if (envVars.get(terraInitEnvVar.getKey()) != null) {
         throw new RuntimeException(
@@ -331,7 +337,12 @@ public class DockerAppsRunner {
     }
   }
 
-  /** Utility method for concatenating a command and its arguments. */
+  /**
+   * Utility method for concatenating a command and its arguments.
+   *
+   * @param cmd command name (e.g. gsutil)
+   * @param cmdArgs command arguments (e.g. ls, gs://my-bucket)
+   */
   public static String buildFullCommand(String cmd, List<String> cmdArgs) {
     String fullCommand = cmd;
     if (cmdArgs != null && cmdArgs.size() > 0) {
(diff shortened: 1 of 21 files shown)


src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
```diff
@@ -340,4 +351,47 @@ public static String buildFullCommand(String cmd, List<String> cmdArgs) {
...
+  /**
+   * Build a map of Terra references to use in parsing the CLI command string, and in setting
+   * environment variables in the Docker container.
+   *
+   * <p>The list of references are TERRA_[...] where [...] is the name of a cloud resource. The
+   * cloud resource can be controlled or external.
+   *
+   * <p>e.g. TERRA_MY_BUCKET -> gs://terra-wsm-test-9b7511ab-my-bucket
+   *
+   * @return
```
wchamber, line 364: empty return tag

src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
```diff
@@ -340,4 +351,47 @@ public static String buildFullCommand(String cmd, List<String> cmdArgs) {
...
+   * @param cmd the original command string
+   * @return the modified command string
+   */
+  private String replaceTerraReferences(Map<String, String> terraReferences, String cmd) {
+    // loop through the map entries
+    String modifiedCmd = cmd;
+    for (Map.Entry<String, String> terraReference : terraReferences.entrySet()) {
+      // loop through the map entries, replacing each one in the command string (case insensitive)
+      modifiedCmd =
+          modifiedCmd.replaceAll(
```
wchamber, line 392: this doesn't feel very robust, but I think it's fine to go through with this for now with a ticket to follow up later.
There's no escaping. What happens if one of my cloud resource id is a substring of my cloud resource names?

src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
```diff
@@ -107,6 +107,12 @@ public void runToolCommand(
     // check that the current workspace is defined
     workspaceContext.requireCurrentWorkspace();
 
+    // substitute any Terra references in the command
```
wchamber, line 110: we should update the javadoc comment on this class, and even consider having a different method for doing terra reference substitution than anything else.
Do we also want to add something to the README on use in a later PR?

src/main/java/bio/terra/cli/command/resources/Create.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.concurrent.Callable;
+import picocli.CommandLine;
+import picocli.CommandLine.Command;
+
+/** This class corresponds to the third-level "terra resources create" command. */
+@Command(name = "create", description = "Create a new controlled resource.")
+public class Create implements Callable<Integer> {
+
+  @CommandLine.Option(
+      names = "--type",
```
wchamber, line 17: should all option params have `--` prefixes? I didn't do that for notebooks.

src/main/java/bio/terra/cli/context/WorkspaceContext.java
```diff
@@ -81,13 +95,21 @@ public void updateWorkspace(WorkspaceDescription terraWorkspaceModel) {
...
   @JsonIgnore
   public UUID getWorkspaceId() {
     return terraWorkspaceModel == null ? null : terraWorkspaceModel.getId();
   }
 
-  /** Getter for the Google project backing the current Terra workspace. */
+  /**
+   * Getter for the Google project backing the current Terra workspace.
+   *
+   * @return the Google project id
```
wchamber, line 111: @gjuggler javadoc boilerplate. (no action needed)

src/main/java/bio/terra/cli/service/utils/GoogleCloudStorage.java
```diff
@@ -0,0 +1,81 @@
...
+  private static final Logger logger = LoggerFactory.getLogger(GoogleCloudStorage.class);
+
+  // the Terra user whose credentials will be used to call authenticated requests
+  private final TerraUser terraUser;
+
+  // the google project id of the workspace
+  private final String googleProjectId;
+
+  // the client object used for talking to Google Cloud Storage
+  private Storage storageClient;
```
wchamber, line 22: I think this should use CRL's StorageCow. Should be close to a drop-in replacement. https://github.com/DataBiosphere/terra-cloud-resource-lib/blob/master/google-storage/src/main/java/bio/terra/cloudres/google/storage/StorageCow.java

src/main/java/bio/terra/cli/service/utils/GoogleCloudStorage.java
```diff
@@ -0,0 +1,81 @@
...
+
+  /**
+   * Build the GCS client object for the given Terra user. If terraUser is null, this method builds
+   * the client object without an access token set.
+   */
+  private void buildClientForTerraUser() {
+    StorageOptions.Builder storageOptions = StorageOptions.newBuilder();
+    storageOptions.setProjectId(googleProjectId);
+
+    if (terraUser != null) {
```
wchamber, line 45: what happens if it is null? What still works?

src/main/java/bio/terra/cli/context/CloudResource.java
```diff
@@ -0,0 +1,38 @@
...
+    this.type = type;
+    this.isControlled = isControlled;
+  }
+
+  /** Type of cloud resource. */
+  public enum Type {
+    bucket(true);
+
+    // true = this cloud resource is also a data reference
+    public final boolean isDataReference;
```
wchamber, line 32: what does it mean to be a data reference? what resources aren't data references? If we're not using it yet, leave it out.

src/main/java/bio/terra/cli/service/WorkspaceManager.java
```diff
@@ -169,4 +174,91 @@ public RoleBindingList listUsersOfWorkspace() {
...
+
+  /**
+   * Create a new controlled resource in the workspace.
+   *
+   * @param resourceType type of resource to create
+   * @param resourceName name of resource to create
+   * @return the cloud resource that was created
+   */
+  public CloudResource createControlledResource(
+      CloudResource.Type resourceType, String resourceName) {
```
wchamber, line 202: I think we're headed towards having a different API for each resource, as they will have different parameters for creation. but also fine to start this way in this PR.
https://github.com/DataBiosphere/terra-workspace-manager/blob/ac1f4767c3ad08786ba637f7249cdb90e7131311/src/main/resources/api/service_openapi.yaml#L296

wchamber (changes requested):

src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
marikomedlock: Fixed.

src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
```diff
@@ -107,6 +107,12 @@ public void runToolCommand(
...
+    // also add the Terra references as environment variables in the container
+    Map<String, String> terraReferences = buildMapOfTerraReferences();
+    command = replaceTerraReferences(terraReferences, command);
```
gjuggler, line 113: I guess I'm wondering why env variables aren't enough on their own here? Does the main Docker command not allow for env variable substitution for some reason? If there's some reference source on why that's the case, a short comment calling out to the docs might help here.

gjuggler (approved):
Will already hit on most everything I might have noticed. I think you've managed a nice tradeoff in reasonable structure on the resource stuff, without going too overboard given it's subject to change and will eventually be more server-side anyway.

src/main/java/bio/terra/cli/command/resources/Create.java
marikomedlock: I was following gcloud example (e.g. --project, --help, --billing-project). I don't have a strong opinion either way, but we should be consistent across commands.

src/main/java/bio/terra/cli/service/utils/GoogleCloudStorage.java
marikomedlock: Well you could only call unauthenticated endpoints...and I don't know of any for GCS. This was a copy/paste leftover from Terra service classes. Removed.

src/main/java/bio/terra/cli/service/WorkspaceManager.java
marikomedlock: Okay, will adapt once the WSM stuff is settled.

src/main/java/bio/terra/cli/context/CloudResource.java
marikomedlock: A data reference means something that will show up in the planned `terra data-refs list` command. I'll remove it here and add it back when I implement that command.

src/main/java/bio/terra/cli/service/utils/GoogleCloudStorage.java
marikomedlock: What benefit would using CRL offer?

wchamber: Minimal at the moment, but consistent dev access to logging and utilities that we haven't yet built. @gjuggler  is CRL only for services or also worth using for the CLI?

src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
marikomedlock: I think the env var substitution is a bash/shell thing, so it's not that Docker doesn't support it, it's that we need to bypass the bash env var substitution on the host machine so that the Terra CLI can handle it.

For example:
`
host-machine> env
HOST_ENV_VAR=123
host-machine> terra nextflow run $HOST_ENV_VAR
`
causes the Terra CLI Java code to see the command arguments `nextflow run 123` only. It doesn't know about any env var substitution that's already happened.

I wanted to bypass the env var substitution on the host machine so I can do it in the Terra CLI. So I used a slightly different syntax:
`
host-machine> terra nextflow run {HOST_ENV_VAR}
`
causes the Terra CLI Java code to see the command arguments `nextflow run {HOST_ENV_VAR}` and then I can do the substitution in the Terra CLI before passing the modified command string to Docker.

src/main/java/bio/terra/cli/service/utils/GoogleCloudStorage.java
marikomedlock: Ok, just noting here that this code will all be ripped out once WSM controlled resources are ready. However, there may still be some updates to controlled resources that we'd handle directly -- I'm not totally clear on that, yet.

src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
marikomedlock: Great point. Added a follow-on ticket to address this [PF-459](https://broadworkbench.atlassian.net/browse/PF-459)

src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
marikomedlock: I added a brief section on the resources commands to the README. But there is already a follow-on ticket [PF-421](https://broadworkbench.atlassian.net/browse/PF-421) for writing a section on the env vars and using references in app commands. I want to use the GLS NF tutorial for that section, but I'm getting a permission error running that. So the follow-on ticket is to debug the permission error and then write the section.

src/main/java/bio/terra/cli/service/utils/GoogleCloudStorage.java
gjuggler: Good question re: the scope of CRL usage. I think we should use it wherever it's useful. In this case, I agree it's probably not worth adding the dependency for short-lived code. But if we do find ourselves repeatedly adding direct cloud platform interactions into the CLI, we should reassess.

Re: the CLI updating bucket policies directly, based on Dan's explorations of WSM controlled resources, current plans have all controlled bucket modifications needing to go through the WSM API, basically because permissions are not fine-grained enough. See [this section](https://docs.google.com/document/d/1F28KHeVpTh0Zuqdux30ONabMgjj6CNmyv7B8iqcXnsE/edit?resourcekey=0-5ndsiOmBJdG7bB1gZXOy1g#bookmark=id.au48n2159vom) for details.

wchamber (approved):

src/main/java/bio/terra/cli/service/utils/GoogleCloudStorage.java
marikomedlock: Thanks for the bookmark on bucket permissions. For other operations or types of controlled resources, we may end up talking directly to the cloud -- in that case, I'll look at using CRL.

PR #13, opened 2021-02-16
PF-444, PF-453: SAM retries, new user registration

marikomedlock:
1. Retry to all SAM requests.
2. When looking up a user in SAM returns Not Found, try to register a new user.
3. Some error-handling cleanup around SAM requests. Exceptions in SamService are now propagated up to the command classes. A future ticket PF-423 will do more cleanup of error-handling to make it more consistent across classes.

### DIFF
--- a/src/main/java/bio/terra/cli/service/utils/HttpUtils.java
+++ b/src/main/java/bio/terra/cli/service/utils/HttpUtils.java
@@ -131,8 +131,9 @@ public class HttpUtils {
    * @throws Exception if makeRequest throws an exception that is not retryable, or if the maximum
    *     number of retries was exhausted
    */
-  public static <T> T callWithRetries(
-      HttpRequestOperator<T> makeRequest, Predicate<Exception> isRetryable) throws Exception {
+  public static <T, E extends Exception> T callWithRetries(
+      HttpRequestOperator<T, E> makeRequest, Predicate<Exception> isRetryable)
+      throws E, InterruptedException {
     return callWithRetries(
         DEFAULT_MAXIMUM_RETRIES, DEFAULT_MILLISECONDS_SLEEP_FOR_RETRY, makeRequest, isRetryable);
   }
@@ -148,13 +149,14 @@ public class HttpUtils {
    * @throws Exception if makeRequest throws an exception that is not retryable, or if the maximum
    *     number of retries was exhausted
    */
-  public static <T> T callWithRetries(
+  public static <T, E extends Exception> T callWithRetries(
       int maxRetries,
       int sleepDurationMS,
-      HttpRequestOperator<T> makeRequest,
+      HttpRequestOperator<T, E> makeRequest,
       Predicate<Exception> isRetryable)
-      throws Exception {
+      throws E, InterruptedException {
     int numTries = 0;
+    Exception lastRetryableException = null;
     do {
       numTries++;
       try {
@@ -164,6 +166,9 @@ public class HttpUtils {
         // if the exception is not retryable, then quit polling
         if (!isRetryable.test(ex)) {
           throw ex;
+        } else {
+          // keep track of the last retryable exception so we can re-throw it in case of a timeout
+          lastRetryableException = ex;
         }
         logger.info("Caught retryable exception: {}", ex);
       }
@@ -172,7 +177,10 @@ public class HttpUtils {
       Thread.sleep(sleepDurationMS);
     } while (numTries <= maxRetries);
 
-    throw new RuntimeException("Http request with retries timed out after " + numTries + " tries.");
+    // request with retries timed out
+    throw new RuntimeException(
+        "Http request with retries timed out after " + numTries + " tries.",
+        lastRetryableException);
   }
 
   /**
@@ -181,7 +189,7 @@ public class HttpUtils {
    *
    * @param <T> type of the Http response (i.e. return type of the makeRequest method)
    */
-  public interface HttpRequestOperator<T> {
-    T makeRequest() throws Exception;
+  public interface HttpRequestOperator<T, E extends Exception> {
+    T makeRequest() throws E;
   }
 }
--- a/src/main/java/bio/terra/cli/service/utils/HttpUtils.java
+++ b/src/main/java/bio/terra/cli/service/utils/HttpUtils.java
@@ -11,9 +11,19 @@ import java.net.URL;
 import java.net.URLEncoder;
 import java.nio.charset.Charset;
 import java.util.Map;
+import java.util.function.Predicate;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /** Utility methods for making raw HTTP requests (e.g. in place of using a client library). */
 public class HttpUtils {
+  private static final Logger logger = LoggerFactory.getLogger(HttpUtils.class);
+
+  // default value for the maximum number of times to retry HTTP requests
+  private static final int DEFAULT_MAXIMUM_RETRIES = 10;
+
+  // default value for the number of milliseconds to sleep between retries
+  private static final int DEFAULT_MILLISECONDS_SLEEP_FOR_RETRY = 1000;
 
   private HttpUtils() {}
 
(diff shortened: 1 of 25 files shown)


src/main/java/bio/terra/cli/service/utils/HttpUtils.java
```diff
@@ -108,4 +118,70 @@ public static HttpResponse sendHttpRequest(
...
+   * @return the Http response
+   * @throws Exception if makeRequest throws an exception that is not retryable, or if the maximum
+   *     number of retries was exhausted
+   */
+  public static <T> T callWithRetries(
+      int maxRetries,
+      int sleepDurationMS,
+      HttpRequestOperator<T> makeRequest,
+      Predicate<Exception> isRetryable)
+      throws Exception {
```
wchamber, line 156: `throws Exception` isn't very helpful. Having a checked exception means we want the caller of the function to be able to handle the exception somehow. We should give them a meaningful type if we expect them to do something about it. If we don't, this function should handle what it can and then throw some runtime exception.

It looks like this is causing us to change the type to throw checked exceptions all the way up the stack. That seems like we don't actually want clients to do anything with the exception.

In this case, the Sam ApIClient calls don't actually throw checked exceptions, so we don't need to throw checked exceptions either. Our `HttpRequestOperator` doesn't need to throw a checked exception, and neither does our `callWithRetries` function. Instead, template the isRetryable with ApiException. When `makeRequest` throws an instance of that type, pass it to isRetryable. Otherwise throw the same runtime exception.
```
public static <T, E> T callWithRetries(
   ...
   Predicate<E> isRetryable) {
   ...
     try {
        logger.info("Http request attempt #{}", numTries);
        return makeRequest.makeRequest();
      } catch (RuntimeException ex) {
        // if the exception is not retryable, then quit polling
        if (!(ex instanceof E) || !isRetryable.test((E) ex)) {
          throw ex;
        }
        logger.info("Caught retryable exception: {}", ex);
      }
 ```

File: src/main/java/bio/terra/cli/service/utils/HttpUtils.java [hunk 2]
```diff
@@ -108,4 +118,70 @@ public static HttpResponse sendHttpRequest(
...
+  }
+
+  /**
+   * Function interface for making a retryable Http request. This interface is explicitly defined so
+   * that it can throw an exception (i.e. Supplier does not have this method annotation).
+   *
+   * @param <T> type of the Http response (i.e. return type of the makeRequest method)
+   */
+  public interface HttpRequestOperator<T> {
+    T makeRequest() throws Exception;
```
wchamber, line 185: we can template this on the exception being thrown as well that way we don't have to throw a generic  `Exception`. `HttpRequestOperator<T, E> { T makeRequest() throws E; }` But it doesn't look like any of our functions throw checked exceptions yet, so we do not need an exception throwing functional interface at this time. 
When we want to do retries on functions that both throw checked exceptions and don't, then we'll also want a checked exception version.

src/main/java/bio/terra/cli/service/utils/SamService.java
```diff
@@ -67,34 +68,17 @@ private void buildClientForTerraUser() {
...
-   * @return the SAM status object
+   * @return the SAM status object, null if there was an error checking the status
    */
   public SystemStatus getStatus() {
     StatusApi statusApi = new StatusApi(apiClient);
     SystemStatus status = null;
     try {
-      status = statusApi.getSystemStatus();
+      status =
+          HttpUtils.callWithRetries(() -> statusApi.getSystemStatus(), (ex) -> isRetryable(ex));
```
wchamber, line 81: I *think* we can replace the lambda `(ex) -> isRetryable(ex)` with the double colon operator method reference `SamService::isRetryable`

src/main/java/bio/terra/cli/service/utils/HttpUtils.java
```diff
@@ -108,4 +118,70 @@ public static HttpResponse sendHttpRequest(
...
+          throw ex;
+        }
+        logger.info("Caught retryable exception: {}", ex);
+      }
+
+      // sleep before retrying
+      Thread.sleep(sleepDurationMS);
+    } while (numTries <= maxRetries);
+
+    throw new RuntimeException("Http request with retries timed out after " + numTries + " tries.");
```
wchamber, line 175: I think it'd be more helpful to throw the last seen exception when we run out of retries.

wchamber (changes requested):

src/main/java/bio/terra/cli/service/utils/HttpUtils.java
marikomedlock: Agreed, done.

src/main/java/bio/terra/cli/service/utils/SamService.java
marikomedlock: Done.

src/main/java/bio/terra/cli/service/utils/HttpUtils.java
marikomedlock: I'm confused about "the Sam ApIClient calls don't actually throw checked exceptions". All the Sam ApiClient calls throw ApiException, which is a checked exception.

src/main/java/bio/terra/cli/service/utils/HttpUtils.java
marikomedlock: Adding an additional template for the exception type seems good -- I'll give that a try.

src/main/java/bio/terra/cli/service/utils/HttpUtils.java
```diff
@@ -11,9 +11,19 @@
...
 
 /** Utility methods for making raw HTTP requests (e.g. in place of using a client library). */
 public class HttpUtils {
+  private static final Logger logger = LoggerFactory.getLogger(HttpUtils.class);
+
+  // default value for the maximum number of times to retry HTTP requests
+  private static final int DEFAULT_MAXIMUM_RETRIES = 10;
+
+  // default value for the number of milliseconds to sleep between retries
+  private static final int DEFAULT_MILLISECONDS_SLEEP_FOR_RETRY = 1000;
```
wchamber, line 26: nit: Prefer using java.time.Duration for durations e.g. `private static final Duration DEFAULT_SLEEP_FOR_RETRY = Duration.ofSeconds(1);`. I realize that we'll have to convert it `toMillis()` for Thread sleep, but it's still much less error prone to work in terms of Durations until then. (yes, this is a tiny use case, but good java time habits/the number of bugs because of primitive time durations.)

src/main/java/bio/terra/cli/service/utils/HttpUtils.java
```diff
@@ -108,4 +118,78 @@ public static HttpResponse sendHttpRequest(
...
+    int numTries = 0;
+    Exception lastRetryableException = null;
+    do {
+      numTries++;
+      try {
+        logger.info("Http request attempt #{}", numTries);
+        return makeRequest.makeRequest();
+      } catch (Exception ex) {
+        // if the exception is not retryable, then quit polling
+        if (!isRetryable.test(ex)) {
```
wchamber, line 167: Do the instanceof checking here instead of forcing isRetryable to handle every exception type. So change `Predicate<E> isRetryable`. We're already committing to makeRequest using a particular exception type, `E`, let's extend that to the predicate.

src/main/java/bio/terra/cli/service/utils/HttpUtils.java
```diff
@@ -108,4 +118,78 @@ public static HttpResponse sendHttpRequest(
...
+        lastRetryableException);
+  }
+
+  /**
+   * Function interface for making a retryable Http request. This interface is explicitly defined so
+   * that it can throw an exception (i.e. Supplier does not have this method annotation).
+   *
+   * @param <T> type of the Http response (i.e. return type of the makeRequest method)
+   */
+  public interface HttpRequestOperator<T, E extends Exception> {
```
wchamber, line 192: nit: add `@FunctionalInterface`

src/main/java/bio/terra/cli/service/utils/HttpUtils.java
```diff
@@ -108,4 +118,78 @@ public static HttpResponse sendHttpRequest(
...
+   * @return the Http response
+   * @throws Exception if makeRequest throws an exception that is not retryable, or if the maximum
+   *     number of retries was exhausted
+   */
+  public static <T, E extends Exception> T callWithRetries(
+      int maxRetries,
+      int sleepDurationMS,
+      HttpRequestOperator<T, E> makeRequest,
+      Predicate<Exception> isRetryable)
+      throws E, InterruptedException {
```
wchamber, line 157: I don't see this throwing an InterruptedException. I don't think we even want our callers to care about InterruptedExceptions. Throws `E` is enough even if we run out of retries.

src/main/java/bio/terra/cli/service/utils/HttpUtils.java
```diff
@@ -108,4 +118,78 @@ public static HttpResponse sendHttpRequest(
...
+      }
+
+      // sleep before retrying
+      Thread.sleep(sleepDurationMS);
+    } while (numTries <= maxRetries);
+
+    // request with retries timed out
+    throw new RuntimeException(
+        "Http request with retries timed out after " + numTries + " tries.",
+        lastRetryableException);
```
wchamber, line 183: I think that instead of wrapping this in a RuntimeException we should throw the last exception.

src/main/java/bio/terra/cli/service/utils/HttpUtils.java
```diff
@@ -108,4 +118,78 @@ public static HttpResponse sendHttpRequest(
...
+        } else {
+          // keep track of the last retryable exception so we can re-throw it in case of a timeout
+          lastRetryableException = ex;
+        }
+        logger.info("Caught retryable exception: {}", ex);
+      }
+
+      // sleep before retrying
+      Thread.sleep(sleepDurationMS);
+    } while (numTries <= maxRetries);
```
wchamber, line 178: this sleeps even if it's out of retries. Maybe do an early throw when we catch the exception if we're out of retries?

src/main/java/bio/terra/cli/service/utils/HttpUtils.java
wchamber: Sorry, that's right. Switching to throwing the ApiException lgtm.

src/main/java/bio/terra/cli/service/utils/HttpUtils.java
marikomedlock: Can you check the class type of a type parameter at runtime? I don't think you can do instanceof [type parameter] without storing the Class beforehand somehow.

src/main/java/bio/terra/cli/service/utils/HttpUtils.java
marikomedlock: ok

src/main/java/bio/terra/cli/service/utils/HttpUtils.java
marikomedlock: done.

src/main/java/bio/terra/cli/service/utils/HttpUtils.java
marikomedlock: Thread.sleep throws an InterruptedException. I suppose I could put a try/catch in the retry method and wrap it in a runtime exception, but I think it's a more informative error message in the calling methods.

src/main/java/bio/terra/cli/service/utils/HttpUtils.java
marikomedlock: I disagree on this one. I think it's helpful to see that it was a timeout that caused the error, and still include the full exception stack trace.

src/main/java/bio/terra/cli/service/utils/HttpUtils.java
marikomedlock: done.

src/main/java/bio/terra/cli/service/utils/HttpUtils.java
wchamber: Argh, sorry, you're right. What you have is good.

src/main/java/bio/terra/cli/service/utils/HttpUtils.java
wchamber: I'll push back a little more. It's always a timeout when we run out of retries. What we want to do after the timeout may be dependent on the exception. When we wrap the error in the runtime exception, the calling code can no longer do anything about the ApiException.

wchamber (approved):
thanks! sorry for the bad comments :(
Last push on the runtime after retries, but up to you.

gjuggler (approved):

PR #15, opened 2021-02-18
PF-452: Groups and Spend commands.

marikomedlock:
- Added `terra groups` commands for managing SAM groups
- Added `terra spend` commands for managing access to the WSM spend profile.

- Updated README with instructions for granting access to the spend profile, with a preference for adding new users to a group I created called `enterprise-pilot-testers`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,8 +41,14 @@ by Workspace Manager. Your email needs to either be added as a user of that spen
 to a Terra group that is a user of that spend profile. This needs to be done by someone else with
 owner access to that spend profile.
 
-- [Preferred] Add a user to a Terra group that is a user of the spend profile with `terra groups add-user --group=enterprise-pilot-testers --policy=admin mmdevverily4@gmail.com`
-- Add a user directly to the spend profile with `terra spend enable --policy=owner mmdevverily@gmail.com`
+- [Preferred] Add a user to a Terra group that is a user of the spend profile. To also grant permission
+to add new members to the group, use `policy=admin` instead.
+
+`terra groups add-user --group=enterprise-pilot-testers --policy=member mmdevverily4@gmail.com`
+
+- Add a user directly to the spend profile. To also grant permission to add new users to the spend profile,
+user `policy=owner` instead.
+`terra spend enable --policy=user mmdevverily@gmail.com`
 
 #### External data 
 To allow supported applications (i.e. the ones shown by `terra app list`) to read or write data
--- a/README.md
+++ b/README.md
@@ -2,13 +2,18 @@
 
 1. [Setup and run](#setup-and-run)
 2. [Requirements](#requirements)
-    * [Authentication](#authentication)
+    * [Login](#login)
+    * [Spend Profile Access](#spend-profile-access)
     * [External data](#external-data)
 3. [Example usage](#example-usage)
 4. [Commands description](#commands-description)
+    * [Authentication](#authentication)
     * [Server](#server)
     * [Workspace](#workspace)
-    * [Supported tools](#supported-tools)
+    * [Resources](#resources)
+    * [Applications](#applications)
+    * [Groups](#groups)
+    * [Spend](#spend)
 
 -----
 
@@ -23,18 +28,21 @@ terra
 1. Java 11
 2. Docker 20.10.2 (Must be running)
 
-#### Authentication
+#### Login
 1. Use a Google account that is not a Google/Verily corporate account.
 2. `terra auth login` launches an OAuth flow that pops out a browser window with a warning login
 page ("! Google hasn't verified this app"). This shows up because the CLI is not yet a Google-verified
 app. Click through the warnings ("Advanced" -> "Go to ... (unsafe)") to complete the login.
 
-Note: The default server has been temporarily changed from `terra-dev` to `wchamber-dev`.
-This is because the Terra dev services are behind a firewall that requires users to be
-on the Broad VPN. There is an open ticket with Broad DevOps to address this. In the meantime,
-personal development deployments are apparently not behind the firewall, so we're using
-one of those as a temporary workaround. If you do have access to the Broad VPN, you can run 
-`terra server set terra-dev` to change it back to official Terra dev environment.
+#### Spend Profile Access
+In order to spend money (e.g. by creating a project and resources within it) in Terra, you need
+access to a billing account via a spend profile. Currently, there is a single spend profile used
+by Workspace Manager. Your email needs to either be added as a user of that spend profile or added
+to a Terra group that is a user of that spend profile. This needs to be done by someone else with
+owner access to that spend profile.
+
+- [Preferred] Add a user to a Terra group that is a user of the spend profile with `terra groups add-user --group=enterprise-pilot-testers --policy=admin mmdevverily4@gmail.com`
+- Add a user directly to the spend profile with `terra spend enable --policy=owner mmdevverily@gmail.com`
 
 #### External data 
 To allow supported applications (i.e. the ones shown by `terra app list`) to read or write data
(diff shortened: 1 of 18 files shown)


README.md
```diff
@@ -23,18 +28,21 @@ terra
...
-one of those as a temporary workaround. If you do have access to the Broad VPN, you can run 
-`terra server set terra-dev` to change it back to official Terra dev environment.
+#### Spend Profile Access
+In order to spend money (e.g. by creating a project and resources within it) in Terra, you need
+access to a billing account via a spend profile. Currently, there is a single spend profile used
+by Workspace Manager. Your email needs to either be added as a user of that spend profile or added
+to a Terra group that is a user of that spend profile. This needs to be done by someone else with
+owner access to that spend profile.
+
+- [Preferred] Add a user to a Terra group that is a user of the spend profile with `terra groups add-user --group=enterprise-pilot-testers --policy=admin mmdevverily4@gmail.com`
```
wchamber, line 44: Does --policy=admin mean the added user can add other users to the group. Do we want to default to giving every user the permission to add other users?

src/main/java/bio/terra/cli/command/groups/AddUser.java
```diff
@@ -0,0 +1,38 @@
...
+import picocli.CommandLine;
+import picocli.CommandLine.Command;
+
+/** This class corresponds to the third-level "terra groups add-user" command. */
+@Command(name = "add-user", description = "Add a user to a group.")
+public class AddUser implements Callable<Integer> {
+  @CommandLine.Parameters(index = "0", description = "The email of the user.")
+  private String user;
+
+  @CommandLine.Option(names = "--group", required = true, description = "The name of the group")
```
wchamber, line 17: not this PR, just top of mind: no informed opinion, but we should write down some guidance about when we `--` prefix options, vs no prefix options, vs use command line parameters. Not urgent, but good for us to develop some consistency. (I care much more about have a consistent Terra CLI style that we can point to than what our actual guidelines are)
Sorry if this is already written and I've missed it.

src/main/java/bio/terra/cli/command/groups/AddUser.java
```diff
@@ -0,0 +1,38 @@
...
+      required = true,
+      description = "The name of the policy: ${COMPLETION-CANDIDATES}")
+  private SamService.GroupPolicy policy;
+
+  @Override
+  public Integer call() {
+    GlobalContext globalContext = GlobalContext.readFromFile();
+    WorkspaceContext workspaceContext = WorkspaceContext.readFromFile();
+
+    new AuthenticationManager(globalContext, workspaceContext).loginTerraUser();
```
wchamber, line 31: doesn't need to be addressed in this PR: Do I need to have a workspace context to use the authentication manager? It'd probably be nice to be able to run the group/spend commands outside of a workspace context since those commands are independent of having a workspace. While the majority of commands are going to be workspace related, I think we will have a sizable fraction of commands that are above the level of workspaces.

src/main/java/bio/terra/cli/command/groups/ListUsers.java
```diff
@@ -0,0 +1,40 @@
...
+import bio.terra.cli.context.GlobalContext;
+import bio.terra.cli.context.WorkspaceContext;
+import bio.terra.cli.service.utils.SamService;
+import java.util.List;
+import java.util.concurrent.Callable;
+import picocli.CommandLine;
+import picocli.CommandLine.Command;
+
+/** This class corresponds to the third-level "terra groups list-users" command. */
+@Command(name = "list-users", description = "List the users in a group.")
```
wchamber, line 13: nit: Maybe `List the users within a policy in a group`? As-is, it sounds like this has all members of a group, when this is dividing membership by policy membership within a group. (Does the Sam "member" policy always have all members, or can you be an admin but not a member?)

src/main/java/bio/terra/cli/command/groups/RemoveUser.java
```diff
@@ -0,0 +1,38 @@
...
+import bio.terra.cli.auth.AuthenticationManager;
+import bio.terra.cli.context.GlobalContext;
+import bio.terra.cli.context.WorkspaceContext;
+import bio.terra.cli.service.utils.SamService;
+import java.util.concurrent.Callable;
+import picocli.CommandLine;
+import picocli.CommandLine.Command;
+
+/** This class corresponds to the third-level "terra groups remove-user" command. */
+@Command(name = "remove-user", description = "Remove a user from a group.")
```
wchamber, line 12: ditto, scoped to group+policy

src/main/java/bio/terra/cli/service/utils/SamService.java
```diff
@@ -146,6 +151,238 @@ public void getOrRegisterUser() {
...
+  }
+
+  /**
+   * Possible values for the policies on a SAM group. These values are defined as an enum in SAM's
+   * API YAML
+   * (https://github.com/broadinstitute/sam/blob/61135c798873d20a308be1e440b862bf9767c243/src/main/resources/swagger/api-docs.yaml#L383)
+   * but I don't see an enum in the client library. It looks like that was a bug in Swagger codegen
+   * until v2.1.5 (https://github.com/swagger-api/swagger-codegen/pull/1740), but I'm not sure if
+   * that applies to the version that SAM is using.
+   */
```
wchamber, line 219: not urgent, but might be good for us to eventually file a few tickets for all the issues we've encountered with the Sam client library. Expectations may vary, but probably still the right thing to do :/

wchamber (approved):

README.md
```diff
@@ -23,18 +28,21 @@ terra
...
+- Add a user directly to the spend profile with `terra spend enable --policy=owner mmdevverily@gmail.com`
```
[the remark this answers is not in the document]
marikomedlock: Added a note about the difference.

README.md
marikomedlock: Yes. I think for the solutions team, we want to just give them admin so they don't need to ask us every time they want to use a new account. But for future users, probably not. Updated to default to member, but with instructions for admin also.

src/main/java/bio/terra/cli/command/groups/AddUser.java
marikomedlock: Agreed. Even more minor, but the varying capitalization of the parameter/option descriptions has also been bothering me.

It's not written down anywhere, yet. Feels like this should be low priority though -- work through functionality first, then we can talk about formatting later.

src/main/java/bio/terra/cli/command/groups/AddUser.java
marikomedlock: No, you don't need one. And you can run the group/spend commands without a workspace context.

That workspace context argument in the AuthManager constructor should go away once I get to the multiple identity contexts ticket. The pet SA key file fetch should not be happening during login anyway, and that's the only reason we need the workspace context in that class.

src/main/java/bio/terra/cli/command/groups/ListUsers.java
marikomedlock: You can be an admin but not a member, they're separate designations. Modified description.

src/main/java/bio/terra/cli/command/groups/RemoveUser.java
marikomedlock: done

src/main/java/bio/terra/cli/service/utils/SamService.java
marikomedlock: I would like to talk to the CI team about some of these things -- especially the pet SA key fetch. But I just don't have the bandwidth to follow up on them right now.

gjuggler (approved):

PR #16, opened 2021-02-18
PF-459: Remove manual environment variable substitution.

marikomedlock:
- Added a section in the README about workspace context for applications and how to use references.

- Removed manual environment variable substitution. For now we'll just escape the environment variable to bypass the host machine.

(If that syntax turns out to be too clunky, we can add this back later and make it less error-prone. For now, better to have it working reliably, especially since we're still not exactly sure about how invoking applications will look post solutions team + others feedback.)

### DIFF
--- a/src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
+++ b/src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
@@ -114,7 +114,15 @@ public class DockerAppsRunner {
     workspaceContext.requireCurrentWorkspace();
 
     // add the Terra references as environment variables in the container
-    envVars.putAll(buildMapOfTerraReferences());
+    Map<String, String> terraReferences = buildMapOfTerraReferences();
+    for (Map.Entry<String, String> workspaceReferenceEnvVar : terraReferences.entrySet()) {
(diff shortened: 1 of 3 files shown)


src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
```diff
@@ -113,11 +113,8 @@ public void runToolCommand(
     // check that the current workspace is defined
     workspaceContext.requireCurrentWorkspace();
 
-    // substitute any Terra references in the command
-    // also add the Terra references as environment variables in the container
-    Map<String, String> terraReferences = buildMapOfTerraReferences();
-    command = replaceTerraReferences(terraReferences, command);
-    envVars.putAll(terraReferences);
+    // add the Terra references as environment variables in the container
+    envVars.putAll(buildMapOfTerraReferences());
```
wchamber, line 117: optional: consider warning on env variable collision?

wchamber (approved):

gjuggler (approved):

src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
marikomedlock: Good point. Added.

PR #17, opened 2021-02-19
PF-399: Gradle tasks for pull, build, publish Docker image.

marikomedlock:
1. Added scripts for building and publishing a Docker image.
2. Added Gradle tasks that wrap those scripts and direct Docker commands: pullDockerImage, buildDockerImage, publishDockerImage.
3. Added instructions for building, pulling, publishing and updating the default image.
4. Changed the `tools/local-dev.sh` script to pull the default image from GCR instead of building it locally.

This is a temporary stand-in for an actual release process. But once we have GH actions or some automated way to do releases, we can have it call these scripts/tasks.

The main purpose of this PR is to un-block Nicole, who was having problems building the image locally on a Cloud Remote Desktop. The goal for this is to have CLI testers just run `source tools/local-dev.sh` and not have to build the Docker image locally anymore.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,0 +1,11 @@
+# default Docker image to use for launching applications
+# keep these properties in-sync with the DockerAppsRunner.DEFAULT_DOCKER_IMAGE_ID Java property
+# dockerGcrHostname/dockerGcrProject/dockerImageName:dockerImageTag
+dockerGcrHostname=gcr.io
+dockerGcrProject=terra-cli-dev
+dockerImageName=terra-cli/v0.0
+dockerImageTag=stable
+
+# name used when building the Docker image locally
+dockerLocalImageName=terra-cli/local
+dockerLocalImageTag=draft
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,6 +1,6 @@
 # default Docker image to use for launching applications
 # keep these properties in-sync with the DockerAppsRunner.DEFAULT_DOCKER_IMAGE_ID Java property
-# dockerGcrHostname/dockerGcrProject/dockerImageName:dockerImageTag
+# concatenate the below properties into a single path: dockerGcrHostname/dockerGcrProject/dockerImageName:dockerImageTag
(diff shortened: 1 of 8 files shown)


gradle.properties
```diff
@@ -0,0 +1,11 @@
+# default Docker image to use for launching applications
+# keep these properties in-sync with the DockerAppsRunner.DEFAULT_DOCKER_IMAGE_ID Java property
+# dockerGcrHostname/dockerGcrProject/dockerImageName:dockerImageTag
```
wchamber, line 3: nit: should be removed?

gradle.properties
```diff
@@ -0,0 +1,11 @@
+# default Docker image to use for launching applications
```
wchamber, line 1: just curious, no real opinion: any particular reason to define this in gradle.properties instead of as variables in gradle.build?

wchamber (approved):
thanks! documentation and PR description are both super helpful.

gradle.properties
marikomedlock: Do you mean line 3? No, that's showing how to concatenate the various properties into the full image path.

gradle.properties
marikomedlock: I had a vague sense that gradle.properties was the preferred approach. But also no strong opinion on my part, so if you know of a reason one way or the other, happy to change.

gradle.properties
wchamber: Yep, that wasn't clear to me, sorry.

gradle.properties
wchamber: I do not, was hoping you did :)

marikomedlock: Hah. okay Gradle docs recommend using a properties file. https://docs.gradle.org/current/userguide/organizing_gradle_projects.html#declare_properties_in_gradle_properties_file

gradle.properties
marikomedlock: Added comment about concatenation.

gjuggler (approved):

PR #18, opened 2021-02-22
PF-484: Store multiple pet SA key files per user.

marikomedlock:
1. pet SA key file path = [global context dir]/pet-keys/[user id]/[workspace id]
    * Each user now has a sub-directory of the global context directory for their pet SA key files.
    * Each pet SA key file is named for the workspace that it belongs to.
    * Only the user-specific sub-directory is mounted to the Docker image when running app commands. Previously, the entire pet-keys directory was mounted. The GOOGLE_APPLICATION_CREDENTIALS environment variable is still pointed to the pet SA key file for the current workspace.

2. Added existence checks to AuthenticationManager.loginTerraUser().
    * Skip the SAM fetches if the user information is already populated.
    * Only update the global context persisted on disk if the user has changed.

If you have a previous installation of the CLI, you'll need to wipe the global context directory so it can be re-generated with the updated format. `rm -R ~/.terra`

### DIFF
--- a/src/main/java/bio/terra/cli/context/GlobalContext.java
+++ b/src/main/java/bio/terra/cli/context/GlobalContext.java
@@ -153,9 +153,9 @@ public class GlobalContext {
   // ====================================================
   // Directory and file names
   //   - top-level directory: $HOME/.terra
-  //   - persisted global context file: global-context.json
-  //   - sub-directory for persisting pet SA keys: pet-keys/[terra user id]
-  //   - pet SA key filename: [workspace id]
+  //       - persisted global context file: global-context.json
+  //       - sub-directory for persisting pet SA keys: pet-keys/[terra user id]
+  //           - pet SA key filename: [workspace id]
 
   /** Getter for the global context directory. */
   public static Path resolveGlobalContextDir() {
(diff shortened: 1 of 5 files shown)


src/main/java/bio/terra/cli/context/GlobalContext.java
```diff
@@ -153,7 +154,8 @@ public void updateDockerImageId(String dockerImageId) {
   // Directory and file names
   //   - top-level directory: $HOME/.terra
   //   - persisted global context file: global-context.json
-  //   - sub-directory for persisting pet SA keys: pet-keys
+  //   - sub-directory for persisting pet SA keys: pet-keys/[terra user id]
+  //   - pet SA key filename: [workspace id]
```
wchamber, line 158: nit: maybe indent `- pet SA key filename` to make it clear it's under the sub-directory?

wchamber (approved):

src/main/java/bio/terra/cli/context/GlobalContext.java
marikomedlock: good idea. done.

gjuggler (approved):

PR #19, opened 2021-02-22
PF-445: Add flag for launching browser automatically or not.

marikomedlock:
- Added a boolean flag to the global context for the OAuth login flow.
    - true/auto = launch the browser window automatically and listen on a local port for the token response
    - false/manual = print the url to stdout, expect the user to open that url in a browser on any machine and copy/paste the token response to stdin
    - default = true/auto, which matches the existing behavior
- Added two new commands to manipulate this flag: `terra auth set-browser`, `terra auth get-browser`.

The purpose of this flag is to support logging in on a machine (e.g. CloudShell, notebook VM) that does not have a default (or any) browser available to it. Setting the flag to false/manual means the user can copy the url into a browser on a different machine (e.g. their laptop), confirm the scopes and get the token response, then copy/paste that back into a shell on the machine where they want to use the Terra CLI.

Example usage:
```
> terra auth set-browser manual
Browser will be launched manually (CHANGED)

> terra auth login
Please open the following address in a browser on any machine:
  https://accounts.google.com/o/oauth2/auth?access_type=offline&approval_prompt=force&client_id=[...]
Please enter code: *****
Login successful: mmdevverily@gmail.com
```

Side note: Long/medium-term, I'm not sure this is the right command hierarchy. It may be better to have a single `config` command grouping for things like this flag and the Docker image id. Not sure exactly which things should go there (server?), yet, but this seems lower priority to me for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,9 @@ terra
 2. `terra auth login` launches an OAuth flow that pops out a browser window with a warning login
 page ("! Google hasn't verified this app"). This shows up because the CLI is not yet a Google-verified
 app. Click through the warnings ("Advanced" -> "Go to ... (unsafe)") to complete the login.
+3. If the machine where you're running the CLI does not have a browser available to it, then use the
+manual login flow by setting the browser flag `terra auth set-browser manual`. See the [Authentication](#authentication)
+section below for more details.
 
 #### Spend profile access
 In order to spend money (e.g. by creating a project and resources within it) in Terra, you need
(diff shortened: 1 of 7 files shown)


gjuggler:
> Side note: Long/medium-term, I'm not sure this is the right command hierarchy. It may be better to have a single `config` command grouping for things like this flag and the Docker image id. Not sure exactly which things should go there (server?), yet, but this seems lower priority to me for now.

Thanks for the context! I agree on the need, and also the relatively low priority. IIRC `gcloud` and `kubectl` both have a pretty generic "set global config property" type setup for these kinds of things.

gjuggler (approved):

wchamber (approved):

wchamber:
> > Side note: Long/medium-term, I'm not sure this is the right command hierarchy. It may be better to have a single `config` command grouping for things like this flag and the Docker image id. Not sure exactly which things should go there (server?), yet, but this seems lower priority to me for now.
> 
> Thanks for the context! I agree on the need, and also the relatively low priority. IIRC `gcloud` and `kubectl` both have a pretty generic "set global config property" type setup for these kinds of things.

notebooks default location would be another config.

marikomedlock:
> > > Side note: Long/medium-term, I'm not sure this is the right command hierarchy. It may be better to have a single `config` command grouping for things like this flag and the Docker image id. Not sure exactly which things should go there (server?), yet, but this seems lower priority to me for now.
> > 
> > 
> > Thanks for the context! I agree on the need, and also the relatively low priority. IIRC `gcloud` and `kubectl` both have a pretty generic "set global config property" type setup for these kinds of things.
> 
> notebooks default location would be another config.

Added a ticket to track this: https://broadworkbench.atlassian.net/browse/PF-488

PR #21, opened 2021-02-24
Add '--' prefixes to option parameters of notebooks commands.

wchamber:
This will make the notebooks commands consistent with the other commands, which should help usability.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@
     * [Workspace](#workspace)
     * [Resources](#resources)
     * [Applications](#applications)
+    * [Notebooks](#notebooks)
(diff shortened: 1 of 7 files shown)


marikomedlock (approved):
lgtm. Could you also add a short section on notebooks commands to the README? Doesn't have to be in this PR.

wchamber:
> lgtm. Could you also add a short section on notebooks commands to the README? Doesn't have to be in this PR.

added one, PTAL. Feel too close to it to provide much meaningful info. Would be cool to have this autogenerated one day, very low priority though.

gjuggler (approved):

PR #22, opened 2021-02-24
PF-421: Change project id env var to GOOGLE_CLOUD_PROJECT.

marikomedlock:
- Changed the name of the environment variable that we use to store the workspace project id, from `TERRA_GOOGLE_PROJECT_ID` -> `GOOGLE_CLOUD_PROJECT`. This new environment variable overrides the project specified in the `nextflow.config` file and is also documented in the Google Cloud Java client library [README](https://github.com/googleapis/google-cloud-java/blob/master/README.md#specifying-a-project-id), so it seems like the better choice here (similar to how GOOGLE_APPLICATION_CREDENTIALS is recognized by different apps).
- Published a new Docker image and bumped the default version used by the CLI.
- Updated the instructions for publishing a new Docker image to include authentication. (I think I didn't need to do this before because I happened to be logged in with a user that had permissions.)
- Added a new `--default` flag to the `terra app set-image` command, so that we have a way to set the default image. Called this new command in the `tools/local-dev.sh` script so we don't have to alert everyone every time we publish a new Docker image. Instead, running `source tools/local-dev.sh` should update their default.
- Replaced underscores with hyphens in the "controlled resource" bucket creation command. I think this is in line with "help users do the right thing", since Nextflow requires bucket paths to be valid URLs and according to the solutions team they've run into other tools with the same requirement. The allowed set of characters for resource names is still alphanumeric characters and underscores, which I think makes sense because users will type the environment variables directly in their app code/files, so better to have the user-facing names all be consistent with underscores. This will eventually need to be a discussion with WSM/larger team on how to avoid collisions in resource naming.
```
> terra resources describe --name=my_bucket
Name: my_bucket
Type: bucket
Cloud Id: gs://terra-wsm-dev-e844792d-my-bucket
> terra app execute echo \${TERRA_MY_BUCKET}
[...]
gs://terra-wsm-dev-e844792d-my-bucket
```

These changes came from debugging Nextflow on a WSM-provided project (notes are [here](https://docs.google.com/document/d/1Wqa5ssHpp8M--RPCbD_fibEs0gn98XaFGngxzH-m1h4/edit#bookmark=id.puo7xa9tdmuv) for reference).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -38,7 +38,7 @@ For any change in this directory to take effect:
     ```
 2. Update the image id that the CLI uses. (See output of previous command for image name and tag.)
     ```
-    > terra app set-image terra-cli/local:b5fdce0
+    > terra app set-image --image=terra-cli/local:b5fdce0
     ```
 
 #### Pull an existing image
@@ -61,7 +61,7 @@ To use a specific Docker image from GCR:
     ```
 2. Update the image id that the CLI uses. (See output of previous command for image name and tag.)
     ```
-    > terra app set-image gcr.io/terra-cli-dev/terra-cli/v0.0:b5fdce0
+    > terra app set-image --image=gcr.io/terra-cli-dev/terra-cli/v0.0:b5fdce0
     ```
 
 #### Publish a new image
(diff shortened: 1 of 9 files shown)


gjuggler (approved):

wchamber (approved):
Sorry for the slow review, I missed this one. Please don't hesitate to ping me if I haven't gotten back to you in ~business day, it probably got lost from my todo list and I won't mind the reminder even if it hasn't  :)

PR #23, opened 2021-02-26
PF-427 Search parent directories for context file. Preserve relative path when running in container.

marikomedlock:
- Recursively search the current directory hierarchy for the workspace context file.
- Preserve the current directory, relative to the workspace directory, when running in the container.
- Cleanup the file and directory helper methods for handling the global context and workspace.
- Mount the global context directory and the workspace directory onto the Docker container.

The main goal for these changes is to use any `terra` commands, including those that start a Docker container, in sub-directories of the workspace directory. Previously, you could only use `terra` commands in the top-level workspace directory -- changing to a sub-directory would fail to find the workspace context.
```
> terra workspace create
Workspace successfully created: 03935501-8978-4188-a58a-3fb3d876fd35

> git clone https://github.com/nextflow-io/rnaseq-nf.git
[...git output...]

> cd rnaseq-nf

# command that does not start a container
> terra resources create --type=bucket --name=mybucket
bucket successfully created: gs://terra-wsm-dev-19c49e13-mybucket
Workspace resource successfully added: mybucket

# command that does start a container
> terra nextflow config main.nf
[...nextflow output...]
```

A lesser goal of these changes is to mount the entire global and workspace context directories onto the Docker container, in preparation for installing the Terra CLI on the container. (I needed to mount the workspace directory onto the container for these changes anyway, so I went ahead and mounted the global context directory while I was there.) Installing the Terra CLI on the container would let users include things like `$(terra data-refs resolve my_bucket)` in a bash script called by their app.

### DIFF
--- a/src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
+++ b/src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
@@ -49,7 +49,7 @@ public class DockerAppsRunner {
   public DockerAppsRunner(GlobalContext globalContext, WorkspaceContext workspaceContext) {
     this.globalContext = globalContext;
     this.workspaceContext = workspaceContext;
-    this.dockerClient = null;
+    this.dockerClient = buildDockerClient();
   }
 
   // ====================================================
@@ -72,8 +72,6 @@ public class DockerAppsRunner {
    * @return true if the Docker image property was updated, false otherwise
    */
   public boolean updateImageId(String imageId) {
-    buildDockerClient();
-
     // check if image exists
     try {
       dockerClient.inspectImageCmd(imageId).exec();
@@ -153,10 +151,7 @@ public class DockerAppsRunner {
 
     // create and start the docker container. run the terra_init script first, then the given
     // command
-    buildDockerClient();
-    String containerId =
-        startDockerContainerWithTerraInit(
-            command, getWorkingDirOnContainer().toString(), envVars, bindMounts);
+    String containerId = startDockerContainerWithTerraInit(command, envVars, bindMounts);
 
     // read the container logs, which contains the command output, and write them to stdout
     outputLogsForDockerContainer(containerId);
@@ -199,14 +194,13 @@ public class DockerAppsRunner {
    * terra_init script because it tries to read the pet key file.
    *
    * @param command the full string command to execute in a bash shell (bash -c ..cmd..)
-   * @param workingDir the directory where the commmand will be executed
    * @param envVars a mapping of environment variable names to values
    * @param bindMounts a mapping of container mount point to the local directory being mounted
    * @throws RuntimeException if an environment variable or bind mount used by the terra_init script
    *     overlaps or conflicts with one passed into this method
    */
   private String startDockerContainerWithTerraInit(
-      String command, String workingDir, Map<String, String> envVars, Map<Path, Path> bindMounts) {
+      String command, Map<String, String> envVars, Map<Path, Path> bindMounts) {
     // check that there is a current user, because the terra_init script will try to read the pet
     // key file
     TerraUser currentUser = globalContext.requireCurrentTerraUser();
@@ -232,21 +226,22 @@ public class DockerAppsRunner {
     }
     envVars.putAll(terraInitEnvVars);
 
-    return startDockerContainer(fullCommand, workingDir, envVars, bindMounts);
+    return startDockerContainer(
+        fullCommand, getWorkingDirOnContainer().toString(), envVars, bindMounts);
   }
 
   // ====================================================
   // Docker containers
 
   /** Build the Docker client object with standard options. */
-  private void buildDockerClient() {
+  private static DockerClient buildDockerClient() {
     DockerClientConfig config = DefaultDockerClientConfig.createDefaultConfigBuilder().build();
     DockerHttpClient httpClient =
         new ApacheDockerHttpClient.Builder()
             .dockerHost(config.getDockerHost())
             .sslConfig(config.getSSLConfig())
             .build();
-    dockerClient = DockerClientImpl.getInstance(config, httpClient);
+    return DockerClientImpl.getInstance(config, httpClient);
   }
 
   /**
--- a/src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
+++ b/src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
@@ -21,7 +21,6 @@ import com.github.dockerjava.transport.DockerHttpClient;
 import java.io.File;
 import java.nio.charset.Charset;
 import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
@@ -40,14 +39,8 @@ public class DockerAppsRunner {
   private final WorkspaceContext workspaceContext;
   private DockerClient dockerClient;
 
-  // mount point for the global context directory on the container
-  // TODO: this is the default $HOME directory on the container, which is where we look for the
-  // global context directory. add an env var override for this, so we can mount the global context
-  // somewhere else (e.g. under /usr)
-  private static final String GLOBAL_CONTEXT_MOUNT_POINT = "/root/.terra";
-
-  // mount point for the workspace context directory on the container
-  private static final String WORKSPACE_CONTEXT_MOUNT_POINT = "/usr/local/etc/.terra";
+  private static final String CONTAINER_HOME_DIR = "/root";
+  private static final String CONTAINER_WORKSPACE_DIR = "/usr/local/etc";
 
   public DockerAppsRunner(GlobalContext globalContext, WorkspaceContext workspaceContext) {
     this.globalContext = globalContext;
@@ -126,25 +119,33 @@ public class DockerAppsRunner {
     }
     envVars.putAll(terraReferences);
 
-    // mount the global and workspace directories to the Docker image
-    File workspaceDir = WorkspaceContext.resolveWorkspaceDir();
+    // mount the global context directory and the workspace directory to the container
+    //  e.g. global context dir on host = $HOME/.terra
+    //       global context dir on container = CONTAINER_HOME_DIR/.terra
+    //       workspace dir on host = $HOME/terra-workspaces/workspace123
+    //       workspace dir on container = CONTAINER_WORKSPACE_DIR
+    Path globalContextDirOnContainer =
+        GlobalContext.getGlobalContextDir(Path.of(CONTAINER_HOME_DIR));
+    Path workspaceDirOnContainer = Path.of(CONTAINER_WORKSPACE_DIR);
+
     Map<String, File> bindMounts = new HashMap<>();
-    bindMounts.put(GLOBAL_CONTEXT_MOUNT_POINT, GlobalContext.resolveGlobalContextDir().toFile());
-    bindMounts.put(WORKSPACE_CONTEXT_MOUNT_POINT, workspaceDir);
+    bindMounts.put(
+        globalContextDirOnContainer.toString(), GlobalContext.getGlobalContextDir().toFile());
+    bindMounts.put(workspaceDirOnContainer.toString(), WorkspaceContext.getWorkspaceDir());
 
-    // default the working directory to the current directory, relative to the workspace directory
-    Path currentDir = Paths.get("").toAbsolutePath();
-    Path currentRelativeToWorkspaceDir =
-        WorkspaceContext.resolveWorkspaceDir().toPath().toAbsolutePath().relativize(currentDir);
-    String workingDir =
-        Path.of(WORKSPACE_CONTEXT_MOUNT_POINT).resolve(currentRelativeToWorkspaceDir).toString();
-
-    buildDockerClient();
+    // set the working directory on the container to match the working directory on the host,
+    // relative to the workspace directory
+    //  e.g. current dir on host = $HOME/terra-workspaces/workspace123/nextflow/rnaseq-nf
+    //       current dir on container = CONTAINER_WORKSPACE_DIR/nextflow/rnaseq-nf
+    Path workingDir =
+        Path.of(CONTAINER_WORKSPACE_DIR)
+            .resolve(WorkspaceContext.getCurrentDirRelativeToWorkspaceDir());
 
     // create and start the docker container. run the terra_init script first, then the given
     // command
+    buildDockerClient();
     String containerId =
-        startDockerContainerWithTerraInit(command, workingDir, envVars, bindMounts);
+        startDockerContainerWithTerraInit(command, workingDir.toString(), envVars, bindMounts);
 
     // read the container logs, which contains the command output, and write them to stdout
     outputLogsForDockerContainer(containerId);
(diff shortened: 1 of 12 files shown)


src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
```diff
@@ -124,12 +122,33 @@ public void runToolCommand(
...
+    // mount the global context directory and the workspace directory to the container
+    //  e.g. global context dir on host = $HOME/.terra
+    //       global context dir on container = CONTAINER_HOME_DIR/.terra
+    //       workspace dir on host = $HOME/terra-workspaces/workspace123
+    //       workspace dir on container = CONTAINER_WORKSPACE_DIR
+    Path globalContextDirOnContainer =
+        GlobalContext.getGlobalContextDir(Path.of(CONTAINER_HOME_DIR));
+    Path workspaceDirOnContainer = Path.of(CONTAINER_WORKSPACE_DIR);
+
+    Map<String, File> bindMounts = new HashMap<>();
```
wchamber, line 134: optional style: keep the keys as `Path`s instead of `String`s to preserve some meaning until they're turned into `Volume`s.

src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
```diff
@@ -124,12 +122,33 @@ public void runToolCommand(
...
     envVars.putAll(terraReferences);
 
-    buildDockerClient();
+    // mount the global context directory and the workspace directory to the container
+    //  e.g. global context dir on host = $HOME/.terra
+    //       global context dir on container = CONTAINER_HOME_DIR/.terra
+    //       workspace dir on host = $HOME/terra-workspaces/workspace123
+    //       workspace dir on container = CONTAINER_WORKSPACE_DIR
+    Path globalContextDirOnContainer =
+        GlobalContext.getGlobalContextDir(Path.of(CONTAINER_HOME_DIR));
```
wchamber, line 131: I'm a little confused by the paths. We use `getGlobalContextDir()` for the host global context dir, but `getGlobalContextDir(Path relativeToDir)` to get the right container path? Similarly for `getPetSaKeyFile`'s two overloads? Should the relativeToDir overloads just be `getGlobalContextDirOnContainer`?
It feels like the path logic right now is split between this class and its `CONTAINER_` vars and the `GlobalContext`.

src/main/java/bio/terra/cli/context/GlobalContext.java
```diff
@@ -170,43 +165,82 @@ public void updateDockerImageId(String dockerImageId) {
...
-    return resolveGlobalContextDir().resolve(PET_KEYS_DIRNAME);
+  @JsonIgnore
+  public static Path getGlobalContextDir(Path relativeToDir) {
+    return relativeToDir.resolve(GLOBAL_CONTEXT_DIRNAME).toAbsolutePath();
   }
 
   /**
-   * Getter for the sub-directory of the global context directory that holds the pet SA key files
-   * for the given user.
+   * Get a handle to the global context file.
```
wchamber, line 192: what's the difference between a file and a file handle?

src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
wchamber: spitballing suggestions, maybe relegate all the file logic to a single class? And maybe there's an instance of the class for host & one for the container? Right now there's some file path logic in both the GlobalContext & WorkspaceContext. I don't know if it makes sense to take file path resolution out of there.

gjuggler: +1 on being confused here. I added a comment on the GlobalContext method itself, but IMO removing the "relativeTo" parameter and using explicit `getFooOnContainer` and `getFooOnHost` variants, as Will suggested, should help a lot.

src/main/java/bio/terra/cli/context/GlobalContext.java
```diff
@@ -170,43 +165,82 @@ public void updateDockerImageId(String dockerImageId) {
...
+  public static Path getGlobalContextDir() {
+    // TODO: allow overriding the global context directory path (e.g. env var != user home
+    // directory)
+    return getGlobalContextDir(Paths.get(System.getProperty("user.home")));
   }
 
   /**
-   * Getter for the sub-directory of the global context directory that holds the pet SA key files
-   * for all users.
+   * Get the global context directory path relative to the given directory.
```
gjuggler, line 181: This Javadoc and the method signature are confusing. It would help to give an example or two, and possibly be more precise with expectations around relative vs. absolute paths on the input and output.

(That said, Will's suggestion to adjust the approach and use explicit overrides, rather than a generic "relativeToDir" param, will obviate most of the need for documentation here.)

src/main/java/bio/terra/cli/context/GlobalContext.java
gjuggler: There's an interesting challenge her in distinguishing between 3 possible return types (`Path`, `File`, and `String`). IMO the Javadoc should state a preference and/or justification for which return values are preferred and why, e.g. "All methods dealing with directory or file paths return a Path, unless otherwise indicated with a `String` or `File` suffix."

For increased consistency, I wonder if converting all File return-type methods to return a Path instead might help reduce the mental burden on users of the class. Asking users to call .toFile() when they want to get a File isn't too much of an inconvenience, especially if it isn't obvious to this class which one is preferred.

src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
```diff
@@ -38,8 +39,11 @@
   private final WorkspaceContext workspaceContext;
   private DockerClient dockerClient;
 
-  // This is where the pet key files will be mounted on the Docker container.
-  public static final String PET_KEYS_MOUNT_POINT = "/usr/local/etc/terra_cli";
+  // default $HOME directory on the container (this is where we expect to look for the global
+  // context)
+  private static final String CONTAINER_HOME_DIR = "/root";
+  // mount point for the workspace directory
```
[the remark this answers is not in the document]
marikomedlock: I don't have strong opinions about comment style, so that change sounds fine to me. However, I prefer consistency within the codebase, and changing this here would make it inconsistent with how I've done comments throughout. Perhaps add a new ticket with any style changes you'd like to see and we can take care of them all at once in a separate PR.

src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
marikomedlock: Good idea. Done.

gjuggler (commented):
Mostly similar to Will's comments, with some add'l thoughts on method return types.

Non-blocking for this PR, but the addition of recursive directory search definitely makes the lack of unit tests start to feel more glaring. Let's discuss sometime how we'll approach prioritizing test infra against the rest of the roadmap.

src/main/java/bio/terra/cli/context/GlobalContext.java
marikomedlock: I usually think of a file handle as the object wrapper (e.g. Java File object) around a (filesystem) file descriptor. But I wasn't attempting to make any particular distinction here, so I'll remove the "handles" to avoid confusion.

src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
marikomedlock: Yeah, this is the part I really struggled with. I wanted to keep the structure of the global and workspace context directories in those classes, but ran into trouble when trying to map that to the container paths.

I'm hesitant to split this class in two because almost all the code just uses the host machine version of the global context. So it feels like the container paths are a special case. I'll take a stab at moving the container-specific logic into the DockerAppsRunner class.

src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
```diff
@@ -124,12 +122,33 @@ public void runToolCommand(
...
+    Path workspaceDirOnContainer = Path.of(CONTAINER_WORKSPACE_DIR);
+
+    Map<String, File> bindMounts = new HashMap<>();
+    bindMounts.put(
+        globalContextDirOnContainer.toString(), GlobalContext.getGlobalContextDir().toFile());
+    bindMounts.put(workspaceDirOnContainer.toString(), WorkspaceContext.getWorkspaceDir());
+
+    // set the working directory on the container to match the working directory on the host,
+    // relative to the workspace directory
+    //  e.g. current dir on host = $HOME/terra-workspaces/workspace123/nextflow/rnaseq-nf
```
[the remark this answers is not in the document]
marikomedlock: No, just inconsistent. Thanks for the catch. Changed to "working"

src/main/java/bio/terra/cli/context/GlobalContext.java
marikomedlock: Good point about the varying return types. I'll give it a pass to make them more consistent on Path.

marikomedlock: Changed all file/directory related methods in GlobalContext and WorkspaceContext to use Path as return type.

src/main/java/bio/terra/cli/context/GlobalContext.java
marikomedlock: Replaced this method with logic in the DockerAppsRunner class, and included examples there.

src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
marikomedlock: I took another pass at this. PTAL to see if this seems more/less logical.
- Left all logic about the structure of the context directories in the Global/WorkspaceContext classes.
- Added helper methods to the DockerAppsRunner class to convert paths on the host machine to paths on the container.

marikomedlock:
> Non-blocking for this PR, but the addition of recursive directory search definitely makes the lack of unit tests start to feel more glaring. Let's discuss sometime how we'll approach prioritizing test infra against the rest of the roadmap.

Agreed about needing tests. My first thought is that install should come first, but after that I'm less sure of priorities.

src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
```diff
@@ -51,6 +52,9 @@ public DockerAppsRunner(GlobalContext globalContext, WorkspaceContext workspaceC
     this.dockerClient = null;
   }
 
+  // ====================================================
+  // Docker images
```
wchamber, line 56: not this PR, but when it starts getting useful to provide sub-class structure, it can be a sign that it's worth breaking this out into helper classes and utilities.

src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
```diff
@@ -218,6 +235,20 @@ private String startDockerContainerWithTerraInit(
...
+
+  /** Build the Docker client object with standard options. */
+  private void buildDockerClient() {
+    DockerClientConfig config = DefaultDockerClientConfig.createDefaultConfigBuilder().build();
+    DockerHttpClient httpClient =
+        new ApacheDockerHttpClient.Builder()
+            .dockerHost(config.getDockerHost())
+            .sslConfig(config.getSSLConfig())
+            .build();
+    dockerClient = DockerClientImpl.getInstance(config, httpClient);
```
wchamber, line 249: WDYT about returning the dockerClient from this function? If it's initialized here and not valid to use before calling this function, returning the dockerClient and using it as a local variable instead of a class member makes ordering  dependency clear. Or move it to the DockerAppsRunner constructor?

src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
```diff
@@ -166,10 +206,7 @@ private void buildDockerClient() {
    *     overlaps or conflicts with one passed into this method
    */
   private String startDockerContainerWithTerraInit(
-      String command,
-      String workingDir,
-      Map<String, String> envVars,
-      Map<String, File> bindMounts) {
+      String command, String workingDir, Map<String, String> envVars, Map<Path, Path> bindMounts) {
```
wchamber, line 209: it doesn't look like `workingDir` ever changes based on arguments. Instead of passing it through this and `startDockerContainer`, can we call `getWorkingDirOnContainer()` in `startDockerContainer`?

src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
```diff
@@ -222,10 +259,7 @@ private String startDockerContainerWithTerraInit(
    * @throws RuntimeException if the local directory does not exist or is not a directory
    */
   private String startDockerContainer(
```
wchamber, line 261: Should this only ever be called from `startDockerContainerWithTerraInit`?

src/main/java/bio/terra/cli/context/WorkspaceContext.java
```diff
@@ -189,18 +186,85 @@ public void removeCloudResource(String name) {
...
+    try {
+      return getExistingWorkspaceContextFile(currentDir);
+    } catch (FileNotFoundException fnfEx) {
+      return currentDir.resolve(WORKSPACE_CONTEXT_DIRNAME).resolve(WORKSPACE_CONTEXT_FILENAME);
+    }
+  }
 
   /**
-   * Getter for the file where the workspace context is persisted.
+   * Get a handle for an existing workspace context file in the current directory hierarchy.
```
wchamber, line 228: does this return an absolute path or a relative path?

src/main/java/bio/terra/cli/context/WorkspaceContext.java
```diff
@@ -189,18 +186,85 @@ public void removeCloudResource(String name) {
...
+      value = "NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE",
+      justification =
+          "An NPE would only happen here if there was an error getting the workspace context file, and an exception would be thrown in that method instead.")
+  @JsonIgnore
+  public static Path getWorkspaceDir() {
+    return getWorkspaceContextFile().getParent().getParent();
+  }
+
+  /**
+   * Get a handle for the workspace context file.
```
wchamber, line 206: handles or paths? is it absolute or relative to the current directory?

src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
```diff
@@ -78,6 +86,24 @@ public boolean updateImageId(String imageId) {
...
+   * Utility method for concatenating a command and its arguments.
+   *
+   * @param cmd command name (e.g. gsutil)
+   * @param cmdArgs command arguments (e.g. ls, gs://my-bucket)
+   */
+  public static String buildFullCommand(String cmd, List<String> cmdArgs) {
+    String fullCommand = cmd;
+    if (cmdArgs != null && cmdArgs.size() > 0) {
+      final String argSeparator = " ";
+      fullCommand += argSeparator + String.join(argSeparator, cmdArgs);
```
wchamber, line 102: very optional: using streams for this join `Stream.concat(Stream.of(cmd), cmdArgs.toStream()).collect(Collectors.joining(" "));`

wchamber (approved):

src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
marikomedlock: Agreed. I think breaking out the Docker images and container sections into a helper class could be a way to do this. Will keep in mind for future clean-up work.

src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
marikomedlock: Good idea. Moved it to the constructor.

src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
marikomedlock: Yes, done.

gjuggler (approved):

src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
marikomedlock: Yes. I was just separating out the Docker-specific part of starting a container from the Terra CLI-specific part. Per our conversation above about moving some of these methods into a helper class, I think the startDockerContainer would go into a DockerHelper class, and the ..WithTerraInit would stay here.

src/main/java/bio/terra/cli/context/WorkspaceContext.java
marikomedlock: Absolute. Updated comment.

src/main/java/bio/terra/cli/context/WorkspaceContext.java
marikomedlock: Ah I missed these references to handles, thanks for the catch. Updated comment to indicate absolute.

src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
marikomedlock: I prefer avoiding object creation unless it makes the code a lot clearer or faster somehow.

PR #24, opened 2021-03-01
PF-508: Invite user if workspace add-user returns bad request.

marikomedlock:
- Propagate exceptions thrown by the Workspace Manager client up to the caller. Previously, they were logged by otherwise failed silently.
- The only exception to this error propagation are calls to the `/status` endpoint, where an exception is interpreted by the CLI the same as a bad status.
- If `terra workspace add-user` fails with a bad request (i.e. code=400), then try to invite them first and then add them to the workspace.

In the future, this invite behavior might be better placed in WSM -- I'm not sure of that, but I've put it on the list of possible requests of WSM from the CLI.

### DIFF
--- a/src/main/java/bio/terra/cli/service/utils/DataRepoService.java
+++ b/src/main/java/bio/terra/cli/service/utils/DataRepoService.java
@@ -84,10 +84,8 @@ public class DataRepoService {
     UnauthenticatedApi unauthenticatedApi = new UnauthenticatedApi(apiClient);
     try {
       return unauthenticatedApi.serviceStatus();
-    } catch (Exception ex) {
-      // catch any exception, including client-side ones (e.g. UnknownHostException)
-      logger.error("Error getting Data Repo status", ex);
-      return null;
+    } catch (ApiException ex) {
+      throw new RuntimeException("Error getting Data Repo status", ex);
     }
   }
 }
--- a/src/main/java/bio/terra/cli/service/utils/DataRepoService.java
+++ b/src/main/java/bio/terra/cli/service/utils/DataRepoService.java
@@ -4,6 +4,7 @@ import bio.terra.cli.context.ServerSpecification;
 import bio.terra.cli.context.TerraUser;
 import bio.terra.datarepo.api.UnauthenticatedApi;
 import bio.terra.datarepo.client.ApiClient;
+import bio.terra.datarepo.client.ApiException;
 import bio.terra.datarepo.model.RepositoryConfigurationModel;
 import bio.terra.datarepo.model.RepositoryStatusModel;
 import com.google.auth.oauth2.AccessToken;
@@ -68,8 +69,8 @@ public class DataRepoService {
     RepositoryConfigurationModel repositoryConfig = null;
     try {
       repositoryConfig = unauthenticatedApi.retrieveRepositoryConfig();
-    } catch (Exception ex) {
-      logger.error("Error getting Data Repo configuration", ex);
+    } catch (ApiException ex) {
+      throw new RuntimeException("Error getting Data Repo version", ex);
     }
     return repositoryConfig;
   }
@@ -81,12 +82,12 @@ public class DataRepoService {
    */
   public RepositoryStatusModel getStatus() {
     UnauthenticatedApi unauthenticatedApi = new UnauthenticatedApi(apiClient);
-    RepositoryStatusModel status = null;
     try {
-      status = unauthenticatedApi.serviceStatus();
+      return unauthenticatedApi.serviceStatus();
     } catch (Exception ex) {
+      // catch any exception, including client-side ones (e.g. UnknownHostException)
       logger.error("Error getting Data Repo status", ex);
+      return null;
     }
-    return status;
   }
 }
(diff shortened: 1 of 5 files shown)


src/main/java/bio/terra/cli/service/utils/DataRepoService.java
```diff
@@ -81,12 +82,12 @@ public RepositoryConfigurationModel getVersion() {
...
   public RepositoryStatusModel getStatus() {
     UnauthenticatedApi unauthenticatedApi = new UnauthenticatedApi(apiClient);
-    RepositoryStatusModel status = null;
     try {
-      status = unauthenticatedApi.serviceStatus();
+      return unauthenticatedApi.serviceStatus();
     } catch (Exception ex) {
+      // catch any exception, including client-side ones (e.g. UnknownHostException)
       logger.error("Error getting Data Repo status", ex);
+      return null;
```
wchamber, line 90: consider adding a comment about why it's helpful to swallow the getStatus exception: to allow returning info from the other exceptions. Or, throw the exception and handle the swallowing higher up for the status command specific case. ditto other status endpoints. Or, encode it in the return value of each status endpoint, since we want to treat api errors as equivalent to bad statuses.

src/main/java/bio/terra/cli/service/utils/WorkspaceManagerService.java
```diff
@@ -214,14 +220,52 @@ public void deleteWorkspace(UUID workspaceId) {
...
+        UserStatusDetails userStatusDetails =
+            new SamService(server, terraUser).inviteUser(userEmail);
+        logger.info("invited new user: {}", userStatusDetails);
+
+        // now try to add the user to the workspace role
+        // retry if it returns with a bad request (because the invite sometimes takes a few seconds
+        // to propagate -- not sure why)
+        HttpUtils.callWithRetries(
+            () -> {
+              workspaceApi.grantRole(grantRoleRequestBody, workspaceId, iamRole);
```
wchamber, line 245: just curious, I think what's here is fine. Sam calls are always wrapped in retries, but not WM API calls, except for this case. Is Sam just unreliable and WM reliable?

wchamber (approved):

src/main/java/bio/terra/cli/service/utils/WorkspaceManagerService.java
marikomedlock: - I wrapped this case in retries because when I first tested this code, it fails on the first grantRole call after inviting a user, but succeeds a few seconds later. So it was to spot fix this particular case.
- I think any time we're talking to a service, we should use retries, including for WM. I just haven't added it yet because, as you say, it's been much more reliable than SAM, so it's not been a problem. I added a [ticket](https://broadworkbench.atlassian.net/browse/PF-513) so we don't forget about this. Also probably good to check with WM team what errors should be retryable.

src/main/java/bio/terra/cli/service/utils/DataRepoService.java
marikomedlock: Moved the exception swallowing up one level into ServerManager, so that the *Service classes don't have any CLI-logic in them. Also added a comment about the reason for this logic.

src/main/java/bio/terra/cli/service/utils/SamService.java
```diff
@@ -81,16 +82,32 @@ private void buildClientForTerraUser() {
...
+  }
+
+  /**
+   * Call the SAM "/api/users/v1/invite/{inviteeEmail}" endpoint to invite a user and track them.
+   * This is not the same thing as registering a user.
+   *
+   * @param userEmail email to invite
+   * @return SAM object with details about the user
+   */
+  public UserStatusDetails inviteUser(String userEmail) {
```
jaycarlton, line 101: does this actually communicate with the user? Is there an email template or something we need to set up?

src/main/java/bio/terra/cli/service/utils/WorkspaceManagerService.java
```diff
@@ -172,20 +177,19 @@ public WorkspaceDescription createWorkspace() {
    */
   public WorkspaceDescription getWorkspace(UUID workspaceId) {
     WorkspaceApi workspaceApi = new WorkspaceApi(apiClient);
-    WorkspaceDescription workspaceWithContext = null;
     try {
       // fetch the Terra workspace object
-      workspaceWithContext = workspaceApi.getWorkspace(workspaceId);
+      WorkspaceDescription workspaceWithContext = workspaceApi.getWorkspace(workspaceId);
       String googleProjectId =
           (workspaceWithContext.getGoogleContext() == null)
```
jaycarlton, line 184: nit: I find Optional chains easier to read than the ternary operator.

jaycarlton (approved):

src/main/java/bio/terra/cli/service/utils/SamService.java
marikomedlock: My understanding (from @wchamber on [Slack](https://broadinstitute.slack.com/archives/C016V54EK0B/p1614371604039900) and through testing with throwaway gmail accounts) is that there is no communication, it just gives them a subject id in SAM so that we can refer to them in the system. Then when they do register, SAM will connect them with this same subject id. So the effect is that we can "share" a workspace with a user before they actually register.

src/main/java/bio/terra/cli/service/utils/WorkspaceManagerService.java
marikomedlock: I don't have super strong feelings about Optional, but prefer to avoid the creation of objects unless it's strictly necessary.

PR #25, opened 2021-03-03
PF-483: Add data-refs commands.

marikomedlock:
- Added `terra data-refs` commands: `list`, `add`, `delete`, `check-access`, `resolve`.
- Included the user's proxy group email in the TerraUser object and the output of the `terra auth status` command. This is so that it's easier to look up the email that you need to share a bucket with.
```
terra auth status
Current Terra user: mmdevverily@gmail.com
Current Terra user's proxy group: PROXY_110017243614237806241@dev.test.firecloud.org
LOGGED IN
```
- The `check-access` command (currently only for buckets) checks for `objects.list` permission. I'm not sure if this is the right thing long-term -- perhaps we need flags for different access checks (e.g. `terra check-access --writer`)? -- but this is just a guess at what might be useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@
     * [Server](#server)
     * [Workspace](#workspace)
     * [Resources](#resources)
+    * [Data References](#data-references)
     * [Applications](#applications)
     * [Notebooks](#notebooks)
     * [Groups](#groups)
@@ -125,6 +126,7 @@ Commands:
   server     Connect to a Terra server.
   workspace  Setup a Terra workspace.
   resources  Manage controlled resources in the workspace.
+  data-refs  Reference data in the workspace context.
   app        Run applications in the workspace.
   notebooks  Use AI Notebooks in the workspace.
   groups     Manage groups of users.
@@ -138,7 +140,9 @@ Each sub-group of commands is described in a sub-section below:
 - Server
 - Workspace
 - Resources
+- Data References
 - Applications
+- Notebooks
 - Groups
 - Spend
 
@@ -226,6 +230,23 @@ Commands:
 
 A controlled resource is a cloud resource managed by the Terra workspace on behalf of the user.
 Currently, the only supported controlled resource is a bucket.
+
+#### Data References
+```
+Usage: terra data-refs [COMMAND]
+Reference data in the workspace context.
+Commands:
+  list          List all data references.
+  add           Add a new data reference.
+  delete        Delete an existing data reference.
+  check-access  Check if you have access to a data reference.
+  resolve       Resolve a data reference to its cloud id or path.
+```
+
+A data reference points to a cloud resource that the user wants to read or write data to.
+Data references include some types of controlled resources (e.g. bucket but not VM) and 
+un-controlled resources that are external to the workspace.
+Currently, the only supported data reference is a GCS bucket.
 
 #### Applications
 ```
--- a/README.md
+++ b/README.md
@@ -243,9 +243,9 @@ Commands:
   resolve       Resolve a data reference to its cloud id or path.
 ```
 
-A data reference points to a cloud resource that the user wants to read or write data to.
-Data references include some types of controlled resources (e.g. bucket but not VM) and 
-un-controlled resources that are external to the workspace.
+A data reference points to a data (i.e. not compute) type of cloud resource that the user wants to
+read or write to. Data references include some types of controlled resources (e.g. bucket but not
+VM) within the workspace and un-controlled resources external to the workspace.
 Currently, the only supported data reference is a GCS bucket.
 
 #### Applications
--- a/src/main/java/bio/terra/cli/service/WorkspaceManager.java
+++ b/src/main/java/bio/terra/cli/service/WorkspaceManager.java
@@ -303,6 +303,8 @@ public class WorkspaceManager {
    *
    * @param referenceType type of reference to add
    * @param referenceName name of reference to add
+   * @param cloudId unique identifier for this resource in the cloud (e.g. bucket uri, bq dataset
+   *     id)
    * @return the data reference that was added
    */
   public CloudResource addDataReference(
--- a/src/main/java/bio/terra/cli/service/WorkspaceManager.java
+++ b/src/main/java/bio/terra/cli/service/WorkspaceManager.java
@@ -185,6 +185,9 @@ public class WorkspaceManager {
   public CloudResource getControlledResource(String resourceName) {
     // TODO: change this method to call WSM controlled resource endpoints once they're ready
     CloudResource cloudResource = workspaceContext.getCloudResource(resourceName);
+    if (cloudResource == null) {
+      throw new RuntimeException(resourceName + " not found.");
+    }
     if (!cloudResource.isControlled) {
       throw new RuntimeException(resourceName + " is not a controlled resource.");
     }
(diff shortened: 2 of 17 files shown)


README.md
```diff
@@ -227,6 +231,23 @@ Commands:
...
+Commands:
+  list          List all data references.
+  add           Add a new data reference.
+  delete        Delete an existing data reference.
+  check-access  Check if you have access to a data reference.
+  resolve       Resolve a data reference to its cloud id or path.
+```
+
+A data reference points to a cloud resource that the user wants to read or write data to.
+Data references include some types of controlled resources (e.g. bucket but not VM) and
```
wchamber, line 247: nit: maybe note that it's about data based controlled resources, not compute based controlled resources?

src/main/java/bio/terra/cli/service/utils/SamService.java
```diff
@@ -168,6 +169,25 @@ public void getOrRegisterUser() {
     }
   }
 
+  /**
+   * Call the SAM "/api/google/v1/user/proxyGroup/{email}" endpoint to get the email for the current
+   * user's proxy group.
+   *
+   * <p>Update the Terra User object with the proxy group email.
+   */
+  public void getProxyGroupEmail() {
```
wchamber, line 178: it's a little surprising to me that a `get` function returns void and mutates state. Couple ideas, a) rename to `storeProxyGroupEmail`, b) return the value and modify the terraUser in the calling function i.e. `AuthenticationManager`, or c) make TerraUser a smarter class that has `getOrRetrieveProxyGroupEmail` that grabs it from local storage or finds it in Sam as needed. I see this follows existing functions `getUser` and `getOrRegisterUser`, so also reasonable to change all in a later PR.

src/main/java/bio/terra/cli/service/WorkspaceManager.java
```diff
@@ -269,4 +275,90 @@ public CloudResource deleteControlledResource(String resourceName) {
...
+   *
+   * <p>This method checks that the data reference exists using the user's credentials. The
+   * reference does not have to be within the workspace project.
+   *
+   * @param referenceType type of reference to add
+   * @param referenceName name of reference to add
+   * @return the data reference that was added
+   */
+  public CloudResource addDataReference(
+      CloudResource.Type referenceType, String referenceName, String cloudId) {
```
wchamber, line 309: maybe `cloudUri` instead of `cloudId`?

src/main/java/bio/terra/cli/service/WorkspaceManager.java
```diff
@@ -269,4 +275,90 @@ public CloudResource deleteControlledResource(String resourceName) {
...
+  }
+
+  /**
+   * Add a new data reference in the workspace.
+   *
+   * <p>This method checks that the data reference exists using the user's credentials. The
+   * reference does not have to be within the workspace project.
+   *
+   * @param referenceType type of reference to add
+   * @param referenceName name of reference to add
```
wchamber, line 305: nit: add a comment for `cloudId` expected format.

src/main/java/bio/terra/cli/service/utils/GoogleCloudStorage.java
```diff
@@ -13,37 +15,44 @@
...
   }
 
-  /** Build the GCS client object for the given Terra user. */
-  private void buildClientForTerraUser() {
+  /**
+   * Build the GCS client object for the given credentials and project.
+   *
+   * @return the GCS client object
+   */
+  private Storage buildClientForTerraUser() {
```
wchamber, line 46: `ForTerraUser` isn't as accurate now since it's just looking at the passed credentials. `buildClient` is probably fine.
Since this is being called only as a helper in the constructor, consider making this `static` and passing the required arguments directly. I don't think we need to keep the `googleCredentials` otherwise on this class.

wchamber (approved):

README.md
marikomedlock: done.

src/main/java/bio/terra/cli/service/WorkspaceManager.java
marikomedlock: I prefer cloudId here, since this is intended to handle other types of data references (e.g. BQ datasets) in the future. When I get to actually implementing those, maybe that will be awkward and it'll make more sense to split them up into different method, or maybe WSM endpoints will be ready and we'll call them. Either way, I prefer to keep it generic for now and specific later once we have more example resource types.

src/main/java/bio/terra/cli/service/WorkspaceManager.java
marikomedlock: added.

src/main/java/bio/terra/cli/service/utils/GoogleCloudStorage.java
marikomedlock: Good catch. Removed the `ForTerraUser` part of the method name and removed the `googleCredentials` as a property.

src/main/java/bio/terra/cli/service/utils/SamService.java
marikomedlock: Agreed. Updated them to return values instead.

src/main/java/bio/terra/cli/service/utils/GoogleCloudStorage.java
```diff
@@ -74,4 +79,33 @@ public void deleteBucket(String bucketUri) {
...
+    try {
+      String bucketName = bucketUri.replaceFirst("^gs://", "");
+
+      // try listing objects in the bucket
+      storageClient.list(
+          bucketName,
+          Storage.BlobListOption.userProject(googleProjectId),
+          Storage.BlobListOption.fields());
+      return true;
+    } catch (StorageException storageEx) {
```
gjuggler, line 102: [Not necessarily for this PR]

I went through a round of testing and got tripped up here. I was using a workspace that I'd created maybe 1 or 2 weeks ago. And I kept getting "Invalid or inaccessible bucket path" exceptions when trying to create a data-ref, even after double-checking that my access was set up properly.

After turning on debug-level logging in logback.xml, I was able to see the underlying error:

`com.google.cloud.storage.StorageException: gjuggler@gmail.com does not have serviceusage.services.use access to the Google Cloud project.`

Aha – this is because we're always passing in the "context" project ID to the GCS list call, which is required for requester-pays buckets. However, a user must have the serviceusage.services.use permission on this project in order for a call to succeed ([ref](https://cloud.google.com/storage/docs/requester-pays#requirements)).

It seems clear that https://github.com/DataBiosphere/terra-workspace-manager/pull/230 should fix this for newer workspaces. But I thought it's worth noting that this is needed not just for Nextflow, but for any use of the workspace context as a GCS requester-pays billing project.

On a secondary note, it would have been nice to somehow get a more specific error message here. Two possible improvements to make:

(1) Add a user-accessible --debug or --verboase flag that shows some of the logger logs. This would make it easier for me to debug without having to recompile. Maybe worth filing a low-priority follow up ticket?
(2) Potentially increase the specificity of the user-facing message on the access-check. In this case, it might make more sense to inspect the StorageException a bit more and re-throw a more user-friendly exception message, rather than a simple true or false, since there may be multiple causes of a failed access-check.

gjuggler (approved):
Sorry for the delay! Code looks good; just one comment from testing this out locally.

src/main/java/bio/terra/cli/service/utils/GoogleCloudStorage.java
marikomedlock: Good point about the requester-pays buckets. Yes, that's why I'm passing the project into the GCS class, but I hadn't tested it on an older workspace.

I created a new [ticket](https://broadworkbench.atlassian.net/browse/PF-542) to control logging without requiring a re-compile. Prefer to hold off on better exception interpretation for now, until we get some feedback from the solutions team about what `check-access` should mean.

PR #26, opened 2021-03-05
PF-527 Change the notebooks default image family

wchamber:
Per solutions team, the r experimental image that has Java installed is a better default choice.
Later we can look at finding an image that isn't experimental, but this image at least will allow installation of the terra CLI out of the box on the notebook instance.

Add `showDefaultValues` to all notebooks commands.


marikomedlock (approved):
Would be good to have this as a static constant field once we move notebook logic out of the command classes.

marikomedlock (approved):
Also, perhaps include the default value in the option description?

wchamber:
> Also, perhaps include the default value in the option description?

added it for all notebook commands for all values.

PR #27, opened 2021-03-09
PF-527 Add the workspace id to the Notebooks instance metadata.

wchamber:
Set the terra workspace id as metadata on created Notebooks instance VM metadata.
While this isn't a full solution for auto populating useful Terra variables for a seamless notebooks experience, this does make the workspace id accessible programmatically from the VM, which we can use for further integration.
See https://cloud.google.com/compute/docs/storing-retrieving-metadata

### DIFF
--- a/src/main/java/bio/terra/cli/command/notebooks/Create.java
+++ b/src/main/java/bio/terra/cli/command/notebooks/Create.java
@@ -79,7 +79,9 @@ public class Create implements Callable<Integer> {
             // The metadata installed-extensions causes the AI Notebooks setup to install some
             // Google JupyterLab extensions. Found by manual inspection of what is created with the
             // cloud console GUI.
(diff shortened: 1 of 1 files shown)


wchamber:
> It would be nice if in the notebook VM startup, we could query the GCP metadata and set the workspace for them via Terra CLI or env var or both. Is that what you're thinking of in terms of further integration?
> 
> I think the [project](https://cloud.google.com/compute/docs/storing-retrieving-metadata#project-instance-metadata) metadata is already available on the VM. If so, then what do you think of a reverse lookup endpoint on WSM to get the workspace id associated with the project?
Yep. Doing a lookup from projectId -> workspace id would also work, but since we don't have those endpoints on WSM yet, making the workspace id available in the metadata for the VM also works.
Couple ideas there - set the env var & terra CLI installation as a post-startup-script, or just provide examples.

PR #34, opened 2021-03-13
PF-569: Bump version on merge to main with GH action.

marikomedlock:
- Added a GH action to add a GH tag and bump the version string in `build.gradle` on each merge to the `main` branch.
- The tag bumper action uses the BroadBot GH token, added to this repo by DevOps.
- Left a placeholder step for calling the `publish-release.sh` script in a future PR (needs secrets added to GH repo).

### DIFF
--- a/.github/workflows/on-pr-merge.yml
+++ b/.github/workflows/on-pr-merge.yml
@@ -4,7 +4,8 @@ on:
   push:
     branches:
       - main
-
+  pull_request:
+    branches: [ '**' ]
 jobs:
   bump-version-publish-release:
     runs-on: ubuntu-latest
@@ -20,7 +21,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           RELEASE_BRANCHES: main
           VERSION_FILE_PATH: build.gradle
-          VERSION_LINE_MATCH: ^version
+          VERSION_LINE_MATCH: "^version\ "
       - name: Publish a release
         run: |
           echo "TODO (PF-570): ./tools/publish-release.sh $RELEASE_VERSION true"
--- a/.github/workflows/on-pr-merge.yml
+++ b/.github/workflows/on-pr-merge.yml
@@ -21,7 +21,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           RELEASE_BRANCHES: main
           VERSION_FILE_PATH: build.gradle
-          VERSION_LINE_MATCH: "^version\ "
+          VERSION_LINE_MATCH: ^version\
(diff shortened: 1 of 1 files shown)


.github/workflows/on-pr-merge.yml
```diff
@@ -0,0 +1,29 @@
+name: Bump version and publish release
+on:
+  workflow_dispatch: {}
```
jaycarlton, line 3: do you want people to be able to run this from the Actions page without pushing code to main?

.github/workflows/on-pr-merge.yml
```diff
@@ -0,0 +1,29 @@
...
+        uses: actions/checkout@master
+        with:
+          token: ${{ secrets.BROADBOT_GITHUB_TOKEN }}
+      - name: Bump tag and build version
+        uses: databiosphere/github-actions/actions/bumper@v0.0.3
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          RELEASE_BRANCHES: main
+          VERSION_FILE_PATH: build.gradle
+          VERSION_LINE_MATCH: ^version
```
jaycarlton, line 23: you might want to put a word boundary on the right so it doesn't match `versionOptions` or something down the road.

.github/workflows/on-pr-merge.yml
```diff
@@ -0,0 +1,29 @@
...
+      - name: Publish a release
+        run: |
+          echo "TODO (PF-570): ./tools/publish-release.sh $RELEASE_VERSION true"
+        env:
+          GITHUB_TOKEN: ${{ secrets.BROADBOT_GITHUB_TOKEN }}
+          RELEASE_VERSION: ${{ steps.tag.outputs.tag }}
```
jaycarlton, line 29: how does this string work? I don't see `tag` used as an identifier.

jaycarlton (approved):

.github/workflows/on-pr-merge.yml
marikomedlock: Yes, exactly. In case of a manual release, I think it would be slightly better to trigger a GH action than to run a script on your laptop.

.github/workflows/on-pr-merge.yml
marikomedlock: It's an [output](https://github.com/DataBiosphere/github-actions/tree/master/actions/bumper#outputs) of the previous step.

.github/workflows/on-pr-merge.yml
marikomedlock: I think that's a good point, but going to leave it as is for now, to be consistent with other repos (stairway, common-lib).

PR #35, opened 2021-03-15
PF-570: Publish release on merge to main.

marikomedlock:
- The GH action that runs on merge to `main` now does the following:
  - Bumps the version in build.gradle and tags the branch with the new version. (Previously, it was doing the tagging, but there was a bug in the build.gradle updating.)
  - Publishes a release, which includes a GH release with the install package and uploading a Docker image to GCR.

- The Docker image path was previously duplicated in two places: the publishing scripts (because they need to know where to push the image to), and the Java code (because it needs to know which Docker image to use to start the container). Consolidated this to just use `gradle.properties`.
  - Added convenience Gradle tasks to read these properties. The publishing scripts use these.
  - The Gradle build uses the properties directly to write the full image path to the JAR manifest.
  - The Java code reads the full image path from the JAR manifest.
  - This change wasn't really feasible without the GH action that automatically publishes on each merge to main, because it was cumbersome to have to manually publish a new Docker image and install package for each PR.

- Updated the README with a note about re-installing and manual uninstalling.

**TODO** Update the branch names in the GH action yml file to put them back to `main` before merging.


README.md
```diff
@@ -41,16 +44,8 @@ curl -L https://github.com/DataBiosphere/terra-cli/releases/latest/download/down
...
-`download-install.sh` is a convenience script that downloads the latest (or specific version) of the install package,
-unarchives it, runs the `install.sh` script included inside, and then deletes the install package.
-
-You can also skip the `download-install.sh` script and install manually.
-- Download the `terra-cli.tar` install package directly from the 
-[GitHub releases page.](https://github.com/DataBiosphere/terra-cli/releases)
-- Unarchive the `tar` file.
-- Run the install script from the unarchived directory: `./install.sh`
+Re-installing will overwrite any existing installation (i.e. all JARs and scripts will be overwritten), but will not
+modify the `$PATH`. So if you have added it to your `$PATH`, that step needs to be repeated after each install.
```
jmczerk, line 48: This is a little confusing to me.  Is this meant to address the case where the user has moved the install somewhere other than `~/.terra` ?

.github/workflows/on-pr-merge.yml
```diff
@@ -3,27 +3,49 @@ on:
...
+          GITHUB_TOKEN: ${{ secrets.BROADBOT_GITHUB_TOKEN }}
+      - name: Render config
+        id: render_config
+        run: |
+          # this step does the equivalent of the tools/render-config.sh script
+          # on local machines, the script fetches a SA from Vault
+          # in GH actions, the SA key is stored in a GH repo secret
+          # regardless of how it was fetched, the publish-release script expects this
+          # key to be stored in rendered/ci-account.json
+          mkdir -p rendered
```
jmczerk, line 41: Is there any concern about leaving `rendered/ci-account.json` hanging around at the end of the workflow run?

README.md
marikomedlock: The install script installs the Terra CLI in the current directory, but does not modify your `$PATH` (i.e. you have to call `./terra` instead of `terra`). This is a common install pattern, and specifically mimics the installer for Nextflow.

However, it also means that if you do want to add the Terra CLI to your path, you need to do so as a separate step after install. This note was just a reminder that neither installing nor re-installing affect the `$PATH`.

.github/workflows/on-pr-merge.yml
marikomedlock: No, the GH action runners won't be reused as is.

README.md
jmczerk: Ah, I see.  So the libs/scripts get overwritten in `~/.terra` and the new exe lands in `pwd` (which may or may not overwrite the existing exe, depending on where the existing exe lives).  This is just a reminder to make sure `$PATH` resolves to the new exe (regardless of how it got onto the path).

marikomedlock: Yes, exactly.

jmczerk (approved):

jaycarlton (approved):

CONTRIBUTING.md
```diff
@@ -119,6 +119,9 @@ To use a specific Docker image from GCR:
     ```
 
 #### Build a new image
+The `tools/publish-release.sh` script builds and publishes a new image, and updates the default image path already.
```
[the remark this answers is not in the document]
marikomedlock: I was trying to say that merging a PR and installing should take care of all this Docker image stuff for you, so these notes are mostly useful for debugging/development when you need to make an image available outside of that normal process. But that was not very clear -- did some rewording.

.github/workflows/on-pr-merge.yml
```diff
@@ -3,27 +3,50 @@ on:
...
       - name: Bump tag and build version
+        id: bump_tag
         uses: databiosphere/github-actions/actions/bumper@v0.0.3
         env:
           GITHUB_TOKEN: ${{ secrets.BROADBOT_GITHUB_TOKEN }}
-          RELEASE_BRANCHES: main
+          DEFAULT_BUMP: minor
+          RELEASE_BRANCHES: mm-PF-569-debug-bumper-gha # TODO: change this back to main before merge
           VERSION_FILE_PATH: build.gradle
           VERSION_LINE_MATCH: ^version
```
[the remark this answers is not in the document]
marikomedlock: Good idea. Done.

build.gradle
```diff
@@ -20,7 +20,7 @@ if (hasProperty('buildScan')) {
 }
 
 // The tools/publish-release.sh script depends on this version string being of the format "version = '1.2.3'"
-version = '0.9.0'
+version = '0.18.0'
```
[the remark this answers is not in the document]
marikomedlock: Yep good catch. Reset.

PR #36, opened 2021-03-16
PF-423: Separate exceptions into user-facing and internal.

marikomedlock:
- Separated exceptions into user-actionable and system error sub-classes. Replaced all RuntimeExceptions with one of the new sub-classes.
- Added a custom exception handler to the main `picocli` class to handle user-actionable and system, and unexpected exceptions differently.
- Added two logback appenders: console and file (`$HOME/.terra/terra.log` in the global context directory). Replaced the `logback.xml` file with programmatic logging setup, so that we can manipulate the log level via a command in a future PR (see [PF-542](https://broadworkbench.atlassian.net/browse/PF-542)).
- The goal of the two appenders is to have more verbose logging go to the file, and shorter error messages go to the console. To this end, the default log levels are `OFF` and `INFO` for console and file, respectively.
- The file appender rolls log files. It stores a maximum of 5 log files at a time, of at most 5 MB each. That seems way more than enough for the usage patterns I'd expect for this CLI, so we may want to adjust that down in the future.
- Explicitly specified a color scheme for the commands. User actionable errors, including mistyping commands, prints in regular blue text. System and unexpected errors print with a bold red [ERROR] in front of the actual message.

Note to reviewers: There are a lot of files changed, but most are just replacements of `RuntimeException` with one of the new exception sub-classes, or changes to `logger.__` calls to make the levels more consistent across classes. The main things to review are:
- `Main.java` is the CLI entry point (i.e. the Java `main` method) and contains the custom exception handler.
- `SystemException.java` and `UserActionableException.java` are the new exception sub-classes.
- `GlobalContext.java` contains the pointers to the log levels and log file.
- `Logger.java` contains the appenders setup.


src/main/java/bio/terra/cli/command/Main.java
```diff
@@ -60,18 +71,71 @@ public static void main(String... args) {
...
+
+    private CommandLine commandLine;
+
+    @Override
     public int handleExecutionException(Exception ex, CommandLine cmd, ParseResult parseResult) {
+      this.commandLine = cmd;
+
+      String errorMessage;
+      boolean printPointerToLogFile;
+      if (ex instanceof UserFacingException) {
```
jmczerk, line 102: It looks like `Logger#LOG_FORMAT` is configured to prefix messages with their log level?  If so, you might consider moving the `"[ERROR] "` prefix append to `printErrorText()` time.

src/main/java/bio/terra/cli/command/Main.java
```diff
@@ -60,18 +71,71 @@ public static void main(String... args) {
...
+        errorMessage = "[ERROR] " + ex.getMessage();
+        printPointerToLogFile = false;
+      } else if (ex instanceof InternalErrorException) {
+        errorMessage =
+            "[ERROR] "
+                + ex.getMessage()
+                + ": "
```
jmczerk, line 109: Nit: could consider moving the ` + ": "` into the ternary true condition to prevent trailing colon when `getCause()` is null.

src/main/java/bio/terra/cli/context/utils/Logger.java
```diff
@@ -0,0 +1,92 @@
...
+   * Setup a file and console appender for the root logger. Each may use a different logging level,
+   * as specified in the global context.
+   */
+  public void setupLogging() {
+    LoggerContext loggerContext = (LoggerContext) LoggerFactory.getILoggerFactory();
+
+    // turn on logging for the logger for errors only
+    StatusPrinter.printIfErrorsOccured(loggerContext);
+
+    // build the file appender
```
jmczerk, line 41: Is rolling the logs a concern for the CLI, or do we expect them to be pretty self-limiting size-wise due to the usage model?

jmczerk (approved):
A couple minor questions/nits, but looks good otherwise.

src/main/java/bio/terra/cli/command/Main.java
marikomedlock: Yes, you're right. Done.

src/main/java/bio/terra/cli/command/Main.java
marikomedlock: Agreed, done.

src/main/java/bio/terra/cli/context/utils/Logger.java
marikomedlock: Good point. I think they'll be pretty self-limiting due to usage, but it's probably better to roll the files just in case. I'll take a stab at it.

src/main/java/bio/terra/cli/auth/AuthenticationManager.java
```diff
@@ -187,22 +188,22 @@ public void populateCurrentTerraUser() {
...
     } else {
       // ask SAM for the project-specific pet SA key
       HttpUtils.HttpResponse petSaKeySamResponse =
           new SamService(globalContext.server, terraUser).getPetSaKeyForProject(workspaceContext);
       if (!HttpStatusCodes.isSuccess(petSaKeySamResponse.statusCode)) {
-        logger.info("SAM response to pet SA key request: {})", petSaKeySamResponse.responseBody);
-        throw new RuntimeException(
+        logger.debug("SAM response to pet SA key request: {})", petSaKeySamResponse.responseBody);
+        throw new InternalErrorException(
             "Error fetching pet SA key from SAM (status code = "
```
gjuggler, line 207: [follow-up] Not for this PR, but as we look towards an external-facing release, it will make sense to scrub through these exceptions for acronyms or words that won't make sense to end users. For internal system errors, where we don't think they're user caused or user actionable, we may want to err on the side of more opaqueness in the user-facing message. (Of course there's a limit... I wouldn't just give a made-up error code, but possibly a generic message like "An error occurred while fetching Terra service credentials for your user account" combined with a more precise error locator code "(code 1234)".

src/main/java/bio/terra/cli/command/Main.java
```diff
@@ -61,18 +72,61 @@ public static void main(String... args) {
...
+      boolean printPointerToLogFile;
+      if (ex instanceof UserFacingException) {
+        errorMessage = ex.getMessage();
+        printPointerToLogFile = false;
+      } else if (ex instanceof InternalErrorException) {
+        errorMessage =
+            ex.getMessage() + (ex.getCause() != null ? ": " + ex.getCause().getMessage() : "");
+        printPointerToLogFile = true;
+      } else {
+        errorMessage = ex.getClass().getCanonicalName() + ": " + ex.getMessage();
```
gjuggler, line 104: Potential formatting tweak to give users a bit more context:

"An unexpected error occurred in {className}: {message}"

src/main/java/bio/terra/cli/command/Main.java
```diff
@@ -61,18 +72,61 @@ public static void main(String... args) {
...
 
   /**
-   * Custom exception handler class that suppresses the stack trace when an exception is thrown.
-   * Instead, it just prints the exception message and exits the process.
+   * Custom handler class that intercepts all exceptions.
+   *
+   * <p>There are three categories of exceptions. All print a message to stderr and log the
+   * exception.
+   *
+   * <p>- User-facing = user can fix
```
gjuggler, line 80: (Caveat: I may be dramatically over-thinking things here. Take with a grain of salt and feel free to ignore at your leisure!)

I thought I was grokking the categorization here until about 2/3 way through this PR, when I realized that both user-facing and Internal exceptions are getting shown to the user with very similar formatting... so now I'm having more trouble precisely distinguishing between the two use cases.

The main difference in handling seems to be that (1) for Internal exceptions we'll include a message from the causal exception, and (2) for Internal exceptions we point users to the log file for more details (presumably so they can find a stack trace).

But for both types of exceptions, we're showing users a big red '[ERROR]'-prefixed line with the exception message included. This feels quite similar to me, and may make it hard to describe clear conventions around what level of user-facing polish should be applied to each type of exception.

Spitballing a couple potential changes that may help draw out more of a difference:

1) Consider naming options.

Surveying the usage patterns, `UserFacingException` is mostly thrown when there is a simple precondition failure (e.g. "no workspace context") or a par-for-the-course error code returned from the service (e.g. "resource not found"). Maybe these could be called `UserActionableException` or simply `UserException`? Whereas `InternalErrorException` is usually about a server failure (5xx), file parsing failure, or some other internal workings of the system. Possibly `SystemException` captures that concisely? I realize this isn't too different from the current names... but somehow I found the `UserFacing` term slightly jarring, because ultimately all of these exceptions are shown to the user.

2) Consider a formatting difference between user-actionable and internal system exceptions.

If I use incorrect characters in an object name or try to fetch an ID that doesn't exist, I might expect a less-scary looking error message than if a server is down or there is an error parsing some core config file. To some extent, for the former type of "user error", I'd probably want it to look similar to what happens when I mis-type a command name. So maybe there's some stronger delineation we can aim for here in the formatted output, e.g.

UserActionable: normal color, less loud (maybe drop the [ERROR] prefix?), looks like a simple message "Hey, you did something wrong" or "hey, we couldn't find that thing you asked for".
Internal / system: some red somewhere (maybe just the [ERROR] prefix is red), with loud with clear boilerplate indicating that a relatively serious system error has occurred.

src/main/java/bio/terra/cli/context/utils/Logger.java
```diff
@@ -0,0 +1,184 @@
...
+import ch.qos.logback.core.rolling.RollingFileAppender;
+import ch.qos.logback.core.rolling.SizeBasedTriggeringPolicy;
+import ch.qos.logback.core.spi.FilterReply;
+import ch.qos.logback.core.util.FileSize;
+import ch.qos.logback.core.util.StatusPrinter;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import java.io.File;
+import org.slf4j.LoggerFactory;
+
+/** Utility methods for logging. */
```
gjuggler, line 23: nit: this is more than just utility methods. Maybe re-phrase as "Utility class to initialize a Logback context for the main CLI app. Requires an already-loaded GlobalContext for defining logging file outputs and loading config properties."

src/main/java/bio/terra/cli/context/utils/Logger.java
```diff
@@ -0,0 +1,184 @@
...
+      OutputStreamAppender appender, LoggerContext loggerContext, Level loggingLevel) {
+    PatternLayoutEncoder encoder = new PatternLayoutEncoder();
+    encoder.setContext(loggerContext);
+    encoder.setPattern(LOG_FORMAT);
+    encoder.start();
+    appender.setEncoder(encoder);
+
+    // filter out any logs that are below the fileLoggingLevel specified in the global context
+    appender.clearAllFilters();
+    appender.addFilter(
```
gjuggler, line 146: nit / side note: I'm a little surprised there isn't some built-in class that does this already.

src/main/java/bio/terra/cli/context/utils/Logger.java
```diff
@@ -0,0 +1,184 @@
...
+          public FilterReply decide(final ILoggingEvent event) {
+            return event.getLevel().isGreaterOrEqual(loggingLevel)
+                ? FilterReply.ACCEPT
+                : FilterReply.DENY;
+          }
+        });
+  }
+
+  /**
+   * Wrapper class around the ch.qos.logback.classic.Level class.
```
gjuggler, line 158: [No action / side comment] I *think* I understand the intent here. But the only use case I can envision for this is in GlobalContext to configure the logback log level (either as part of a stored configuration or via command-line arguments). In which case, it's not too inappropriate to say that GlobalContext should know that our implementation is logback, because it's directly configuring the logback system.

Is there some other scenario you're protecting against, where these levels become more widely-used throughout the codebase?

Or, maybe the concern is more about how these values might get serialized alongside the GlobalContext? In which case a small note about serialization control might help clarify the need.

src/main/java/bio/terra/cli/context/utils/Logger.java
```diff
@@ -0,0 +1,184 @@
...
+   * Setup a file and console appender for the root logger. Each may use a different logging level,
+   * as specified in the global context.
+   */
+  @SuppressFBWarnings(
+      value = "NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE",
+      justification =
+          "An NPE would only happen here if there was an error getting the global context directory,"
+              + " and an exception would have been thrown earlier when we first read in the global "
+              + "context file anyway.")
+  public void setupLogging() {
```
gjuggler, line 52: [Side note / comment] While it might have technically been possible to set up a console / file-based output with a logback.xml and various properties getting passed in (like the location of the log file), 💯 to doing things programmatically — it's far easier to understand, debug, and improve the flow over time. I ran into a similar challenge in setting up JSON logging on the service side, with a similar solution. (In fact there are some improvements I'd like to make after reading your code here, like the `StatusPrinter.printIfErrorsOccurred` safety net.)

gjuggler (approved):
Nice improvements! I dropped a bunch of fairly long comments / ideas, but there's nothing that I would block merging on. If you do pick up any larger changes feel free to ping me for another look, but I'm flipping the approval bit now.

src/main/java/bio/terra/cli/context/utils/Logger.java
```diff
@@ -0,0 +1,184 @@
...
+    LoggerContext loggerContext = (LoggerContext) LoggerFactory.getILoggerFactory();
+    loggerContext.reset();
+
+    // turn on logging for the logger for errors only
+    StatusPrinter.printIfErrorsOccured(loggerContext);
+
```
jmczerk, line 58: Thanks for adding the `RollingFileAppender`!

src/main/java/bio/terra/cli/command/exception/UserFacingException.java
```diff
@@ -0,0 +1,26 @@
+package bio.terra.cli.command.exception;
+
+/**
+ * Custom exception class for user-facing exceptions. These represent errors that the user can fix.
+ * (e.g. "Invalid bucket path" when creating a data reference).
```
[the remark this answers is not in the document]
marikomedlock: Good point. Added.

src/main/java/bio/terra/cli/context/GlobalContext.java
```diff
@@ -54,14 +64,17 @@ private GlobalContext(ServerSpecification server, String dockerImageId) {
...
+   * first call this function.
+   *
    * @return an instance of this class
    */
   public static GlobalContext readFromFile() {
     // try to read in an instance of the global context file
     try {
       return FileUtils.readFileIntoJavaObject(getGlobalContextFile().toFile(), GlobalContext.class);
     } catch (IOException ioEx) {
-      logger.warn("Global context file not found or error reading it.", ioEx);
```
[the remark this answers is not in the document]
marikomedlock: This is a pretty common error case (e.g. file not found when a user first runs the CLI and there is no pre-existing global context file) and it should not be fatal. We handle this by returning an object populated with default values in the lines below. So, I think there's no need to log or print the error returned here. Added a comment to this effect in the catch block.

src/main/java/bio/terra/cli/command/Main.java
```diff
@@ -61,18 +72,61 @@ public static void main(String... args) {
...
       implements CommandLine.IExecutionExceptionHandler {
+
+    @Override
     public int handleExecutionException(Exception ex, CommandLine cmd, ParseResult parseResult) {
+      String errorMessage;
+      boolean printPointerToLogFile;
+      if (ex instanceof UserFacingException) {
+        errorMessage = ex.getMessage();
+        printPointerToLogFile = false;
+      } else if (ex instanceof InternalErrorException) {
```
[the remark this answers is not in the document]
marikomedlock: This looks great. Agreed this is probably a couple months away, but I'll keep in mind for then. Thanks!

src/main/java/bio/terra/cli/auth/AuthenticationManager.java
```diff
@@ -187,22 +188,22 @@ public void populateCurrentTerraUser() {
   public void fetchPetSaCredentials(TerraUser terraUser) {
     // if the current workspace is not defined, then we don't know which pet SA to fetch
     if (workspaceContext.isEmpty()) {
-      logger.info("There is no current workspace defined.");
+      logger.debug("There is no current workspace defined.");
```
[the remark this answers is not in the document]
marikomedlock: Added.

src/main/java/bio/terra/cli/context/utils/Logger.java
marikomedlock: Thank you for the great suggestion! I'm working on addressing Greg's comments now, and then I think I'm going to ping you both for a re-review, if you don't mind. Enough has changed since the original that another once-over by you guys would be much appreciated.

src/main/java/bio/terra/cli/auth/AuthenticationManager.java
marikomedlock: Good idea. Added this to the list of style improvements in [PF-574](https://broadworkbench.atlassian.net/browse/PF-574).

src/main/java/bio/terra/cli/command/Main.java
marikomedlock: Changed.

src/main/java/bio/terra/cli/context/utils/Logger.java
marikomedlock: Updated.

src/main/java/bio/terra/cli/context/utils/Logger.java
marikomedlock: You're right. Found `ThresholdFilter` and used that instead.

src/main/java/bio/terra/cli/context/utils/Logger.java
marikomedlock: Without this wrapper class, we'd leak logger implementation into:
- GlobalContext, including the serialized version persisted in a file.  ch.qos.logback.classic.Level is an object, not a string, and I didn't want this file to include logger implementation-specific object structure. That would make it harder to change later (i.e. because if we replace the cho.qos library with another one, then we won't have the appropriate  Level class any more to do the deserialization)
- (Upcoming) Config command to set the log level(s), including displaying the possible values.

Added a note about this to the code.

src/main/java/bio/terra/cli/command/Main.java
marikomedlock: This makes sense about there not being a big enough difference between user-actionable and system/unexpected exceptions.
1. Renamed the exception classes to UserActionableException and SystemException.
2. Changed the formatting so that system and unexpected exceptions show a bold red message, and user-actionable exceptions show a regular size blue message. Since command typing errors (e.g. forgot a required flag) are also user-actionable, I also changed them to follow this convention.

I also explicitly defined the color scheme in the `Main` class, so that we can modify it more easily in the future. Below is an example of what it looks like now (I inserted test exceptions in the `terra status` and `terra version` commands):
Screen Shot 2021-03-17 at 2 20 39 PM

marikomedlock: I don't have any strong preferences around the colors here, so if there's any changes you guys would like to see, just let me know. (There's no orange though FYI 🤷‍♀️ ).

gjuggler (approved):
🥳  awesome and thoughtful changes! And thanks for the screenshot. That looks really polished, IMO.

src/main/java/bio/terra/cli/context/utils/Logger.java
```diff
@@ -0,0 +1,191 @@
...
+/**
+ * Utility class to initialize and configure a Logback context and loggers for the main CLI app.
+ * Requires an already-loaded GlobalContext for defining logging output files and loading config
+ * properties.
+ */
+public class Logger {
+  private static final String LOG_FORMAT =
+      "%d{HH:mm:ss.SSS} [%thread] %-5level %logger{50} - %msg%n";
+
```
jmczerk, line 29: Intuitively 50 MB * 5 files seems an order of magnitude high to me; maybe 5MB * 5?  I'll defer to your intuition though, as I don't know how big these get when debugging...

jmczerk (approved):
Thanks for the screenshot, it was really helpful to see the color scheme in action.  One minor note on log file limits, but otherwise this looks great to me.

src/main/java/bio/terra/cli/context/utils/Logger.java
marikomedlock: I have no idea either. We may want to change this in the future once we actually start using these debug logs. Mostly I'm thinking of these logs as a hedge for when people other than us start installing this. Changed to 5MB * 5 instead.

PR #38, opened 2021-03-18
PF-533: List workspaces command

marikomedlock:
- Added a `terra workspace list` command to list all workspaces the user has read access to.
  * If there is a workspace already mounted, it starts this one.
  * Optional `--offset` flag for users with many workspaces. Default is to start from the beginning. Each call fetches up to 30 workspaces.
- Updated the README to include the new command, and notes from yesterday's NF demo.
- Cleaned up some incorrect `@throws` Javadocs after the recent exception handling refactoring.

### DIFF
--- a/src/main/java/bio/terra/cli/command/workspace/List.java
+++ b/src/main/java/bio/terra/cli/command/workspace/List.java
@@ -24,6 +24,13 @@ public class List implements Callable<Integer> {
           "The offset to use when listing workspaces. (Zero means to start from the beginning.)")
   private int offset;
 
+  @CommandLine.Option(
+      names = "--limit",
+      required = false,
+      defaultValue = "30",
+      description = "The maximum number of workspaces to return.")
+  private int limit;
+
   @Override
   public Integer call() {
     GlobalContext globalContext = GlobalContext.readFromFile();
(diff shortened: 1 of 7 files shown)


zloery (approved):
LGTM. I think we'll probably add workspace name + description fields to the `WorkspaceDescription` objects WSM returns at some point in the near future which would also mean adding it here, but it looks like you're already tracking that in PF-534

src/main/java/bio/terra/cli/command/workspace/List.java
```diff
@@ -0,0 +1,48 @@
...
+    showDefaultValues = true)
+public class List implements Callable<Integer> {
+
+  @CommandLine.Option(
+      names = "--offset",
+      required = false,
+      defaultValue = "0",
+      description =
+          "The offset to use when listing workspaces. (Zero means to start from the beginning.)")
+  private int offset;
```
jaycarlton, line 25: Why not add an option for the limit as well?

jaycarlton (approved):
Looks good.

src/main/java/bio/terra/cli/command/workspace/List.java
marikomedlock: You know, I initially thought that was maybe too much flexibility because why would someone want 100s of workspaces returned. But on second thought, seems like I can just make a reasonable default and then let users optionally override. I'll update this.

PR #39, opened 2021-03-19
PF-488: Global config command.

marikomedlock:
- Added a `config` group of commands to handle various options that control the behavior of the CLI. The `list`/`get-value`/`set` pattern is the same as `gcloud config` commands.
```
Usage: terra config [COMMAND]
Configure the CLI.
Commands:
  list       List all configuration properties and their values.
  get-value  Get a configuration property value.
  set        Set a configuration property value.
```
- Currently there are two examples of config values:
  * Whether to launch a browser automatically or not during the login flow.
  * Default Docker image id to use for running apps
```
Usage: terra config get-value [COMMAND]
Get a configuration property value.
Commands:
  browser  Check whether a browser is launched automatically during the login
             process.
  image    Get the Docker image used for launching applications.
```
- Removed the commands that previously controlled these options:
  * `terra auth get-browser` -> `terra config get-value browser`
  * `terra auth set-browser` -> `terra config set browser`
  * `terra app get-image` -> `terra config get-value image`
  * `terra app set-image` -> `terra config set image`


src/main/java/bio/terra/cli/command/config/List.java
```diff
@@ -0,0 +1,22 @@
...
+/** This class corresponds to the third-level "terra config list" command. */
+@CommandLine.Command(
+    name = "list",
+    description = "List all configuration properties and their values.")
+public class List implements Callable<Integer> {
+
+  @Override
+  public Integer call() {
+    GlobalContext globalContext = GlobalContext.readFromFile();
+
```
jmczerk, line 16: I like the human readable description for the parameter; it might be useful to also print out the corresponding parameter name as well.

jmczerk (approved):
This is a nice change, should make it easier to add configurable going forward.  

Added one minor suggestion on `list` output.  I also don't personally understand why `gcloud` added the `-value` to the getter (and not the setter/lister), but I'm OK with us doing so if we see value in aligning to their interface.

src/main/java/bio/terra/cli/command/config/List.java
marikomedlock: Good idea. Added the parameter name to the `list` output.

PR #40, opened 2021-03-22
PF-446: Common base class for commands. (part 1)

marikomedlock:
The goals of this change are to:
- Add a `--format=json` option to all relevant commands. (This PR adds it to ~1/3 of the commands so far). This is a request from the solutions team to make it easier to script with the Terra CLI.
- Reduce duplicate code in the command classes. In particular, many classes were reading in the global & workspace context files, and logging in the user.
---------------------------------
Summary of the changes included here:
- Added `BaseCommand`, a common base class for all commands. It handles reading the global & workspace context, and logging in the user if the command requires it.
- Added `FormatOption`, a picocli [Mixin](https://picocli.info/#_mixins) class that will be used by most commands. It includes utility methods for printing the command in text or JSON format, depending on the value of the `--format` flag. These utility methods can be called directly by the command classes. Currently, the only two possible values are `text` and `json`.
- Updated command classes at the top-level (i.e. `status`, `version`), and under the `auth`, `app`, and `data-refs` groups, to extend the new base class and use the mixin if it makes sense. (A future PR will do the same for other command classes. This is just to keep the number of files manageable for this initial review.)
---------------------------------
**REVIEWERS**: This change affected a lot of files, but many changes are very similar.
- The main changes are in:
  - `BaseCommand` the common base class
  - `FormatOption` the Mixin class that defines the `--format=json` flag and helper methods for printing output.
- Examples of how this changed various command classes:
  - `auth.Login` does not implement the `--format=json` flag
  - `command.Version` does not require login
  - `passthrough.Gcloud` does not have a return value
  - `datarefs.CheckAccess` defines a custom return type
  - `datarefs.Add` uses an existing return type


src/main/java/bio/terra/cli/command/baseclasses/BaseCommand.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    // get the return value object
+    T returnValue = execute();
+
+    // print the return value
+    printReturnValue(returnValue);
+
+    // set the command exit code
+    return 0;
```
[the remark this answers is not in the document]
marikomedlock: Yeah, I really struggled with the right structure here. I think a discussion over GVC would be helpful -- I'll put something on the calendar for later today/tomorrow for the three of us.

- RE base command class: I think there's value in each command class not needing to add boilerplate to load context and login. If a command doesn't need context (e.g. auth status doesn't need workspace context), that's fine it just doesn't need to use that property. The context load won't fail regardless.
- RE mixins: I tried this originally, but the first problem I ran into was that mixins only gave me the option/annotation (i.e. `--format=json` flag), not any boilerplate handling. So it saved the duplicated flag definition, but not the duplicated printing methods. Then I tried making an interface that required the printText & printJson methods instead -- so commands would use a mixin + implement an accompanying interface. I was on the fence about whether a mixin + interface was better/worse than sublcassing. Would appreciate other's opinions.

- RE needing to define `AuthStatusReturnValue`: I don't completely follow this part of your comment. The reason for that class is to support a JSON-formatted output. That wouldn't go away if we got rid of the base class.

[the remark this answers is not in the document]
marikomedlock: Thanks @wchamber @jmczerk for the discussion today! PTAL at take 2.

src/main/java/bio/terra/cli/command/helperclasses/CommandSetup.java
```diff
@@ -0,0 +1,49 @@
...
+      new AuthenticationManager(globalContext, workspaceContext).loginTerraUser();
+    }
+
+    // execute the command
+    execute();
+
+    // set the command exit code
+    return 0;
+  }
+
```
jmczerk, line 33: A nice side effect of this change is that the `void` return on `execute()` makes it unambiguous that commands should be using exceptions instead of return codes for errors (as a newbie to CLI repo I had to go look at Main to make sure).  It might be worth adding a comment to the effect that `main()` will handle exceptions thrown by commands to the docs here.

src/main/java/bio/terra/cli/command/Status.java
```diff
@@ -1,29 +1,64 @@
...
-    GlobalContext globalContext = GlobalContext.readFromFile();
-    WorkspaceContext workspaceContext = WorkspaceContext.readFromFile();
+  protected void execute() {
+    StatusReturnValue statusReturnValue =
+        new StatusReturnValue(globalContext.server, workspaceContext.terraWorkspaceModel);
+    formatFlag.printReturnValue(statusReturnValue, returnValue -> this.printText(returnValue));
+  }
+
+  /** POJO class for printing out this command's output. */
+  public static class StatusReturnValue {
```
wchamber, line 25: do these "return value" classes still need to be `public`?
optional: do we have any more preferable naming terminology than "return value"? I don't hate it, but the values aren't being returned anymore so much as printed or outputted? brainstorming: `StatusOutput`, `StatusOutputValue`, `StatusPrintValue`? not sure those are better, up to you

src/main/java/bio/terra/cli/command/Version.java
```diff
@@ -1,16 +1,30 @@
...
+public class Version extends CommandSetup {
 
+  @CommandLine.Mixin FormatFlag formatFlag;
+
+  /** Return value is just the version string. */
   @Override
-  public Integer call() {
-    System.out.println(bio.terra.cli.context.utils.Version.getVersion());
+  protected void execute() {
+    String versionReturnValue = bio.terra.cli.context.utils.Version.getVersion();
```
wchamber, line 17: nit: `version`'s probably fine, or just inlining it.

src/main/java/bio/terra/cli/command/auth/Status.java
```diff
@@ -1,38 +1,100 @@
...
     }
-    TerraUser currentTerraUser = currentTerraUserOpt.get();
-    System.out.println("Current Terra user: " + currentTerraUser.terraUserEmail);
-    System.out.println(
-        "Current Terra user's proxy group: " + currentTerraUser.terraProxyGroupEmail);
 
-    // check if the current user needs to re-authenticate (i.e. is logged out)
-    System.out.println("LOGGED " + (currentTerraUser.requiresReauthentication() ? "OUT" : "IN"));
+    /** Constructor for when there is NOT a current user defined. */
+    public AuthStatusReturnValue() {
```
wchamber, line 65: consider using a factory function for special constructions to make it more self documenting at the call site. e.g. `public static AuthStatusReturnValue createForNoDefinedUser()` or some better name.

src/main/java/bio/terra/cli/command/Status.java
```diff
@@ -1,29 +1,64 @@
...
+    public StatusReturnValue(ServerSpecification server, WorkspaceDescription workspace) {
+      this.server = server;
+      this.workspace = workspace;
+    }
+  }
+
+  /**
+   * Print this command's output in text format.
+   *
+   * @param returnValue command return value object
```
wchamber, line 41: optional nit: I don't think this javadoc adds much on this or other commands, but up to you.

src/main/java/bio/terra/cli/command/Status.java
```diff
@@ -1,29 +1,64 @@
...
+      OUT.println("Terra workspace: " + workspaceContext.getWorkspaceId());
+      OUT.println("Google project: " + workspaceContext.getGoogleProject());
     }
+  }
 
-    return 0;
+  /**
+   * This command never requires login.
+   *
+   * @return false, always
```
wchamber, line 58: optional nit: ditto not adding much. The notion that commands do or don't always require login is better suited for the base class - the `@Override` is a good hint to look at the base class for documentation. ditto other commands.

src/main/java/bio/terra/cli/command/auth/Status.java
```diff
@@ -1,38 +1,100 @@
...
 
-  @Override
-  public Integer call() {
-    GlobalContext globalContext = GlobalContext.readFromFile();
-    WorkspaceContext workspaceContext = WorkspaceContext.readFromFile();
+  @CommandLine.Mixin FormatFlag formatFlag;
 
+  /**
+   * Populate the current user in the global context and return an object with a subset of the
+   * TerraUser properties.
```
wchamber, line 19: not returning anything anymore

src/main/java/bio/terra/cli/command/datarefs/CheckAccess.java
```diff
@@ -1,52 +1,78 @@
...
-    boolean petSaHasAccess = dataReference.checkAccessForPetSa(currentTerraUser, workspaceContext);
+    boolean proxyGroupHasAccess =
+        dataReference.checkAccessForPetSa(currentTerraUser, workspaceContext);
+
+    CheckAccessReturnValue checkAccessReturnValue =
+        new CheckAccessReturnValue(userHasAccess, proxyGroupHasAccess);
+    formatFlag.printReturnValue(
+        checkAccessReturnValue,
+        returnValue -> {
+          this.printText(returnValue);
```
wchamber, line 40: nit: `CheckAccess::printText`?

src/main/java/bio/terra/cli/command/helperclasses/FormatFlag.java
```diff
@@ -0,0 +1,101 @@
...
+  /** This enum specifies the format options for printing the command output. */
+  private enum FormatOptions {
+    json,
+    text;
+  }
+
+  // output stream to use for writing command return value
+  private static final PrintStream OUT = System.out;
+
+  /**
```
jmczerk, line 32: Nit: should this be "calls printText method on the return value object" (instead of "of the return value object")?

src/main/java/bio/terra/cli/command/helperclasses/CommandSetup.java
```diff
@@ -0,0 +1,49 @@
+package bio.terra.cli.command.helperclasses;
+
+import bio.terra.cli.auth.AuthenticationManager;
+import bio.terra.cli.context.GlobalContext;
+import bio.terra.cli.context.WorkspaceContext;
+import java.io.PrintStream;
+import java.util.concurrent.Callable;
+
+public abstract class CommandSetup implements Callable<Integer> {
```
wchamber, line 9: worth some of the javadoc you had on BaseCommand earlier.
also curious about you're thinking about the rename. If we want all commands to inherit from this, `Base` seems good. If we add much else this to class so that it does something other than `Setup` the current name becomes misleading. Arguably, the PrintStream is not about setup,

src/main/java/bio/terra/cli/command/helperclasses/CommandSetup.java
```diff
@@ -0,0 +1,49 @@
...
+  /**
+   * This method returns true if login is required for the command. Default implementation is to
+   * always require login.
+   *
+   * <p>Sub-classes can use the current global and workspace context to decide whether to require
+   * login.
+   *
+   * @return true if login is required for this command
+   */
+  protected boolean doLogin() {
```
wchamber, line 46: optional nit: consider renaming to `requiresLogin()`

src/main/java/bio/terra/cli/command/helperclasses/CommandSetup.java
```diff
@@ -0,0 +1,49 @@
...
+    // do the login flow if required
+    if (doLogin()) {
+      new AuthenticationManager(globalContext, workspaceContext).loginTerraUser();
+    }
+
+    // execute the command
+    execute();
+
+    // set the command exit code
+    return 0;
```
wchamber, line 31: consider having `execute` return the command exit code. Might be good to start using non-zero exit codes for some failure cases..

src/main/java/bio/terra/cli/command/helperclasses/FormatFlag.java
```diff
@@ -0,0 +1,101 @@
...
+  private FormatOptions format;
+
+  /** This enum specifies the format options for printing the command output. */
+  private enum FormatOptions {
+    json,
+    text;
+  }
+
+  // output stream to use for writing command return value
+  private static final PrintStream OUT = System.out;
```
wchamber, line 30: not going to block the review on it, but I think we want this to be the same as what's in `CommandSetup`. These could both be initialized from some `OutputUtils.OUT` or something else.

jmczerk (approved):
I really like the way that this turned out using the format flag mixin.  I also really like the amount of code this removes, and the side effect that non-use of return values becomes explicit.

src/main/java/bio/terra/cli/command/helperclasses/FormatFlag.java
```diff
@@ -0,0 +1,101 @@
...
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.ObjectWriter;
+import java.io.PrintStream;
+import java.util.function.Consumer;
+import picocli.CommandLine;
+
+/**
+ * Command helper class that defines the --format flag and provides utility methods to printing a
+ * return value out in different formats.
```
wchamber, line 13: Worth mentioning this is meant to be used as a `@Mixin`?

src/main/java/bio/terra/cli/command/helperclasses/FormatFlag.java
```diff
@@ -0,0 +1,101 @@
...
+ */
+public class FormatFlag {
```
wchamber, line 15: Should we do `Option` instead of `Flag` to match picocli's terminology?

src/main/java/bio/terra/cli/command/helperclasses/FormatFlag.java
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  // output stream to use for writing command return value
+  private static final PrintStream OUT = System.out;
+
+  /**
+   * This method calls the {@link #printJson} method if the --format flag is set to JSON. Otherwise,
+   * it calls the {@link #printText} method of the return value object.
+   *
+   * @param returnValue command return value
```
wchamber, line 36: not really this PR: fwiw, we're aligning on not requiring param and return tags and suppressing the javadoc warnings.

wchamber (approved):
lgtm, just some nits

src/main/java/bio/terra/cli/command/helperclasses/CommandSetup.java
marikomedlock: Good idea. Added comment.

src/main/java/bio/terra/cli/command/helperclasses/FormatFlag.java
marikomedlock: Agreed this was unclear. Changed to "calls the {@link #printText} method, passing the return value object as an argument."

src/main/java/bio/terra/cli/command/Status.java
marikomedlock: Good catch. Changed "return value" classes to `private`.

RE: "return value" & other options. I think "output" or "print" would be slightly misleading for commands that output things besides the end object (e.g. docker commands that stream output, print progress messages). We don't have any of those commands yet, but I think they'll come in the future. I'm open to other suggestions, but I think "return" is slightly clearer for those cases than "print" or "output".

src/main/java/bio/terra/cli/command/Version.java
marikomedlock: done.

src/main/java/bio/terra/cli/command/auth/Status.java
marikomedlock: Updated to use a private constructor and public static methods.

src/main/java/bio/terra/cli/command/Status.java
marikomedlock: I prefer to have at least a brief description on all methods. Since these are `private` methods, I'm good with removing the `@param` though.

src/main/java/bio/terra/cli/command/Status.java
marikomedlock: Same as above, I prefer at least a brief comments on methods. But will remove the `@return` since this is not a `public` method.

src/main/java/bio/terra/cli/command/auth/Status.java
marikomedlock: Updated to "print out a subset of ..."

src/main/java/bio/terra/cli/command/datarefs/CheckAccess.java
marikomedlock: That method isn't `static`, so I don't think I can use that nicer notation. The reason it's not `static` is that it uses the global context object directly. I could have changed it to just use the return value object, but I thought it was useful to include the emails in just the text output, because the same information is already available for scripting in the `auth status` json output.

src/main/java/bio/terra/cli/command/helperclasses/CommandSetup.java
marikomedlock: RE javadoc: Good catch. I meant to copy over the javadoc and forgot. Updated.

RE class name: I don't have strong feelings either way. Happy to change it back to `BaseCommand`. The `PrintStream` stuff should go away in a future PR when I centralize the lookup of stdout and stderr.

src/main/java/bio/terra/cli/command/helperclasses/CommandSetup.java
marikomedlock: done.

src/main/java/bio/terra/cli/command/helperclasses/CommandSetup.java
marikomedlock: Actually, I think it's better for the command classes to not worry about bash exit codes and just throw exceptions. The exception handling, including updating the error code , happens in `Main` specifically in the `UserActionableAndSystemExceptionHandler` inner class.

src/main/java/bio/terra/cli/command/helperclasses/FormatFlag.java
marikomedlock: Yes, agreed. Added a ticket to track this follow-on work. [PF-603](https://broadworkbench.atlassian.net/browse/PF-603)

src/main/java/bio/terra/cli/command/helperclasses/FormatFlag.java
marikomedlock: Added note to the class comment.

src/main/java/bio/terra/cli/command/helperclasses/FormatFlag.java
marikomedlock: Good idea. done.

src/main/java/bio/terra/cli/command/helperclasses/FormatFlag.java
marikomedlock: ok sounds good. incidentally, the `javadoc` Gradle task is not passing. I added a low-priority ticket to fix this. [PF-604](https://broadworkbench.atlassian.net/browse/PF-604)

src/main/java/bio/terra/cli/command/datarefs/CheckAccess.java
wchamber: for future, for non-static methods I think `this::printText` should work.

marikomedlock: Didn't know about that, thanks! Will update in the coming (part 2) PR.

PR #41, opened 2021-03-24
PF-542: Config command for logging level

marikomedlock:
- Added two new config commands: `terra config set logging`, `terra config get-value logging`.
- The `set` command allows updating the logging level in the global context without re-compiling.
Screen Shot 2021-03-24 at 11 31 06 AM

- Also fixed a small un-related bug. The `terra config set image` command was not updating the image path in the global context if the image had not yet been downloaded with `docker pull`. Changed the behavior to always update the global context property, but log a warning if the image is not found.


wchamber (approved):
cool!

jmczerk (approved):
LGTM

PR #42, opened 2021-03-24
PF-446: Common base class for commands. (part 2)

marikomedlock:
This is part 2 of the changes to have:
- optionally include a `--format=json` flag.

This PR includes changes to the `config` and `groups` commands. Also:
- Removed the output for the `config set` commands. To see the current/updated values, users can call the `config get-value` commands, instead. I think this makes it a little more consistent (i.e. always call `get-value` instead of relying sometimes on the output of `set`).

### DIFF
--- a/src/main/java/bio/terra/cli/command/config/List.java
+++ b/src/main/java/bio/terra/cli/command/config/List.java
@@ -1,27 +1,49 @@
 package bio.terra.cli.command.config;
 
-import bio.terra.cli.context.GlobalContext;
-import java.util.concurrent.Callable;
+import static bio.terra.cli.command.config.getvalue.Logging.LoggingReturnValue;
+
+import bio.terra.cli.command.helperclasses.BaseCommand;
+import bio.terra.cli.command.helperclasses.FormatOption;
+import java.util.HashMap;
+import java.util.Map;
 import picocli.CommandLine;
 
 /** This class corresponds to the third-level "terra config list" command. */
 @CommandLine.Command(
     name = "list",
     description = "List all configuration properties and their values.")
-public class List implements Callable<Integer> {
+public class List extends BaseCommand {
 
+  @CommandLine.Mixin FormatOption formatOption;
+
+  /** Print out a list of all the config properties. */
   @Override
-  public Integer call() {
-    GlobalContext globalContext = GlobalContext.readFromFile();
+  protected void execute() {
+    // build a map of the config properties
+    LoggingReturnValue loggingLevels =
+        new LoggingReturnValue(globalContext.consoleLoggingLevel, globalContext.fileLoggingLevel);
+    Map<String, Object> configPropertyMap = new HashMap<>();
+    configPropertyMap.put("browser", globalContext.browserLaunchOption);
+    configPropertyMap.put("image", globalContext.dockerImageId);
+    configPropertyMap.put("logging", loggingLevels);
 
-    System.out.println("[browser] browser launch for login = " + globalContext.browserLaunchOption);
-    System.out.println("[image] docker image id = " + globalContext.dockerImageId);
-    System.out.println(
+    formatOption.printReturnValue(configPropertyMap, this::printText);
+  }
+
+  /** Print this command's output in text format. */
+  private void printText(Map<String, Object> returnValue) {
+    OUT.println("[browser] browser launch for login = " + globalContext.browserLaunchOption);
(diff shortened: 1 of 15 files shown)


src/main/java/bio/terra/cli/command/config/List.java
```diff
@@ -1,27 +1,49 @@
 package bio.terra.cli.command.config;
 
-import bio.terra.cli.context.GlobalContext;
-import java.util.concurrent.Callable;
```
jmczerk, line 4: This looks like a "good" use of static imports to me, but I'm interested to see if @wchamber agrees:

go/java-practices/imports#static-good

jmczerk (approved):
LGTM

src/main/java/bio/terra/cli/command/config/List.java
```diff
@@ -1,27 +1,49 @@
...
-    System.out.println(
+  /** Print this command's output in text format. */
+  private void printText(Map<String, Object> returnValue) {
+    OUT.println("[browser] browser launch for login = " + globalContext.browserLaunchOption);
+    OUT.println("[image] docker image id = " + globalContext.dockerImageId);
+    OUT.println(
         "[logging] console logging level = "
             + globalContext.consoleLoggingLevel
             + ", file logging level = "
             + globalContext.fileLoggingLevel);
```
wchamber, line 41: worth a comment that we're ignoring the `returnValue`, or consider making a return value class for the config that we can use.

wchamber, line 41: should we reuse the `printText` for the `LoggingReturnValue`?

src/main/java/bio/terra/cli/command/config/getvalue/Logging.java
```diff
@@ -1,28 +1,56 @@
...
+    public LoggingReturnValue(
+        Logger.LogLevel consoleLoggingLevel, Logger.LogLevel fileLoggingLevel) {
+      this.consoleLoggingLevel = consoleLoggingLevel;
+      this.fileLoggingLevel = fileLoggingLevel;
+    }
+  }
 
-    System.out.println(
+  /** Print this command's output in text format. */
+  public static void printText(LoggingReturnValue returnValue) {
```
wchamber, line 40: consider making this a non-static member function of LoggingReturnValue since we're exposing it and it's associated strongly with that class. Or does that mess with jackson? Maybe can use `LoggingReturnValue::printText` for the `Consumer<LoggingReturnValue>`?

src/main/java/bio/terra/cli/command/config/set/Browser.java
```diff
@@ -1,38 +1,30 @@
...
+  }
 
-    System.out.println(
-        "Browser launch mode for login is "
-            + globalContext.browserLaunchOption
-            + " ("
-            + (globalContext.browserLaunchOption == prevBrowserLaunchOption
-                ? "UNCHANGED"
-                : "CHANGED")
-            + ").");
```
wchamber, line 34: just curious, no more output? Some feedback to the user seems reasonable. Ditto on other classes.

wchamber (approved):

src/main/java/bio/terra/cli/command/config/List.java
marikomedlock: Good idea. Added a custom return value class instead of relying on a Java Map.

marikomedlock: Yes, done.

src/main/java/bio/terra/cli/command/config/getvalue/Logging.java
marikomedlock: I don't think it would mess with Jackson, but if it's a non-static function, then we don't need to pass the `LoggingReturnValue` instance as an argument. But that argument is required by the `Consumer` function that the `formatOption` method expects.

I could make it a static method the `LoggingReturnValue` class instead, but that's not really consistent with how I've defined them everywhere else, where the return value object definition is also tightly coupled to the method that prints it out. I don't have a strong opinion either way, but prefer consistency with the other command classes for now.

src/main/java/bio/terra/cli/command/config/set/Browser.java
marikomedlock: You know, I had just convinced myself that this was not really useful. Maybe I've been staring at it too long. Happy to put it back if you think it's helpful.

PR #44, opened 2021-03-24
PF-446: Common base class for commands. (part 4)

marikomedlock:
This is part 4 of the changes to have:
- optionally include a --format=json flag

This PR includes changes to the `spend` and `workspace` commands.

### DIFF
--- a/src/main/java/bio/terra/cli/command/workspace/Create.java
+++ b/src/main/java/bio/terra/cli/command/workspace/Create.java
@@ -1,28 +1,30 @@
 package bio.terra.cli.command.workspace;
 
 import bio.terra.cli.auth.AuthenticationManager;
-import bio.terra.cli.context.GlobalContext;
-import bio.terra.cli.context.WorkspaceContext;
+import bio.terra.cli.command.helperclasses.BaseCommand;
+import bio.terra.cli.command.helperclasses.FormatOption;
 import bio.terra.cli.service.WorkspaceManager;
-import java.util.concurrent.Callable;
+import bio.terra.workspace.model.WorkspaceDescription;
+import picocli.CommandLine;
 import picocli.CommandLine.Command;
 
 /** This class corresponds to the third-level "terra workspace create" command. */
 @Command(name = "create", description = "Create a new workspace.")
-public class Create implements Callable<Integer> {
+public class Create extends BaseCommand {
 
+  @CommandLine.Mixin FormatOption formatOption;
+
+  /** Create a new workspace. */
   @Override
-  public Integer call() {
-    GlobalContext globalContext = GlobalContext.readFromFile();
-    WorkspaceContext workspaceContext = WorkspaceContext.readFromFile();
-
-    AuthenticationManager authenticationManager =
-        new AuthenticationManager(globalContext, workspaceContext);
-    authenticationManager.loginTerraUser();
+  protected void execute() {
     new WorkspaceManager(globalContext, workspaceContext).createWorkspace();
-    authenticationManager.fetchPetSaCredentials(globalContext.requireCurrentTerraUser());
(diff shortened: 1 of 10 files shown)


src/main/java/bio/terra/cli/command/workspace/Create.java
```diff
@@ -1,28 +1,30 @@
...
+  @CommandLine.Mixin FormatOption formatOption;
 
-    AuthenticationManager authenticationManager =
-        new AuthenticationManager(globalContext, workspaceContext);
-    authenticationManager.loginTerraUser();
+  /** Create a new workspace. */
+  @Override
+  protected void execute() {
     new WorkspaceManager(globalContext, workspaceContext).createWorkspace();
-    authenticationManager.fetchPetSaCredentials(globalContext.requireCurrentTerraUser());
```
jmczerk, line 23: Had trouble tracking down the impl of `WorkspaceModel` (seems like it might be generated?), does it serialize cleanly to JSON?

src/main/java/bio/terra/cli/command/workspace/ListUsers.java
```diff
@@ -1,32 +1,34 @@
...
+  @CommandLine.Mixin FormatOption formatOption;
 
-    new AuthenticationManager(globalContext, workspaceContext).loginTerraUser();
+  /** List all users of the workspace. */
+  @Override
+  protected void execute() {
     RoleBindingList roleBindings =
         new WorkspaceManager(globalContext, workspaceContext).listUsersOfWorkspace();
-    for (RoleBinding roleBinding : roleBindings) {
-      System.out.println(roleBinding.getRole());
```
jmczerk, line 25: Looks like `RoleBinding` is an `AutoValue`, does this play nice with Jackson?

jmczerk (commented):
Have a couple questions about whether Jackson treats return values properly, but looks good otherwise.

src/main/java/bio/terra/cli/command/workspace/Create.java
marikomedlock: Yes, it's a generated class in the WSM client library. It does serialize nicely

src/main/java/bio/terra/cli/command/workspace/ListUsers.java
marikomedlock: I'm not sure it's an `AutoValue` -- where do you get that from? This is another generated class but from the SAM client library.
```
> terra workspace list-users --format=json
[ {
  "role" : null,
  "members" : [ ]
}, {
  "role" : "READER",
  "members" : [ ]
}, {
  "role" : "WRITER",
  "members" : [ ]
}, {
  "role" : "OWNER",
  "members" : [ "mmdevverily@gmail.com" ]
} ]
```

jmczerk: Ah, I was looking at the wrong class: https://github.com/DataBiosphere/terra-workspace-manager/blob/dev/src/main/java/bio/terra/workspace/service/iam/model/RoleBinding.java

Thanks for the output!

jmczerk (approved):

src/main/java/bio/terra/cli/command/workspace/ListUsers.java
marikomedlock: Actually, this is a great point. Multiple client libraries have very similar model classes (e.g. RoleBinding, Status) and it would be great if we could share generated classes. I think this has come up for discussion before, but it was slightly complicated/unclear how to share snippets of the OpenAPI yaml specification that defines the generated classes.

Not sure if there's a ticket for this, but perhaps worth checking what the status of that was at standup today.

wchamber (approved):

PR #45, opened 2021-03-24
PF-446: Common base class for commands. (part 5)

marikomedlock:
This is **part 5** of the changes to have:
- optionally include a --format=json flag

This PR is the **final** change. It moves logging setup from the `Main` picocli entrypoint class into the `BaseCommand` class. Logging levels are stored in the global context, so we need to setup logging after reading in the global context file. This change moves the logging setup into `BaseCommand` so that we're not reading in the global context file twice for each command.

**NOTE** All the previous PRs (parts 2,3, and 4) can be merged in any order. This PR (part 5) must be merged last. Otherwise, commands that are not yet migrated to use the common base class will not respect the logging levels set in the global context.

### DIFF
--- a/src/main/java/bio/terra/cli/command/helperclasses/BaseCommand.java
+++ b/src/main/java/bio/terra/cli/command/helperclasses/BaseCommand.java
@@ -27,7 +27,7 @@ public abstract class BaseCommand implements Callable<Integer> {
 
   @Override
   public Integer call() {
-    // read in the global and setup logging
+    // read in the global context and setup logging
     globalContext = GlobalContext.readFromFile();
     new Logger(globalContext).setupLogging();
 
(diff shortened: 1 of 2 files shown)


src/main/java/bio/terra/cli/command/helperclasses/BaseCommand.java
```diff
@@ -26,8 +27,11 @@
 
   @Override
   public Integer call() {
-    // read in the global and workspace context
+    // read in the global and setup logging
```
jmczerk, line 30: Ultra-tiny nit, feel free to ignore: "global context" (or could just remove the comment altogether).

jmczerk (approved):

src/main/java/bio/terra/cli/command/helperclasses/BaseCommand.java
marikomedlock: Fixed.

wchamber (approved):

PR #47, opened 2021-03-25
PF-586 Add a default notebook instance name for creation.

wchamber:


wchamber:
@deflaux fyi, auto name generation for creating notebook instances following what's done in the cloud console.

src/main/java/bio/terra/cli/command/notebooks/Create.java
```diff
@@ -92,11 +100,33 @@ protected void execute() {
...
+      return rawInstanceName;
+    }
+    String localDateTimeSuffix =
+        DateTimeFormatter.ofPattern(AUTO_NAME_DATE_FORMAT)
+            .format(Instant.now().atZone(ZoneId.systemDefault()));
+    return extractUsername(user.terraUserEmail) + localDateTimeSuffix;
+  }
+
+  private static String extractUsername(String validEmail) {
+    return validEmail.substring(0, validEmail.indexOf('@'));
```
marikomedlock, line 130: Will this work for an email such as `mariko.medlock@gmail.com`? Asking because the option description seems to mean that `.` is not allowed.

wchamber: Good point. Took a stab at handling more email addresses, PTAL.

src/main/java/bio/terra/cli/command/notebooks/Create.java
```diff
@@ -92,11 +101,57 @@ protected void execute() {
...
+  private static String mangleUsername(String username) {
+    // Strip non alpha-numeric or '-' characters.
+    String mangledName = username.replaceAll("[^a-zA-Z0-9-]", "");
+    if (mangledName.isEmpty()) {
+      mangledName = "notebook";
+    }
+    // Lower case everything, even though only the first character requires lowercase.
+    mangledName = mangledName.toLowerCase();
+    // Make sure the returned name isn't too long to not have the date time suffix.
+    int maxNameLength = 61 - AUTO_NAME_DATE_FORMAT.length();
```
marikomedlock, line 147: nit: Where does 61 come from? I think defining that as a static/constant with a descriptive name would be helpful.

marikomedlock (approved):

src/main/java/bio/terra/cli/command/notebooks/Create.java
wchamber: good point, done.

PR #48, opened 2021-03-25
PF-603: Hold pointers to output streams in a central place.

marikomedlock:
- Added a singleton `Printer` utility class that holds pointers to the output streams we can write to.
- This removes the various places that hard-coded `System.out` and `System.err`, so that this is only specified in a single central place. The goal is to make this easy to change (in code or by users via global context) in the future.
- Also changed the Docker command output to write to stderr instead of stdout. This should make it easier for the commands that invoke Docker to also specify a return value by writing to stdout.

(This is a follow-on ticket from the JSON format flag/base command general overhaul/cleanup.)

- Note that one [approach for picocli testing](https://picocli.info/#_black_box_and_white_box_testing) is to specify output streams other than `System.out` and `System.err` in the `CommandLine` setup (i.e. in `Main`). For this reason, I think we want to setup printing in `Main` instead of in `BaseCommand`, so that for testing, we can override the default output streams by constructing our own `CommandLine` class as shown in the documentation.

### DIFF
--- a/src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
+++ b/src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
@@ -4,6 +4,7 @@ import bio.terra.cli.command.exception.SystemException;
 import bio.terra.cli.context.GlobalContext;
 import bio.terra.cli.context.TerraUser;
 import bio.terra.cli.context.WorkspaceContext;
+import bio.terra.cli.context.utils.Printer;
 import com.github.dockerjava.api.DockerClient;
 import com.github.dockerjava.api.async.ResultCallback;
 import com.github.dockerjava.api.command.CreateContainerCmd;
@@ -20,6 +21,7 @@ import com.github.dockerjava.core.DockerClientImpl;
 import com.github.dockerjava.httpclient5.ApacheDockerHttpClient;
 import com.github.dockerjava.transport.DockerHttpClient;
 import java.io.File;
+import java.io.PrintWriter;
 import java.nio.charset.Charset;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -350,12 +352,9 @@ public class DockerAppsRunner {
     @Override
     public void onNext(Frame frame) {
       String logStr = new String(frame.getPayload(), Charset.forName("UTF-8"));
-
-      // TODO: PF-423. Calling sysout.println here breaks the model of printing user-facing output
-      // from the command classes only.
-      // Revisit this once we have better model for centralizing output across all commands/rest of
-      // the codebase.
-      System.out.print(logStr); // write to stdout
+      PrintWriter err = Printer.getErr();
+      err.print(logStr);
+      err.flush();
 
       if (buildSingleStringOutput) {
         log.append(logStr);
--- a/src/main/java/bio/terra/cli/auth/GoogleCredentialUtils.java
+++ b/src/main/java/bio/terra/cli/auth/GoogleCredentialUtils.java
@@ -1,6 +1,7 @@
 package bio.terra.cli.auth;
 
 import bio.terra.cli.command.exception.SystemException;
+import bio.terra.cli.context.utils.Printer;
 import com.google.api.client.auth.oauth2.Credential;
 import com.google.api.client.auth.oauth2.StoredCredential;
 import com.google.api.client.extensions.java6.auth.oauth2.AbstractPromptReceiver;
@@ -22,6 +23,7 @@ import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.io.PrintWriter;
 import java.nio.charset.Charset;
 import java.security.GeneralSecurityException;
 import java.util.List;
@@ -109,8 +111,9 @@ public final class GoogleCredentialUtils {
   private static class NoLaunchBrowser implements AuthorizationCodeInstalledApp.Browser {
     @Override
     public void browse(String url) {
-      System.out.println("Please open the following address in a browser on any machine:");
-      System.out.println("  " + url);
+      PrintWriter out = Printer.getOut();
+      out.println("Please open the following address in a browser on any machine:");
+      out.println("  " + url);
     }
   }
 
--- a/src/main/java/bio/terra/cli/command/groups/ListUsers.java
+++ b/src/main/java/bio/terra/cli/command/groups/ListUsers.java
@@ -33,7 +33,7 @@ public class ListUsers extends BaseCommand {
   /** Print this command's output in text format. */
   private static void printText(java.util.List<String> returnValue) {
     for (String user : returnValue) {
-      System.out.println(user);
+      OUT.println(user);
     }
   }
 }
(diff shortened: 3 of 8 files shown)


src/main/java/bio/terra/cli/context/utils/Printer.java
```diff
@@ -0,0 +1,89 @@
...
+    return printer.outWriter;
+  }
+
+  /**
+   * Utility method to get the error stream from the singleton.
+   *
+   * @return stream to write errors and running status (e.g. stderr)
+   */
+  public static PrintWriter getErr() {
+    if (printer == null) {
```
jmczerk, line 78: Should we make this an assert to quickly find any cases that break our assumptions, or are there things like test that might break this invariant in an acceptable way?

jmczerk (approved):

src/main/java/bio/terra/cli/context/utils/Printer.java
```diff
@@ -0,0 +1,89 @@
...
+import java.nio.charset.Charset;
+import org.slf4j.LoggerFactory;
+import picocli.CommandLine;
+
+/**
+ * Singleton class for holding a reference to output streams (e.g. stdout, stderr). The purpose of
+ * holding these references in a single place is so that we can write output throughout the codebase
+ * without passing around the output streams from the top-level command classes.
+ */
+public class Printer {
```
jaycarlton, line 14: Are you not using spring in the CLI?

src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
```diff
@@ -350,12 +352,9 @@ public LogContainerTestCallback(boolean buildSingleStringOutput, boolean buildFr
     @Override
     public void onNext(Frame frame) {
       String logStr = new String(frame.getPayload(), Charset.forName("UTF-8"));
-
-      // TODO: PF-423. Calling sysout.println here breaks the model of printing user-facing output
-      // from the command classes only.
-      // Revisit this once we have better model for centralizing output across all commands/rest of
-      // the codebase.
-      System.out.print(logStr); // write to stdout
+      PrintWriter err = Printer.getErr();
```
jaycarlton, line 355: I just did a little research into the differences between PrintWriter and PrintStream. It's not obvious why you went with the former here.

src/main/java/bio/terra/cli/auth/GoogleCredentialUtils.java
```diff
@@ -109,8 +111,9 @@ public String getRedirectUri() {
   private static class NoLaunchBrowser implements AuthorizationCodeInstalledApp.Browser {
     @Override
     public void browse(String url) {
-      System.out.println("Please open the following address in a browser on any machine:");
-      System.out.println("  " + url);
+      PrintWriter out = Printer.getOut();
+      out.println("Please open the following address in a browser on any machine:");
```
jaycarlton, line 115: And do what? For what purpose?

src/main/java/bio/terra/cli/command/groups/ListUsers.java
```diff
@@ -33,7 +33,7 @@ protected void execute() {
   /** Print this command's output in text format. */
   private static void printText(java.util.List<String> returnValue) {
     for (String user : returnValue) {
-      System.out.println(user);
+      OUT.println(user);
```
jaycarlton, line 36: Do the subcommands use this static variable, or should they call `getOut()`?

src/main/java/bio/terra/cli/command/helperclasses/BaseCommand.java
```diff
@@ -18,15 +21,31 @@
...
+            + "the command classes, we can't set the output streams in the constructor and make them "
+            + "static.")
 public abstract class BaseCommand implements Callable<Integer> {
   protected GlobalContext globalContext;
   protected WorkspaceContext workspaceContext;
 
-  // output stream to use for writing command return value
-  protected static final PrintStream OUT = System.out;
+  // output streams for commands to write to
+  protected static PrintWriter OUT;
```
jaycarlton, line 39: This may be the first time I remember seeing `protected static`. You could alternatively have a protected method like `getOutWriter()` that returned the OUT property, which might be more OOPy. Then these would be private instance members and you could avoid some awkwardness. They'd just happen to point to static resources.

jaycarlton (approved):

src/main/java/bio/terra/cli/context/utils/Printer.java
marikomedlock: I thought about that, but decided on logging a warning instead of throwing an exception, because I didn't want to suppress any output or errors just because they happened before printing was setup. I figured that would just make debugging a two-step process.

There are no tests for the CLI yet -- that was considered non-critical for the alpha 1 demo, but (I think) should be critical for alpha 2 (ticket is [PF-498](https://broadworkbench.atlassian.net/browse/PF-498)). I think the only case that would trigger this null check is if we tried to get the Printer streams in the Main.class main method before `Printer.setupPrinting(cmd);`. This method is also the picocli entrypoint into the operations, so this shouldn't happen unless we add new code there later.

src/main/java/bio/terra/cli/context/utils/Printer.java
marikomedlock: No, the CLI doesn't use Spring. Why do you ask?

src/main/java/bio/terra/cli/apps/DockerAppsRunner.java
marikomedlock: I chose PrintWriter because I expect the CLI output to be primarily character data. I would have chosen PrintStream if I was planning to write out binary data.

src/main/java/bio/terra/cli/auth/GoogleCredentialUtils.java
marikomedlock: To walk through the Google OAuth flow. The default login flow for the CLI is to type `terra auth login` and that launches a browser window on your machine that walks through the Google OAuth flow and sends a code back to the CLI that it can use to get access tokens on your behalf. However, this caused problems on machines that don't have a browser installed by default (e.g. notebook VMs). So, this section of code skips the browser launch and instead just prints out a URL that the user can open in a browser themselves. Importantly, this browser does not need to be on the same machine where they are running the CLI. Then they walk through the OAuth flow and manually copy/paste a code back into the CLI.

src/main/java/bio/terra/cli/command/groups/ListUsers.java
marikomedlock: As long as the sub-commands extend the `BaseCommand` class, they should just use the static variable. This variable is set in `BaseCommand` by calling `getOut()`, so there's no need for all sub-classes to repeat the call.

src/main/java/bio/terra/cli/command/helperclasses/BaseCommand.java
marikomedlock: I prefer having sub-commands access these as properties instead of methods. `OUT.println("...command output...")` seems more readable to me than `getOutWriter().println("..command output..")`, and more similar to `System.out.println("..command output..")`, which is what this is replacing.

PR #50, opened 2021-04-08
PF-663 Add support for big query datasets as another resource type.

wchamber:
This add support for using BigQuery datasets in both the `resources` and `data-refs` commands.

This is our second resource type. I'm not really sold on our interfaces with the `WorkspaceManager` class and the switch statements everywhere. On the other hand, this is likely to get a major revision once we switch to using WM APIs instead of direct cloud access for create/delete/list, so I'm not sure how much investment to put in at this point.


jmczerk (approved):
LGTM (assuming this is all tested manually)

PR #53, opened 2021-04-09
PF-662 add a 'resources resolve' command.

wchamber:
This is a convenience command for getting the cloud id of a controlled resource. It parallels the 'data-refs resolve' command.


jmczerk (approved):
Guess that was pretty straightforward! 😄

PR #54, opened 2021-04-09
[PF-634] Add verily-cli environment

zloery:
This adds configuration for the newly created `verilycli` personal environment. Unlike other personal environments, this WSM instance points to dev RBS, so projects will not be cleaned up by Janitor. This is intended to be similar to dev, but will be manually synced instead of automatically updated every time WSM head is changed. This should give us more stability for development without blocking backwards-incompatible WSM changes.

In the near-future, this will be set as the default WSM server. I'm not doing that yet so we have some time to transition existing workspaces.

From CONTRIBUTING.md it seems like I don't need to do any manual publishing, it should all be handled by GHA. Is that true?

### DIFF
--- a/src/main/resources/servers/all-servers.json
+++ b/src/main/resources/servers/all-servers.json
@@ -1,1 +1,1 @@
-[ "terra-dev.json", "terra-verily-dev.json", "mm-dev.json", "wchamber-dev.json" ]
+[ "terra-dev.json", "terra-verily-dev.json", "mm-dev.json", "verily-cli.json", "wchamber-dev.json" ]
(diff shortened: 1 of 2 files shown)


wchamber (approved):
Thanks! Hopefully you were able to install and try out the cli.
Yes, the publishing will happen automatically, pushing a new binary and related docker image.
https://github.com/DataBiosphere/terra-cli/blob/main/.github/workflows/on-pr-merge.yml

zloery:
Yep, both the install and local development documentation were great! props on making that so straightforward

PR #55, opened 2021-04-15
PF-634 Change default server to verily-cli.

wchamber:


zloery (approved):
LGTM. I saw Nicole's message earlier, are we still waiting for anyone else to switch over before making breaking changes?

wchamber:
> LGTM. I saw Nicole's message earlier, are we still waiting for anyone else to switch over before making breaking changes?

I don't think so. Nicole's the main user atm, Matt and Willy are focused on other things. John, Mariko, and I probably have some stray workspaces in dev, but I don't think we need the CLI to do cleanup as needed - we can use the updated WM swagger.

PR #56, opened 2021-04-20
PF-611: Improve error message when Docker isn't running.

marikomedlock:
1. Docker connection failures now return a more readable error message: "Connecting to Docker daemon failed. Check that Docker is installed and running."
  - Added a check to all exceptions thrown by executing a Docker command. If the cause is an IOException with a "Connection refused" message, then we assume that Docker is either not installed or not running.
  - Docker connection failures are considered UserActionableExceptions, while all other Docker exceptions are considered SystemExceptions (i.e. probably not something the user can fix themselves).
  - Below is the error message and bash exit code when I quit Docker Desktop on my Mac.
Screen Shot 2021-04-20 at 6 32 07 PM

2. Fixed a bug where the CLI was not setting the exit code correctly.
  - This should fix the issue with the install and local-dev scripts not exiting with an error code when docker is not installed. Now both scripts should error out on the `terra config set image --default` line.
  - Older versions of picocli called `System.exit` for us, but the most recent version [expects us to call it explicitly](https://picocli.info/#_exit_code).
  - 0 = success, 1 = user-actionable error, 2 = system error, 3 = unexpected error
  - I didn't include the exit codes in the usage help, because I don't think it's very useful information to end users (i.e. I think zero/non-zero is the only meaningful distinction from user's perspective, not the difference between 1/2/3). Reviewers, I'd appreciate opinions on that though. Here's what it would look like if I included the exit code information in the usage help:
Screen Shot 2021-04-20 at 6 27 16 PM

3. Refactored the `DockerAppsRunner` class into two separate classes.
  - `DockerUtils` handles executing Docker commands and parsing any exceptions they throw.
  - `AppsRunner` handles the Terra-related parts of executing apps in a Docker container.

### DIFF
--- a/src/main/java/bio/terra/cli/apps/AppsRunner.java
+++ b/src/main/java/bio/terra/cli/apps/AppsRunner.java
@@ -4,28 +4,8 @@ import bio.terra.cli.command.exception.SystemException;
 import bio.terra.cli.context.GlobalContext;
 import bio.terra.cli.context.TerraUser;
 import bio.terra.cli.context.WorkspaceContext;
-import bio.terra.cli.context.utils.Printer;
-import com.github.dockerjava.api.DockerClient;
-import com.github.dockerjava.api.async.ResultCallback;
-import com.github.dockerjava.api.command.CreateContainerCmd;
-import com.github.dockerjava.api.command.CreateContainerResponse;
-import com.github.dockerjava.api.command.WaitContainerResultCallback;
-import com.github.dockerjava.api.exception.NotFoundException;
-import com.github.dockerjava.api.model.Bind;
-import com.github.dockerjava.api.model.Frame;
-import com.github.dockerjava.api.model.HostConfig;
-import com.github.dockerjava.api.model.Volume;
-import com.github.dockerjava.core.DefaultDockerClientConfig;
-import com.github.dockerjava.core.DockerClientConfig;
-import com.github.dockerjava.core.DockerClientImpl;
-import com.github.dockerjava.httpclient5.ApacheDockerHttpClient;
-import com.github.dockerjava.transport.DockerHttpClient;
-import java.io.File;
-import java.io.PrintWriter;
-import java.nio.charset.Charset;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -36,12 +16,12 @@ import org.slf4j.LoggerFactory;
  * This class runs client-side tools and manipulates the tools-related properties of the global
  * context object.
  */
-public class DockerAppsRunner {
-  private static final Logger logger = LoggerFactory.getLogger(DockerAppsRunner.class);
+public class AppsRunner {
+  private static final Logger logger = LoggerFactory.getLogger(AppsRunner.class);
 
   private final GlobalContext globalContext;
   private final WorkspaceContext workspaceContext;
-  private DockerClient dockerClient;
+  private final DockerUtils dockerUtils = new DockerUtils();
 
   // default $HOME directory on the container (this is where we expect to look for the global
   // context)
@@ -49,10 +29,9 @@ public class DockerAppsRunner {
   // mount point for the workspace directory
   private static final String CONTAINER_WORKSPACE_DIR = "/usr/local/etc";
 
-  public DockerAppsRunner(GlobalContext globalContext, WorkspaceContext workspaceContext) {
+  public AppsRunner(GlobalContext globalContext, WorkspaceContext workspaceContext) {
     this.globalContext = globalContext;
     this.workspaceContext = workspaceContext;
-    this.dockerClient = buildDockerClient();
   }
 
   // ====================================================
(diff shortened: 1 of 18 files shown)


src/main/java/bio/terra/cli/apps/AppsRunner.java
```diff
@@ -36,23 +16,22 @@
  * This class runs client-side tools and manipulates the tools-related properties of the global
  * context object.
  */
-public class DockerAppsRunner {
-  private static final Logger logger = LoggerFactory.getLogger(DockerAppsRunner.class);
+public class AppsRunner {
```
wchamber, line 19: nit: I'm not quite following the "AppsRunner" name. This class seems all about creating the mappings and bindings from the user directories, env vars, workspace context, etc. to the docker environment. I think it's the `runToolCommand` is what's "running the apps", so tools = apps? Not sure what's a better name. I kind of want to focus on the "docker" part instead of what "tool" might be used in the container. brainstorming, `DockerRunner` `DockerCommandRunner` `DockerBinding` idk
No action required, pick your favorite naming.

src/main/java/bio/terra/cli/apps/DockerUtils.java
```diff
@@ -0,0 +1,244 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+/** This class provides utility methods for running Docker containers. */
+public class DockerUtils {
+  private final DockerClient dockerClient;
+
+  public DockerUtils() {
+    this.dockerClient = DockerUtils.buildDockerClient();
```
wchamber, line 35: Little surprising to have a `Utils` class not be static. Maybe just have our own `DockerClient` class or `DockerClientWrapper`?

src/main/java/bio/terra/cli/apps/DockerUtils.java
```diff
@@ -0,0 +1,244 @@
...
+          .withFollowStream(true)
+          .withTailAll()
+          .exec(new LogContainerTestCallback());
+    } catch (RuntimeException rtEx) {
+      throw wrapExceptionIfDockerConnectionFailed(rtEx);
+    }
+  }
+
+  /** Helper class for reading Docker container logs into a string. */
+  private static class LogContainerTestCallback extends ResultCallback.Adapter<Frame> {
```
wchamber, line 171: nit: maybe `LogCallback`? Not sure what's `Test` about this.

wchamber (approved):
Just some naming nits. Welcome back!

jmczerk (approved):

src/main/java/bio/terra/cli/apps/AppsRunner.java
marikomedlock: Good idea. I like the `DockerCommandRunner` suggestion.

src/main/java/bio/terra/cli/apps/DockerUtils.java
marikomedlock: Changed to `DockerClientWrapper`.

src/main/java/bio/terra/cli/apps/DockerUtils.java
marikomedlock: Agreed, I think that's leftover from some debugging when I originally wrote this. Changed to `LogContainerCommandCallback`.

src/main/java/bio/terra/cli/command/Main.java
```diff
@@ -116,20 +119,28 @@ public void run() {}
             .errors(CommandLine.Help.Ansi.Style.fg_red, CommandLine.Help.Ansi.Style.bold)
             .build();
```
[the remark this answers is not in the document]
marikomedlock: I'm not convinced of the value of aligning with the posix exit codes, because I don't think there are great analogs. Our 4 possible exit codes are: success, user-actionable exception, system exception, unexpected exception. None of the exceptions consistently map to the cause (e.g. docker connection error is a user-actionable exception, but so is an invalid argument). 

I'm certainly open to discussion here though if you or anyone else thinks I should reconsider. Let me know if you think this warrants a post stand-up discussion. If we choose to go the posix error code route, I do think we should change the mapping.`EAGAIN` means to try again, which I would not expect to fix a CLI `SystemException`.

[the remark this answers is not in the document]
jmczerk: I'm fine with not using posix error codes, the thinking was just that scripting languages might be able to turn them into meaningful exceptions (I've used libraries to map the return codes of syscalls to exception types in python, for instance).

The bigger concern is just that the meaning is documented so that people using the tool programmatically can interpret the return codes, since there are more than one value.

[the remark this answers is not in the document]
marikomedlock: RE scripting languages auto-generating exceptions: Makes sense. I'm not sure what action the caller could take on catching one of these exceptions from the CLI though -- that's my hesitancy to mapping to posix.

RE documentation: Do I understand correctly then that you would vote for including exit codes in the main usage help? If so, I will include that change as an add-on to my current PR. Sorry, I think I just read past that in your original comment, and am now re-reading it.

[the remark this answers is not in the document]
jmczerk: No worries...  including in either one or both of README.md and/or usage help would be fine with me.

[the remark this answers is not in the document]
marikomedlock: Sounds good. Created follow-on ticket [PF-704](https://broadworkbench.atlassian.net/browse/PF-704). Thanks!

PR #57, opened 2021-04-21
PF-598: Update server specification on install

marikomedlock:
- Included the server setting into the `terra config` group of commands.
  - `terra config server set` duplicates the existing `terra server set`. So the server setting can now be changed in the general `config` group of commands, or in the `server` group of commands.
  - `terra config get-value server` is a new command to fetch the current server name.
  - `terra config list` includes the current server name.
- Updated the server specification in the install and local-dev scripts.
  - The scripts now look up the current server, and call `terra config server set` with that value.
  - This will not change the server, but will pick up any changes to the server specification files in the code (i.e. in the `resources/server/*.json` files).

### DIFF
--- a/src/main/java/bio/terra/cli/command/Server.java
+++ b/src/main/java/bio/terra/cli/command/Server.java
@@ -1,7 +1,6 @@
 package bio.terra.cli.command;
 
 import bio.terra.cli.command.server.List;
-import bio.terra.cli.command.server.Set;
 import bio.terra.cli.command.server.Status;
 import picocli.CommandLine.Command;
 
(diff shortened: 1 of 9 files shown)


wchamber (approved):
Moving the `terra server set` to be under a config subcommand seems less discoverable to me. I think there's a consistency argument for finding everything configured about the CLI in one space, but is there some way to point at `terra config set server` from the `terra server` commands, or just support both?

marikomedlock:
> Moving the `terra server set` to be under a config subcommand seems less discoverable to me. I think there's a consistency argument for finding everything configured about the CLI in one space, but is there some way to point at `terra config set server` from the `terra server` commands, or just support both?

Supporting both sounds like a good idea -- I will look into it. I reasoned that this is similar to `gcloud auth list` and `gcloud config set account`, but actually I find that to be one of the least intuitive parts of `gcloud` that you have to `list` and `set` a property under different command groupings. So maybe not the best `gcloud` example to mimic.

PR #58, opened 2021-04-22
PF-665: Install gcloud SDK from versioned archive.

marikomedlock:
Install the gcloud SDK in the Docker image by downloading a versioned archive, instead of using `apt-get`. Gcloud SDK [documentation](https://cloud.google.com/sdk/docs/install) says that the last 10 releases will be available via `apt-get`. This change allows us to pin the version indefinitely, instead of getting errors every ~10 weeks when the version we pinned is no longer available via `apt-get`. We should still try to keep this up to date, but don't want to cause breakages in case we forget.

I considered always using the latest version of gcloud sdk, but decided against it because that's inconsistent with how we handle most other dependencies (e.g. in `build.gradle` and `nextflow` in the Docker image). As Will pointed out, it's also clunkier to track which CLI release uses which version of the gcloud sdk. Not impossible, since the Docker image for each CLI release is preserved in GCR, so we could execute `gcloud version` in a container that uses that image, but it's not in GitHub history like other dependencies.

Also bumped the gcloud sdk version to the latest as part of this PR (335 -> 337).

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -4,20 +4,23 @@ FROM openjdk:11.0-jre-slim
 # working directory when the docker is run
 WORKDIR /usr
 
-# install curl
+# install curl (to download install packages) and python (for gcloud sdk)
 RUN apt-get update \
-    && apt-get install -y curl
+    && apt-get install -y curl \
+    && apt-get install -y python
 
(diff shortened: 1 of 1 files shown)


wchamber (approved):
thanks!

PR #59, opened 2021-04-22
PF-620: Invite user if `terra spend enable` fails with user not found in SAM.

marikomedlock:
If `terra spend enable` fails with a user not found in SAM error, then try to invite the user before enabling spend.

This is similar behavior to when:
  - `terra workspace add-user` fails with a user not found in SAM error. In this case, we also try to invite the user before adding them to the workspace.
  - `terra auth login` or any command requiring an active login fails with a user not found in SAM error. In this case, we try to register the user before executing the command.

### DIFF
--- a/src/main/java/bio/terra/cli/service/utils/SamService.java
+++ b/src/main/java/bio/terra/cli/service/utils/SamService.java
@@ -350,12 +350,15 @@ public class SamService {
    * "/api/resources/v1/{resourceTypeName}/{resourceId}/policies/{policyName}/memberEmails/{email}"
    * PUT endpoint to add an email address to a resource + policy.
    *
+   * <p>If that returns a Not Found error, then call the SAM "/api/users/v1/invite/{inviteeEmail}"
+   * endpoint to invite the user.
+   *
    * @param resourceType type of resource
    * @param resourceId id of resource
    * @param resourcePolicyName name of resource policy
    * @param userEmail email of the user or group to add
    */
-  public void addUserToResource(
+  public void addUserToResourceOrInviteUser(
       String resourceType, String resourceId, String resourcePolicyName, String userEmail) {
     ResourcesApi resourcesApi = new ResourcesApi(apiClient);
     try {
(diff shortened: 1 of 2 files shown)


src/main/java/bio/terra/cli/service/utils/SamService.java
```diff
@@ -366,7 +369,28 @@ public void addUserToResource(
...
+
+        HttpUtils.callWithRetries(
+            () -> {
+              resourcesApi.addUserToPolicy(resourceType, resourceId, resourcePolicyName, userEmail);
+              return null;
+            },
+            SamService::isRetryable);
+      } catch (ApiException | InterruptedException secondEx) {
+        throw new SystemException("Error adding user to SAM resource.", secondEx);
+      }
```
wchamber, line 393: nit: is it worth adding a helper function for this pattern? code the same from the `WorkspaceManagerService#grantIamRole`, but I guess that's going away.
Maybe a samService function that [checks if the user is registered](https://sam.dsde-dev.broadinstitute.org/#/Users/getUserIds) and uses the invite api if not, and we call that before other functions involving a different user? Does add a Sam round trip for the rarer case of the user not existing.
Alternatively, we could add a function with a signature like `HttpUtils.callWithRetries` for `SamService.retryWithInvitedUser` to do the bad request, not found, add the user, retry the first call.

Possibly all overkill at this point.

src/main/java/bio/terra/cli/service/utils/SamService.java
```diff
@@ -366,7 +369,28 @@ public void addUserToResource(
...
-      throw new SystemException("Error adding user to SAM resource.", ex);
+      if (!(ex instanceof ApiException)
+          || (((ApiException) ex).getCode() != HttpStatusCodes.STATUS_CODE_BAD_REQUEST)) {
+        throw new SystemException("Error adding user to SAM resource.", ex);
+      }
+      logger.info("User not found in SAM. Trying to invite a new user.");
+
+      try {
+        // add to resource failed with Not Found error, now try to invite the user and add them to
+        // the resource again
```
wchamber, line 380: looks like it's bad request, not "not found" for this invite the user case.

wchamber (approved):

src/main/java/bio/terra/cli/service/utils/SamService.java
marikomedlock: Fixed.

src/main/java/bio/terra/cli/service/utils/SamService.java
marikomedlock: - The `WorkspaceManagerService#grantIamRole` isn't going away, I don't think?
- I took a shot at defining this as a pattern (try something/handle one-time error/try again) and using it in the three places that do this in CLI (two with invite, one with register). I'm not sure whether it's a net positive though, so I'm going to put it up as a separate PR and see what you think.

PR #60, opened 2021-04-26
PF-713 Add a --post-startup-script option for notebooks create.

wchamber:
Allow the solutions team to iterate on a post-startup-script in the short term. Long term, this will be a parameter on the WM create notebook instance api.

### DIFF
--- a/src/main/java/bio/terra/cli/command/notebooks/Create.java
+++ b/src/main/java/bio/terra/cli/command/notebooks/Create.java
@@ -58,6 +58,13 @@ public class Create extends BaseCommand {
           "Use this VM image family to find the image; the newest image in this family will be used.")
   private String vmImageFamily;
 
+  @CommandLine.Option(
+      names = "--post-startup-script",
+      description =
+          "Path to a Bash script that automatically runs after a notebook instance fully boots up. "
+              + "The path must be a URL or Cloud Storage path, e.g. 'gs://path-to-file/file-name'")
+  private String postStartupScript;
+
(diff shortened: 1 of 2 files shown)


src/main/java/bio/terra/cli/command/notebooks/Create.java
```diff
@@ -102,6 +109,10 @@ protected void execute() {
     // each zone.
     envVars.put("SUBNET", "projects/" + projectId + "/regions/" + zone + "/subnetworks/subnetwork");
     envVars.put("TERRA_WORKSPACE_ID", workspaceContext.getWorkspaceId().toString());
+    if (postStartupScript != null) {
```
marikomedlock, line 112: What happens if the url/path is invalid or not accessible? Just thinking that if it's an obscure error message, would it be worth checking that we can access that script here?

marikomedlock (approved):

src/main/java/bio/terra/cli/command/notebooks/Create.java
wchamber: The notebook instance will be created successfully and the script will silently fail to run.
While right now the pet SA is used on the VM (and its credentials are used to fetch to postStartupScript), when we switch to WM, it will be a new SA on the VM. So we shortly won't be able to validate for the user that their script is accessible by a SA that doesn't exist yet.
I'm inclined to leave as-is for now, and we'll keep an eye on the usability of this.

PR #64, opened 2021-04-28
PF-705: Call WSM for referenced + controlled resources.

marikomedlock:
- Changed the CLI to call WSM endpoints for managing referenced and controlled resources, instead of hacking it in the CLI code directly.
  - Use a local cache of all resources in a workspace. The purpose of this cache is to avoid a round-trip to WSM, which does its own round-trip to SAM, each time we want to run tools. (We pass a list of environment variables to the Docker container so that tools running there can use the references.) Currently, the cache gets updated as part of each `terra resources` command. That may be too frequent and we can reduce it, but the main goal for now is to make running `terra gsutil`, `terra bq`, `terra nextflow`, `terra gcloud` not any slower.
  - Bumped the WSM client library version to 0.22.0-SNAPSHOT, which includes the new create BQ dataset controlled resource endpoint.
  - The only `resources`-related commands that are not yet calling WSM endpoints are deleting a controlled BQ dataset resource, and checking whether a user has access to a referenced resource. Both are planned additions to WSM. The `check-access` command may undergo some restructuring depending on what the WSM endpoints turn out to be.
- Reorganized the `data-refs` and `resources` command groups into a single `resources` group that handles both referenced and controlled resources. I think this better matches the way WSM treats resources and reduces code duplication.
  - Three of the commands have sub-commands per resource type: `terra resources add-ref`, `terra resources create`, `terra resources resolve`.
Screen Shot 2021-04-28 at 2 30 42 PM
Screen Shot 2021-04-28 at 2 32 08 PM
Screen Shot 2021-04-28 at 2 32 41 PM
Screen Shot 2021-04-28 at 2 33 13 PM

Code changes fall into three main categories:
- Calling the new WSM endpoints. (Mostly in `WorkspaceManagerService`, `WorkspaceManager`, `WorkspaceContext`).
- Removing the temporary resource management previously done directly in the CLI. (Mostly in `WorkspaceManager`, `WorkspaceContext`, `CloudResource`, and the `bio.terra.cli.command.datarefs` package.)
- Restructuring the `terra resources` commands. (Mostly in the `bio.terra.cli.command.resources` package.)

I'm happy to discuss over GVC if that would be helpful for reviewing. I originally planned to do this in two parts (referenced, then controlled resources), but I think it was more confusing to try and maintain both ways of bookkeeping the list of references at the same time.


src/main/java/bio/terra/cli/command/resources/Delete.java
```diff
@@ -1,27 +1,72 @@
...
+          "Stewardship type not supported: " + resourceToDelete.getMetadata().getStewardshipType());
+    }
+
+    formatOption.printReturnValue(resourceToDelete, Delete::printText);
+  }
+
+  /** Print this command's output in text format. */
+  private static void printText(ResourceDescription returnValue) {
+    OUT.println("Successfully deleted resource.");
+    bio.terra.cli.command.resources.Describe.printText(returnValue);
```
wchamber, line 70: I think it might be nice to have a central file for defining the ways to print the different models returned by WM. I would guess we eventually want to print the same models for multiple commands. While it's fine for commands to have their own custom printing of models as needed, a central place would help promote code re-use.

src/main/java/bio/terra/cli/command/resources/Describe.java
```diff
@@ -2,35 +2,73 @@
...
     formatOption.printReturnValue(resource, Describe::printText);
   }
 
   /** Print this command's output in text format. */
-  private static void printText(CloudResource returnValue) {
-    OUT.println("Name: " + returnValue.name);
-    OUT.println("Type: " + returnValue.type);
-    OUT.println("Cloud Id: " + returnValue.cloudId);
+  public static void printText(ResourceDescription returnValue) {
+    OUT.println("Name:         " + returnValue.getMetadata().getName());
```
wchamber, line 30: Probably not this PR, but if we're going to do whitespace pretty printing, it's probably worth doing it programmatically instead of hard coding the amount of whitespace to pad by.

src/main/java/bio/terra/cli/command/resources/create/BqDataset.java
```diff
@@ -0,0 +1,63 @@
...
+      names = "--access",
+      description = "Access scope for the resource: ${COMPLETION-CANDIDATES}")
+  private AccessScope access = AccessScope.SHARED_ACCESS;
+
+  @CommandLine.Option(names = "--dataset-id", required = true, description = "Big Query dataset id")
+  private String bigQueryDatasetId;
+
+  @CommandLine.Option(
+      names = "--location",
+      description = "Bucket location (https://cloud.google.com/storage/docs/locations)")
```
wchamber, line 43: bigquery location?

src/main/java/bio/terra/cli/service/WorkspaceManager.java
```diff
@@ -228,89 +245,224 @@ public RoleBindingList listUsersOfWorkspace() {
...
+        name,
+        () ->
+            new WorkspaceManagerService(
+                    globalContext.server, globalContext.requireCurrentTerraUser())
+                .createReferencedGcsBucket(
+                    workspaceContext.getWorkspaceId(),
+                    name,
+                    description,
+                    cloningInstructions,
+                    gcsBucketName));
```
wchamber, line 325: I think we should avoid these long parameter list chaining to other services. We could either accept that we're tying our commands to the underlying workspace model objects (returning `ResourceDescription` and taking as an argument a `CreateGcpGcsBucketReferenceRequestBody`), or make our own domain objects holding essentially the same values.
As-is, we're both coupling to the workspace API model objects, but also needing to modify two APIs (WorkspaceManager & WorkspaceManagerService) whenever a new field is added.
I think seeing the CLI as an API wrapper might imply its ok to couple to the API model objects through our command layers, whereas in a normal service I'd be very nervous about that. Open to more discussion.

Don't think this PR needs to align everything on the path we're taking forward, but at least the new/modified methods around resources.

src/main/java/bio/terra/cli/command/resources/create/GcsBucket.java
```diff
@@ -0,0 +1,83 @@
...
+        new WorkspaceManager(globalContext, workspaceContext)
+            .createControlledGcsBucket(
+                name,
+                description,
+                cloning,
+                access,
+                bucketName,
+                storageClass,
+                Collections.emptyList(),
+                location);
```
wchamber, line 74: It might be nice to allow specifying a json file with the creation parameters. But that seems very speculative, so probably best just to keep it in mind as a possible future option.

src/main/java/bio/terra/cli/command/resources/create/GcsBucket.java
```diff
@@ -0,0 +1,83 @@
...
+  @CommandLine.Option(
+      names = "--cloning",
+      description =
+          "Instructions for handling when cloning the workspace: ${COMPLETION-CANDIDATES}")
+  private CloningInstructionsEnum cloning = CloningInstructionsEnum.NOTHING;
+
+  @CommandLine.Option(
+      names = "--access",
+      description = "Access scope for the resource: ${COMPLETION-CANDIDATES}")
+  private AccessScope access = AccessScope.SHARED_ACCESS;
```
wchamber, line 38: Consider using MixIns for each of these "standard" fields across different resources. That would let us define them once and reuse them for each resource.

It might even make sense to use a ResourceName mixin in not only create commands, but also delete & update, anywhere we use a "resource name." Been thinking about using more mixins for some of my shared notebook options, e.g. `--location`, next time I'm touching them.

src/main/java/bio/terra/cli/command/resources/resolve/BqDataset.java
```diff
@@ -0,0 +1,59 @@
...
+  @CommandLine.ArgGroup(exclusive = true, multiplicity = "0..1")
+  BqDataset.BqDatasetCloudIdArgGroup argGroup;
+
+  static class BqDatasetCloudIdArgGroup {
+    @CommandLine.Option(
+        names = "--dataset-id",
+        description = "Only return the Big Query dataset id.")
+    private boolean datasetId;
+
+    @CommandLine.Option(names = "--project-id", description = "Only return the GCP project id.")
```
wchamber, line 32: is there a way to describe what's returned if neither is set? Does an enum make more sense for this tri-state, or is there a reason to prefer using exclusive flags? Naively, an enum parameter with a default value seems more self documenting.

src/main/java/bio/terra/cli/context/WorkspaceContext.java
```diff
@@ -30,16 +30,16 @@
   // workspace description object returned by WSM
   public WorkspaceDescription terraWorkspaceModel;
 
-  // map of cloud resources for this workspace (name -> object)
-  public Map<String, CloudResource> cloudResources;
+  // map of cloud resources for this workspace (name -> WSM resource description object)
+  public Map<String, ResourceDescription> resources;
```
wchamber, line 34: Is this a breaking change? Are existing CLI workspace files going to be incompatible? I think it's OK for us to make breaking changes at this point, as long as we're coordinating with the solutions team and know exactly what the impact is.
Not this PR, but this starts to feel like we might want to think about data serialization formats for the CLI @jmczerk.

src/main/java/bio/terra/cli/service/WorkspaceManager.java
```diff
@@ -228,89 +245,224 @@ public RoleBindingList listUsersOfWorkspace() {
...
-   *
-   * @param resourceName name of resource to lookup
-   * @return the cloud resource object
-   * @throws UserActionableException if the resource is not controlled (e.g. external bucket)
+   * Update the cached list of resources (controlled & referenced) in the workspace, by fetching the
+   * up-to-date list from WSM.
    */
-  public CloudResource getControlledResource(String resourceName) {
+  private void updateResourcesCache() {
+    // check that there is a workspace currently mounted
```
wchamber, line 256: no action: I was thinking we might want to have a command for updating the cache, but I guess as long as listing and resolving resources does that we're ok. Maybe worth documenting for users somewhere so that if their collaborator changes something about a workspace they know they need to do something before calling docker run to get a new resource env var. I can imagine it being very confusing to have resources changed out from under you. Not sure what the best way to do that is.

src/main/java/bio/terra/cli/service/WorkspaceManager.java
```diff
@@ -228,89 +245,224 @@ public RoleBindingList listUsersOfWorkspace() {
...
-   * @return the cloud resource that was created
+   * @param createResource function pointer to execute the create request for a particular
+   *     stewardship + resource type combination
+   * @param <T> type of the create response object. this is usually very similar to the
+   *     ResourceDescription object, but not quite identical.
+   * @return the resource description object that was created
    */
-  public CloudResource createControlledResource(
-      CloudResource.Type resourceType, String resourceName) {
+  private <T> ResourceDescription createResource(String resourceName, Supplier<T> createResource) {
```
wchamber, line 449: Do we need to take a Supplier as an argument and control when it's called, or is it ok to just take a `T resource` and rename this. Or maybe this isn't doing enough at all? Not totally against, but not totally sold on this function.

src/main/java/bio/terra/cli/service/WorkspaceManager.java
```diff
@@ -228,89 +245,224 @@ public RoleBindingList listUsersOfWorkspace() {
...
     workspaceContext.requireCurrentWorkspace();
-    // TODO: change this method to call WSM controlled resource endpoints once they're ready
+
     if (!isValidEnvironmentVariableName(resourceName)) {
       throw new UserActionableException(
           "Resource name can contain only alphanumeric and underscore characters.");
```
wchamber, line 454: do we still need to do this validation or is it ok to rely on WM? I think we'll get out of date on what's allowed, and it will be possible to create resources from paths other than the CLI.

src/main/java/bio/terra/cli/service/WorkspaceManager.java
```diff
@@ -324,156 +476,181 @@ private static boolean isValidEnvironmentVariableName(String name) {
...
-   *     id)
-   * @return the data reference that was added
+   * @param resourceName name of resource to delete
+   * @param deleteResource function pointer to execute the delete request for a particular
+   *     stewardship + resource type combination
+   * @return the resource description object that was deleted
    */
-  public CloudResource addDataReference(
-      CloudResource.Type referenceType, String referenceName, String cloudId) {
+  private ResourceDescription deleteResource(String resourceName, Consumer<UUID> deleteResource) {
```
wchamber, line 550: ditto comment on createResource

src/main/java/bio/terra/cli/service/utils/WorkspaceManagerService.java
```diff
@@ -332,4 +365,297 @@ public RoleBindingList getRoles(UUID workspaceId) {
...
+    ControlledGcpResourceApi controlledGcpResourceApi = new ControlledGcpResourceApi(apiClient);
+    CreateControlledGcpBigQueryDatasetRequestBody createRequest =
+        new CreateControlledGcpBigQueryDatasetRequestBody()
+            .common(
+                new ControlledResourceCommonFields()
+                    .name(name)
+                    .description(description)
+                    .cloningInstructions(cloningInstructions)
+                    .accessScope(accessScope)
+                    .managedBy(ManagedBy.USER))
```
wchamber, line 573: Hmm, not this PR, but eventually I think we also need to specify the `PrivateResourceUser` fields for private resources I guess all created resources right now are workspace shared? I think it's a bad request exception when `AccessScope` is `PRIVATE_ACCESS` but no `PrivateResourceUser` roles (and user) is defined.

src/main/java/bio/terra/cli/service/utils/WorkspaceManagerService.java
```diff
@@ -332,4 +365,297 @@ public RoleBindingList getRoles(UUID workspaceId) {
...
+    ControlledGcpResourceApi controlledGcpResourceApi = new ControlledGcpResourceApi(apiClient);
+    String asyncJobId = UUID.randomUUID().toString();
+    DeleteControlledGcpGcsBucketRequest deleteRequest =
+        new DeleteControlledGcpGcsBucketRequest().jobControl(new JobControl().id(asyncJobId));
+    // TODO (PF-719): factor out this polling pattern into a utility method
+    try {
+      DeleteControlledGcpGcsBucketResult deleteResult =
+          controlledGcpResourceApi.deleteBucket(deleteRequest, workspaceId, resourceId);
+      while (deleteResult.getJobReport().getStatus().equals(JobReport.StatusEnum.RUNNING)) {
+        TimeUnit.SECONDS.sleep(5);
```
wchamber, line 641: I don't think as implemented this completes anytime soon for non-empty buckets. I think for the CLI we might want to be able to handle operations that take a long time. But fine not to address in this PR.
@zloery

wchamber (commented):
Cool stuff.

I think that this won't work for the current verilycli WM deployment, or existing workspaces (will lose existing CLI created resources?) How do you think we should roll this out? Should we do a feature branch in the CLI? I'm a little worried about Nicole grabbing the latest CLI and losing stuff.

src/main/java/bio/terra/cli/apps/DockerCommandRunner.java
```diff
@@ -155,15 +159,35 @@ public void runToolCommand(String command, Map<String, String> envVars) {
...
     Map<String, String> terraReferences = new HashMap<>();
     workspaceContext
-        .listCloudResources()
+        .listResources()
         .forEach(
-            cloudResource ->
+            resource ->
                 terraReferences.put(
-                    "TERRA_" + cloudResource.name.toUpperCase(), cloudResource.cloudId));
+                    "TERRA_" + resource.getMetadata().getName().toUpperCase(),
```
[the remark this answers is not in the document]
marikomedlock: Environment variable names are case sensitive, no?
```
$ envvar=123 EnvVar=234; echo $envvar; echo $EnvVar
123
234
```
Agreed we don't need this uppercase. Removed.

src/main/java/bio/terra/cli/command/resources/CheckAccess.java
```diff
@@ -0,0 +1,90 @@
...
+    // the user/pet SA credentials and resource definition, so calling them looks very similar from
+    // the CLI user's perspective.
+    // still, it may be better to break this command into sub-commands for each resource type. that
+    // would allow different options per resource, e.g. checking different resource-specific
+    // permissions.
+    // TODO (PF-717): revisit this command once the WSM endpoints are complete
+    boolean userHasAccess, proxyGroupHasAccess;
+    switch (resource.getMetadata().getResourceType()) {
+      case GCS_BUCKET:
+        userHasAccess = workspaceManager.checkAccessToReferencedGcsBucket(resource, false);
```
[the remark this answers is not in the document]
marikomedlock: Changed from boolean -> enum.

src/main/java/bio/terra/cli/command/resources/List.java
```diff
@@ -2,29 +2,66 @@
...
+          resources.stream()
+              .filter(
+                  resource ->
+                      resource
+                          .getMetadata()
+                          .getStewardshipType()
+                          .equals(
+                              argGroup.controlled
+                                  ? StewardshipType.CONTROLLED
+                                  : StewardshipType.REFERENCED))
```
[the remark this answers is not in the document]
marikomedlock: Moved conditional to helper method on the arg group.

src/main/java/bio/terra/cli/command/resources/Delete.java
marikomedlock: Agreed about a central place being helpful.
Created a new `PrintingUtils` class with two methods so far:
- `printResource` which is just this method relocated
- `printWorkspace` which I added to several workspace-related commands `status`, `workspace create`, `workspace update`, `workspace delete`, `workspace mount`.

src/main/java/bio/terra/cli/command/resources/Describe.java
marikomedlock: Agreed. Added to the laundry list of style-related improvements [here](https://broadworkbench.atlassian.net/browse/PF-574).

src/main/java/bio/terra/cli/command/resources/create/BqDataset.java
marikomedlock: Changed to dataset location.

src/main/java/bio/terra/cli/service/WorkspaceManager.java
marikomedlock: RE coupling to WSM objects: I think one reason to define CLI-specific object is to protect against incompatible changes to the WSM API (e.g. client library v22 has a different version of the object than v23). For the purposes of the alpha/demo target, that seems like a low priority item, so perhaps the benefit of not having to re-define an essentially identical POJO object is worth it for now? I'm less sure this is right approach longer-term.

Also noting that we're not completely coupled to WSM APIs, in the sense that we're not directly using the endpoint return objects in many cases. `WorkspaceManager` makes sure we always get a `ResourceDescription` object when creating a resource (create apis for example often return a very similar, but different object), so that and `WorkspaceDescription` are the only things (I think) that we pass around throughout the code base. For this reason, I think it would be straightforward to switch over to using CLI-specific objects (similar to `TerraUser`) in the future if we decide on that path.

RE long parameter lists: Populating the `CreateGcpGcsBucketReferenceRequestBody` in the `WorkspaceManager` class seems fine since it also handles the response objects, but I'd prefer not to have it in the command class.

src/main/java/bio/terra/cli/command/resources/create/GcsBucket.java
marikomedlock: Agreed that might be coming. For the ticket I'm working on now, `terra resources create gcs-bucket` will optionally take a JSON file that specifies the lifecycle rules, the same as the `gsutil lifecycle set` command [does](https://cloud.google.com/storage/docs/gsutil/commands/lifecycle). I'm planning to use the same JSON [format](https://cloud.google.com/storage/docs/json_api/v1/buckets#resource) as `gsutil`. So if we ever want to allow passing a JSON file for the whole resource definition (not just lifecycle rules), perhaps we can follow the same approach.

src/main/java/bio/terra/cli/context/WorkspaceContext.java
marikomedlock: Yep it's a breaking change in many ways: reorganized commands, changed the context serialization format, resources previously created will be unreachable. Of all the changes, the reorganized commands may be the largest burden, because Nicole's notebook scripts will be broken.

Certainly interested to hear any serialization advice you have!

src/main/java/bio/terra/cli/service/WorkspaceManager.java
marikomedlock: Agreed about adding this to the documentation. I'm planning to update the README in a follow-on ticket, after I get everyone's okay on this command reorganization, in particular.

src/main/java/bio/terra/cli/service/WorkspaceManager.java
marikomedlock: If we keep this function, then it needs to take a `Supplier` because we need to do the name validation check before creating the resource.

For both `createResource` and `deleteResource`, they just make sure all the create/delete methods update the cache correctly. I went back and forth over whether this is worth it or not. Happy to remove -- it just means we have the same couple lines before and same couple lines after in all create/delete functions.

src/main/java/bio/terra/cli/service/WorkspaceManager.java
marikomedlock: At the moment, we still need this validation because WSM is not enforcing it. The last time I brought this up, I think we decided to punt on making a decision.

Another approach would be to stop passing environment variables into the container, install the Terra CLI in the container, and then force apps to call `terra resources resolve ...` instead of relying on environment variables. This would also remove the need to escape things.

Regardless, I think this should be addressed in a separate PR, after some discussion and possibly prototyping/demoing with the solutions team code.

src/main/java/bio/terra/cli/service/utils/WorkspaceManagerService.java
marikomedlock: Certainly it should time out at some point. I'm planning to handle this in a follow-on ticket (see TODO comment above). If this will often take a long time, then I'll make the timeout short and perhaps show a "check back later" type message instead of "failure" if it does time out.

src/main/java/bio/terra/cli/context/WorkspaceContext.java
jmczerk: Probably a good discussion to start having out side the PR...  It seems like a very bounded use case which should be very amenable to evolving schemas without a lot of overhead.  I assume human readability/edit-ability is desirable?

src/main/java/bio/terra/cli/command/resources/Describe.java
```diff
@@ -2,35 +2,73 @@
...
+        OUT.println(
+            "Private user: "
+                + returnValue
+                    .getMetadata()
+                    .getControlledResourceMetadata()
+                    .getPrivateResourceUser()
+                    .getUserName());
+      }
+    }
+
```
jmczerk, line 56: Seeing five similar `switch(ResourceType)` statements raised some concerns, but I thought about it a bit and can’t think of an alternative that saves much boilerplate.  Might be worth revisiting if/when the number of possible resource types grows above 3 or 4…

jmczerk (approved):
Yay deletions!

src/main/java/bio/terra/cli/command/resources/resolve/BqDataset.java
marikomedlock: Changed to use an enum.

src/main/java/bio/terra/cli/context/WorkspaceContext.java
marikomedlock: Yes, I think human readable is always preferable, unless there's a good reason against. I don't expect the context we persist on disk to grow extremely large -- we're reading it all into memory now, so if we do ever get very large workspaces in terms of the number of resources, that would require much larger code changes anyway.

src/main/java/bio/terra/cli/service/WorkspaceManager.java
marikomedlock: Yeah, I'm on the fence about whether these are worth it, too. It was just a code de-duplication effort, basically every `delete...` method needs these lines before:
```
    // get the summary object
    ResourceDescription resourceToDelete = workspaceContext.getResource(resourceName);
```
and these lines after:
```
    updateResourcesCache(); // persist the cloud resource locally
    return resourceToDelete; // return just the summary object
```
Similar for `create`. It's not a lot of code, I was just thinking that the next time we add a new resource, I won't have to remember to copy that. Thoughts?

src/main/java/bio/terra/cli/command/resources/resolve/BqDataset.java
marikomedlock: No, It doesn't appear by default. I'll try making it an enum instead and see what you think.

marikomedlock: Changed from exclusive flags -> enum.
Screen Shot 2021-04-29 at 4 48 23 PM

src/main/java/bio/terra/cli/command/resources/Describe.java
marikomedlock: Perhaps this is a second good reason to use CLI-defined objects? Interfaces or sub-classes seem like they'd get rid of most of these switch statements. (e.g. sub-classes override delete, checkAccess, printText methods).

src/main/java/bio/terra/cli/service/utils/WorkspaceManagerService.java
marikomedlock: Good catch, I overlooked those fields. Added.

src/main/java/bio/terra/cli/command/resources/create/GcsBucket.java
marikomedlock: Great suggestion! Added three new mixin classes: `ResourceName`, `CreateResource`, `CreateControlledResource` that contain varying combinations of options used in multiple commands.

zloery (approved):
LGTM on the workspace manager side!

src/main/java/bio/terra/cli/command/resources/Delete.java
```diff
@@ -1,27 +1,72 @@
...
+          throw new UnsupportedOperationException(
+              "Resource type not supported: " + resourceToDelete.getMetadata().getResourceType());
+      }
+    } else if (resourceToDelete
+        .getMetadata()
+        .getStewardshipType()
+        .equals(StewardshipType.CONTROLLED)) {
+      switch (resourceToDelete.getMetadata().getResourceType()) {
+        case GCS_BUCKET:
+          workspaceManager.deleteControlledGcsBucket(name);
```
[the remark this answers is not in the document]
wchamber: A feature request for WM would be to delete object by object when the bucket isn't too big. The lifecycle to 0 is a good trick and very helpful when the bucket is huge, but taking a day to delete a bucket with like 3 files seems overkill. But probably not super high on the priority list.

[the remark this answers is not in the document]
marikomedlock: No, there's no existing special handling in the CLI for long-running commands. I don't know that we want the CLI to keep checking WSM in the background, maybe just time out with a warning to the user that this sometimes takes awhile.

marikomedlock:
> I think that this won't work for the current verilycli WM deployment, or existing workspaces (will lose existing CLI created resources?) How do you think we should roll this out? Should we do a feature branch in the CLI? I'm a little worried about Nicole grabbing the latest CLI and losing stuff.

Yes, this will be a breaking change for existing workspaces with resources in them, and for the scripts that are calling the now re-organized resources commands. I can think of two approaches: maintain a feature branch in the CLI, merge the code but only publish pre-releases so that their `curl` command that fetches the latest release will continue to fetch the older one. It might be helpful to discuss this briefly over GVC instead of here. I will write a short summary of options and send it out. Regardless, I will not be merging this PR until we decide.

src/main/java/bio/terra/cli/service/utils/WorkspaceManagerService.java
```diff
@@ -393,20 +396,18 @@ public ResourceList enumerateResources(
...
    * @return the GCS bucket resource object
    */
   public GcpGcsBucketResource createReferencedGcsBucket(
-      UUID workspaceId,
-      String name,
-      String description,
-      CloningInstructionsEnum cloningInstructions,
-      String gcsBucketName) {
-    ReferencedGcpResourceApi referencedGcpResourceApi = new ReferencedGcpResourceApi(apiClient);
+      UUID workspaceId, ResourceDescription resourceToAdd) {
```
wchamber, line 403: re: CLI specific objects, using the WSM "result" object model as a creation parameter definitely implies to me that we do want to have CLI specific objects. Follow up PF-729 ticket sounds good, and I'd argue it's maybe medium term instead of long term, but we can see how often we use it and if WM API breaks us.

wchamber (approved):

src/main/java/bio/terra/cli/service/WorkspaceManager.java
marikomedlock: I prefer having business logic validation in the main CLI code, instead of in the command classes. The only validation I think makes sense in the command classes is argument missing/bad argument combination type validation. Maybe a moot point here though, since this logic will likely be removed (either completely or pushed into WSM).

src/main/java/bio/terra/cli/command/resources/Describe.java
marikomedlock: Noted this potential for code cleanup in [PF-729](https://broadworkbench.atlassian.net/browse/PF-729).

src/main/java/bio/terra/cli/command/helperclasses/PrintingUtils.java
```diff
@@ -0,0 +1,64 @@
...
+import static bio.terra.cli.command.helperclasses.BaseCommand.OUT;
+
+import bio.terra.cli.context.WorkspaceContext;
+import bio.terra.workspace.model.ResourceDescription;
+
+/** Utility methods for printing command output. */
+public class PrintingUtils {
+
+  /** Print out a workspace object in text format. */
+  public static void printWorkspace(WorkspaceContext workspaceContext) {
```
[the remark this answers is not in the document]
marikomedlock: Changed both to `printText`.

src/main/java/bio/terra/cli/command/resources/List.java
```diff
@@ -1,49 +1,46 @@
...
-              .filter(
-                  resource ->
-                      resource
-                          .getMetadata()
-                          .getStewardshipType()
-                          .equals(
-                              argGroup.controlled
-                                  ? StewardshipType.CONTROLLED
-                                  : StewardshipType.REFERENCED))
+              .filter(resource -> resource.getMetadata().getResourceType().equals(type))
```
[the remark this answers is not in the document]
marikomedlock: Changed it to use a single `filter` that checks both conditions.

src/main/java/bio/terra/cli/command/resources/CheckAccess.java
```diff
@@ -12,21 +13,17 @@
...
 public class CheckAccess extends BaseCommand {
-  @CommandLine.Option(
-      names = "--name",
-      required = true,
-      description = "Name of the resource, scoped to the workspace.")
-  private String name;
+  @CommandLine.Mixin ResourceName resourceNameMixin;
 
-  @CommandLine.Mixin FormatOption formatOption;
+  @CommandLine.Mixin Format formatOption;
```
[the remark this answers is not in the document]
marikomedlock: I prefer the convention of `ResourceName resourceNameMixin` because I think it makes accessing the value slightly more readable (e.g. `cloningMixin.cloning` instead of `cloning.cloning`).

src/main/java/bio/terra/cli/command/helperclasses/options/CreateResource.java
```diff
@@ -0,0 +1,21 @@
...
+import bio.terra.workspace.model.CloningInstructionsEnum;
+import picocli.CommandLine;
+
+/**
+ * Command helper class that defines the relevant options for create a new controlled or referenced
+ * Terra resource: {@link ResourceName} and --description, --cloning.
+ *
+ * <p>This class is meant to be used as a @CommandLine.Mixin.
+ */
+public class CreateResource extends ResourceName {
```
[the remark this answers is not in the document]
marikomedlock: I am very not sure about this being the right structure for the `command.helperclasses.options` classes. But I would like a few more examples of option sharing before making a different decision. I think this will next come up with the `terra resources update` commands, which will probably want to reuse some of the `terra resources create` options. So, expecting some reorganization in this area in that PR, but don't want to do it prematurely.

RE mixin classes using mixins: I did this originally, but changed it to use inheritance because of the chaining required to access values (e.g. `createControlledResourceOptions.createResourceOptions.name` instead of `createControlledResourceOptions.name`). If we start having weird inheritance patterns, then certainly chaining accessors is better, but I figured we're not there yet.

RE only using one option per mixin: This would make each accessor more reusable across different types of commands. I do think the common groupings (e.g. options that all `terra resources create ...` commands need) are helpful and want to preserve that somehow.

PR #66, opened 2021-05-04
PF-707: Call WSM endpoint to delete controlled BQ datasets.

marikomedlock:
- Replaced a "not implemented yet" exception with a call to the WSM endpoint to delete a controlled BQ dataset resource.
- Bumped the WSM client library version to include the new endpoint.


wchamber (approved):

zloery (approved):
LGTM!

PR #67, opened 2021-05-04
PF-727: Update readme with resources changes

marikomedlock:
- Updated the README with the reorganized `terra resources` group of commands.
- Also did a pass through both README and CONTRIBUTING updating wording and fixing paragraph structure.


wchamber (approved):
Thanks! Not always easy to know when the documentation is helpful, but I know I've appreciated having it.

marikomedlock:
> Thanks! Not always easy to know when the documentation is helpful, but I know I've appreciated having it.

I hear you on not knowing when to look at which documentation page. I've been thinking perhaps a reorganization of the documentation might be helpful. (README for a description of the GitHub repo and what the code does, USER-GUIDE with examples/instructions for CLI users, CONTRIBUTING for developers). Thoughts?

wchamber:
That sounds good to me, though maybe keep the user guide in the front readme to make it even easier for users to find info.
(to be clear, I'm happy with what you've been adding so far :) )

PR #68, opened 2021-05-05
PF-724: Add top-level resolve command.

marikomedlock:
Added a top-level `terra resolve` command. This is just a shortcut for what we expect to be a fairly heavily used command.

It is functionally equivalent to calling the `terra resources resolve ...` command for the appropriate resource type, with all default options. i.e.:
`terra resolve bucketctrl1` = `terra resources resolve gcs-bucket --name=bucketctrl1`
`terra resolve bqref1` = `terra resources resolve bq-dataset --name=bqref1`
where `bucketctrl1`=name of a controlled GCS bucket resource and `bqref1`=name of a referenced BQ dataset resource.

The `terra resources resolve ...` commands are still useful because:
- They provide options around what format of resolved cloud identifier is returned (e.g. `--exclude-prefix` for GCS buckets, `--cloud-id-format=DATASET_ID_ONLY` for BQ datasets)
- They are grouped near the other `resources` related commands, so potentially easier to find.

### DIFF
--- a/src/main/java/bio/terra/cli/command/Main.java
+++ b/src/main/java/bio/terra/cli/command/Main.java
@@ -32,6 +32,7 @@ import picocli.CommandLine.ParseResult;
       Groups.class,
       Spend.class,
       Config.class,
+      Resolve.class,
       Gcloud.class,
       Gsutil.class,
       Bq.class,
--- a/src/main/java/bio/terra/cli/command/Resolve.java
+++ b/src/main/java/bio/terra/cli/command/Resolve.java
@@ -1,0 +1,45 @@
+package bio.terra.cli.command;
+
+import static bio.terra.cli.service.WorkspaceManager.getBigQueryDatasetPath;
+import static bio.terra.cli.service.WorkspaceManager.getGcsBucketUrl;
+
+import bio.terra.cli.command.helperclasses.BaseCommand;
+import bio.terra.cli.command.helperclasses.options.Format;
+import bio.terra.cli.service.WorkspaceManager;
+import bio.terra.workspace.model.ResourceDescription;
+import picocli.CommandLine;
+import picocli.CommandLine.Command;
+
+/** This class corresponds to the second-level "terra resolve" command. */
+@Command(name = "resolve", description = "Resolve a resource to its cloud id or path.")
+public class Resolve extends BaseCommand {
+  @CommandLine.Parameters(
+      index = "0",
+      description = "Name of the resource, scoped to the workspace.")
+  private String resourceName;
+
+  @CommandLine.Mixin Format formatOption;
+
+  /** Resolve a resource in the workspace to its cloud identifier. */
+  @Override
+  protected void execute() {
+    ResourceDescription resource =
+        new WorkspaceManager(globalContext, workspaceContext).getResource(resourceName);
+
+    // return the "default" cloud identifier (i.e. the format returned by the `terra resource
+    // resolve ...` commands if no options are specified)
+    String cloudId;
+    switch (resource.getMetadata().getResourceType()) {
+      case GCS_BUCKET:
+        cloudId = getGcsBucketUrl(resource);
+        break;
+      case BIG_QUERY_DATASET:
+        cloudId = getBigQueryDatasetPath(resource);
+        break;
+      default:
+        throw new UnsupportedOperationException(
+            "Resource type not supported: " + resource.getMetadata().getResourceType());
+    }
+    formatOption.printReturnValue(cloudId);
+  }
+}


wchamber:
I guess I'm not totally convinced that we shouldn't collapse the resolve commands somehow into 1 command.
1. For the resource specific options, could we satisfy that with [ArgGroups](https://picocli.info/#_argument_groups) for each resource type? The usage output gets verbose in that case, but I don't know if that's much of a usability concern. We could have the bucket specific arg group have a dummy required `--gcs-bucket` option to use the arg groups. Each resource type arg group could be exclusive with each other.
2. Having `terra resources resolve` seems like the right place to talk about resolving resources. I don't know that it's always that relevant a use to bring it out of `resources`. If the issue with the current per-resource type resolves is that they're a little verbose or they require that you know the type of resource you have, `terra resources resolve my-resource-name` seems ok. 

OTOH, there's probably not much cost to having multiple commands.

marikomedlock:
@wchamber 
1. Yes, we could use ArgGroups instead, but that seems like nearly identical syntax for slightly more complex Java code. It seemed cleaner to me both UX and code-wise to just have separate command classes. Open to other opinions though -- am I missing some other benefits to combining everything in a single command class?
e.g. `terra resources resolve --gcs-bucket [...options...]` with multiple ArgGroups and case statements to deal with them
vs `terra resources resolve gcs-bucket [...options...]` with multiple command classes.

2. I agree that it's useful to have `resources`-related commands in the `resources` grouping. I believe the main motivation for having a top-level shortcut/alias command is just to have a less verbose way of doing what we expect to be a fairly common thing. The original request came from @gjuggler though, so tagging him here in case I missed another motivation.

wchamber:
1. I don't think it's huge benefits. But not duplicating any kind of resolve code by having a single command class. I don't think the switch statement is bad. Or even that we'd need `--gcs-bucket` to enable resource specific options. But yeah, mainly not duplicating resolve commands: having a single command that does everything you need as opposed to a shorter command that needs to be different some of the time.
2. `terra resources resolve my-resource-name` seems ok to me instead of `terra resolve my-resource-name`. But maybe it's hypocritical for me to say I like the first more than `terra resources resolve gcs-bucket my-resource-name` :)

wchamber (approved):

jmczerk (approved):
I can see some potential for tweaking the UX per the discussion, but I don’t feel strongly enough to advocate for/against doing so.  If the intent of the ticket is to have a short alias for default output where the resource type is inferred, this accomplishes that well.

marikomedlock:
- Thought about this some more, and I think I agree with merging the `resolve gcs-bucket` and `resolve bq-dataset` commands into a single `resolve` command.
  - Avoiding code duplication is always good. The duplication in this case was relatively small, but still it's a benefit.
  - Users shouldn't have to specify the resource type when they're resolving a resource. The CLI already knows the type, so no need to have the user tell us. This is similar to the way other similar commands work (e.g. `terra resources delete`, `terra resources describe`, `terra resources check-access`) -- that is, they don't have resource-type specific sub-commands.

- Made a single class for the `terra resources resolve` implementation and aliased it to `terra resolve`, per the original request. If you specify a bucket-specific option for a bq dataset resource, it will just be ignored.

I'm going to re-request reviews, since this seems like it's changed enough to warrant a second look. Thanks in advance!

src/main/java/bio/terra/cli/command/resources/Resolve.java
```diff
@@ -1,15 +1,83 @@
...
+  @CommandLine.Mixin Format formatOption;
+
+  /** Resolve a resource in the workspace to its cloud identifier. */
+  @Override
+  protected void execute() {
+    ResourceDescription resource =
+        new WorkspaceManager(globalContext, workspaceContext).getResource(resourceNameOption.name);
+
+    // return the "default" cloud identifier (i.e. the format returned by the `terra resource
+    // resolve ...` commands if no options are specified)
```
wchamber, line 40: nit: comment no longer relevant?

src/main/java/bio/terra/cli/command/resources/Resolve.java
```diff
@@ -1,15 +1,83 @@
...
+
+/** This class corresponds to the third-level "terra resources resolve" command. */
+@Command(name = "resolve", description = "Resolve a resource to its cloud id or path.")
+public class Resolve extends BaseCommand {
+  @CommandLine.Mixin ResourceName resourceNameOption;
+
+  @CommandLine.Option(
+      names = "--exclude-bucket-prefix",
+      description = "[For GCS_BUCKET] Exclude the 'gs://' prefix.")
+  private boolean excludePrefix;
```
wchamber, line 22: nit: `excludeBucketPrefix`?

wchamber (approved):

jmczerk (approved):
Looks good!

PR #69, opened 2021-05-07
PF-719: Utility method for polling with retries.

marikomedlock:
- Added `pollWithRetries` utility method that polls or retries a specified number of times, with a sleep in between each.
  - Poll = check the result of an operation to see if it's done (e.g. polling to see if a job is done)
  - Retry = try an operation again if a specific kind of exception is thrown (e.g. retry after 500 errors)
  - This is basically the same as the existing `callWithRetries` utility method, except that it takes an additional lambda parameter `isDone` and tests the result each time instead of just returning the first result that doesn't throw an exception.
- Called the new method when:
  - Deleting a controlled GCS bucket resource
  - Creating a new workspace
- Increased the default number of retries from 10 -> 30.

- Also fixed an unrelated bug, where the resources cache was not being updated when mounting an existing workspace.

### DIFF
--- a/src/main/java/bio/terra/cli/service/utils/HttpUtils.java
+++ b/src/main/java/bio/terra/cli/service/utils/HttpUtils.java
@@ -22,10 +22,10 @@ public class HttpUtils {
   private static final Logger logger = LoggerFactory.getLogger(HttpUtils.class);
 
   // default value for the maximum number of times to retry HTTP requests
-  private static final int DEFAULT_MAXIMUM_RETRIES = 10;
+  public static final int DEFAULT_MAXIMUM_RETRIES = 10;
 
   // default value for the time to sleep between retries
-  private static final Duration DEFAULT_DURATION_SLEEP_FOR_RETRY = Duration.ofSeconds(1);
+  public static final Duration DEFAULT_DURATION_SLEEP_FOR_RETRY = Duration.ofSeconds(1);
 
   private HttpUtils() {}
 
@@ -143,7 +143,7 @@ public class HttpUtils {
   /**
    * Helper method to call a function with retries.
    *
-   * @param maxRetries maximum number of times to retry
+   * @param maxCalls maximum number of times to retry
    * @param sleepDuration time to sleep between tries
    * @param makeRequest function to perform the request
    * @param isRetryable function to test whether the exception is retryable or not
@@ -153,13 +153,13 @@ public class HttpUtils {
    *     number of retries was exhausted
    */
   public static <T, E extends Exception> T callWithRetries(
-      int maxRetries,
+      int maxCalls,
       Duration sleepDuration,
       SupplierWithCheckedException<T, E> makeRequest,
       Predicate<Exception> isRetryable)
       throws E, InterruptedException {
     // isDone always return true
-    return pollWithRetries(maxRetries, sleepDuration, makeRequest, (result) -> true, isRetryable);
+    return pollWithRetries(maxCalls, sleepDuration, makeRequest, (result) -> true, isRetryable);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/main/java/bio/terra/cli/service/utils/HttpUtils.java
```diff
@@ -122,52 +122,120 @@ public static HttpResponse sendHttpRequest(
...
-      int maxRetries,
+      int maxCalls,
       Duration sleepDuration,
       SupplierWithCheckedException<T, E> makeRequest,
       Predicate<Exception> isRetryable)
       throws E, InterruptedException {
+    // isDone always return true
+    return pollWithRetries(maxCalls, sleepDuration, makeRequest, (result) -> true, isRetryable);
+  }
+
```
jmczerk, line 166: This is just a nit, and feel free to ignore if this is my C++ bias showing, but I was surprised that the two parameters that get "defaulted" by overload (`DEFAULT_MAXIMUM_RETRIES` and `DEFAULT_DURATION_SLEEP_FOR_RETRY`) were at the beginning of the signature and not the end.

src/main/java/bio/terra/cli/service/utils/WorkspaceManagerService.java
```diff
@@ -705,4 +709,41 @@ public void deleteControlledBigQueryDataset(UUID workspaceId, UUID resourceId) {
...
+  /**
+   * Helper method that checks a JobReport's status and throws an exception if it's not COMPLETED.
+   *
+   * <p>- Throws a {@link SystemException} if the job FAILED.
+   *
+   * <p>- Throws a {@link UserActionableException} if the job is still RUNNING. Some actions are
+   * expected to take a long time (e.g. deleting a bucket with lots of objects), and a timeout is
+   * not necessarily a failure. The action the user can take is to wait a bit longer and then check
+   * back (e.g. by listing the buckets in the workspace) later to see if the job completed.
+   *
```
jmczerk, line 722: Not for this PR, but do we have any future planned stories around asynchronous commands that allow for completion polling directly through CLI (as opposed to observing side effects)?

jmczerk (approved):

src/main/java/bio/terra/cli/service/utils/HttpUtils.java
```diff
@@ -122,52 +122,120 @@ public static HttpResponse sendHttpRequest(
...
       numTries++;
       try {
-        logger.info("Http request attempt #{}", numTries);
-        return makeRequest.makeRequest();
+        logger.debug("Request attempt #{}", numTries);
+        T result = makeRequest.makeRequest();
+        logger.debug("Result: {}", result);
+
+        if (isDone.test(result) || numTries > maxCalls) {
+          // polling is either done (i.e. job completed) or timed out: return the last result
```
wchamber, line 233: nit: log debug which it is?

wchamber (approved):

src/main/java/bio/terra/cli/service/utils/HttpUtils.java
marikomedlock: Good point. Reordered arguments to put the "defaulted" ones last.

src/main/java/bio/terra/cli/service/utils/HttpUtils.java
marikomedlock: Added.

src/main/java/bio/terra/cli/service/utils/WorkspaceManagerService.java
marikomedlock: We've talked about it briefly a couple times so far, and always decided to punt. I just added a [ticket](https://broadworkbench.atlassian.net/browse/PF-745) to make sure we don't forget about it.

PR #70, opened 2021-05-07
PF-706 Enumerate all resources up to maximum cache size.

marikomedlock:
- Defined a maximum cache size for the number of resources in a workspace = 1000. This corresponds to ~1MB on disk.
  - Currently the CLI passes a map of resource name -> cloud identifier to tools via environment variables. The CLI caches the resource definitions so that when users run tools (e.g. `terra nextflow ...`, `terra gsutil ...`), we don't have to make possibly multiple calls to WSM to fetch the resources each time, which would make the tools very slow to run.
  - This is not a hard limit, but if we reach this many resources in a workspace, then I think we should reconsider how/should the CLI be caching them on disk.
  - Added a config property for this cache size.
Screen Shot 2021-05-07 at 12 19 10 PM
- Called the WSM `enumerateResources` possibly multiple times until we either fetch all resources in the workspace or hit the maximum cache size.

### DIFF
--- a/src/main/java/bio/terra/cli/command/config/List.java
+++ b/src/main/java/bio/terra/cli/command/config/List.java
@@ -26,6 +26,7 @@ public class List extends BaseCommand {
         new ConfigListReturnValue(
             globalContext.browserLaunchOption,
             globalContext.dockerImageId,
+            globalContext.resourcesCacheSize,
             loggingLevels,
             globalContext.server);
 
@@ -36,16 +37,19 @@ public class List extends BaseCommand {
   private static class ConfigListReturnValue {
     public AuthenticationManager.BrowserLaunchOption browser;
     public String image;
+    public int resources;
     public LoggingReturnValue logging;
     public ServerSpecification server;
 
     public ConfigListReturnValue(
         AuthenticationManager.BrowserLaunchOption browser,
         String image,
+        int resources,
         LoggingReturnValue logging,
         ServerSpecification server) {
       this.browser = browser;
       this.image = image;
+      this.resources = resources;
       this.logging = logging;
       this.server = server;
     }
@@ -55,6 +59,8 @@ public class List extends BaseCommand {
   private static void printText(ConfigListReturnValue returnValue) {
     OUT.println("[browser] browser launch for login = " + returnValue.browser);
     OUT.println("[image] docker image id = " + returnValue.image);
+    OUT.println(
+        "[resources] max number of resources to allow per workspace = " + returnValue.resources);
     OUT.println();
     Logging.printText(returnValue.logging);
     OUT.println();
--- a/src/main/java/bio/terra/cli/command/config/List.java
+++ b/src/main/java/bio/terra/cli/command/config/List.java
@@ -60,7 +60,8 @@ public class List extends BaseCommand {
     OUT.println("[browser] browser launch for login = " + returnValue.browser);
     OUT.println("[image] docker image id = " + returnValue.image);
     OUT.println(
-        "[resources] max number of resources to allow per workspace = " + returnValue.resources);
+        "[resource-limit] max number of resources to allow per workspace = "
+            + returnValue.resources);
     OUT.println();
     Logging.printText(returnValue.logging);
     OUT.println();
--- a/src/main/java/bio/terra/cli/command/config/set/Resources.java
+++ b/src/main/java/bio/terra/cli/command/config/set/Resources.java
@@ -20,7 +20,10 @@ public class Resources extends BaseCommand {
         description = "maximum number to allow before throwing an error")
     private int max;
 
-    @CommandLine.Option(names = "--default", description = "use the default number of resources")
+    @CommandLine.Option(
+        names = "--default",
+        description =
+            "use the default number of resources: " + GlobalContext.DEFAULT_RESOURCES_CACHE_SIZE)
     private boolean useDefault;
   }
 
--- a/src/main/java/bio/terra/cli/command/config/set/Resources.java
+++ b/src/main/java/bio/terra/cli/command/config/set/Resources.java
@@ -1,0 +1,55 @@
+package bio.terra.cli.command.config.set;
+
+import bio.terra.cli.command.helperclasses.BaseCommand;
+import bio.terra.cli.context.GlobalContext;
+import picocli.CommandLine;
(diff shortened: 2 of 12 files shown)


src/main/java/bio/terra/cli/context/WorkspaceContext.java
```diff
@@ -33,6 +33,10 @@
   // map of cloud resources for this workspace (name -> WSM resource description object)
   public Map<String, ResourceDescription> resources;
```
jmczerk, line 35: Would it be a lot of trouble to add a `terra config` property to override this, so we have a workaround without a recompile if somebody goes over the edge?

marikomedlock: Great idea. I'll add a new config property for it.

marikomedlock: Added `terra config set resources` and `terra config get-value resources`. Screenshot included in PR description.

jmczerk (approved):

src/main/java/bio/terra/cli/command/config/List.java
```diff
@@ -55,6 +59,8 @@ public ConfigListReturnValue(
   private static void printText(ConfigListReturnValue returnValue) {
     OUT.println("[browser] browser launch for login = " + returnValue.browser);
     OUT.println("[image] docker image id = " + returnValue.image);
+    OUT.println(
+        "[resources] max number of resources to allow per workspace = " + returnValue.resources);
```
jaycarlton, line 63: "allow" makes it sound like we're enforcing how many are actually there. It seems more like "max resources to display per workspace."

src/main/java/bio/terra/cli/command/config/List.java
```diff
@@ -55,6 +59,8 @@ public ConfigListReturnValue(
   private static void printText(ConfigListReturnValue returnValue) {
     OUT.println("[browser] browser launch for login = " + returnValue.browser);
     OUT.println("[image] docker image id = " + returnValue.image);
+    OUT.println(
```
jaycarlton, line 62: maybe `resource-cache-limit` would be more precise, and you might want to configure something else about resources at some point. It's more typing though.

src/main/java/bio/terra/cli/command/config/set/Resources.java
```diff
@@ -0,0 +1,58 @@
...
+    description = "Set the maximum number of resources allowed per workspace.")
+public class Resources extends BaseCommand {
+
+  @CommandLine.ArgGroup(exclusive = true, multiplicity = "1")
+  ImageIdArgGroup argGroup;
+
+  static class ImageIdArgGroup {
+    @CommandLine.Option(
+        names = "--max",
+        description = "maximum number to allow before throwing an error")
```
jaycarlton, line 20: "throwing an error" sounds more programmer-y than user-y. Can a user still work with the first N resources happily, or is the order indeterminate?

src/main/java/bio/terra/cli/command/config/set/Resources.java
```diff
@@ -0,0 +1,58 @@
...
+@Command(
+    name = "resources",
+    description = "Set the maximum number of resources allowed per workspace.")
+public class Resources extends BaseCommand {
+
+  @CommandLine.ArgGroup(exclusive = true, multiplicity = "1")
+  ImageIdArgGroup argGroup;
+
+  static class ImageIdArgGroup {
+    @CommandLine.Option(
```
jaycarlton, line 18: Can you restrict this to be greater than zero? Otherwise you need a good way to describe/display zero resources.

src/main/java/bio/terra/cli/service/utils/WorkspaceManagerService.java
```diff
@@ -368,21 +367,51 @@ public RoleBindingList getRoles(UUID workspaceId) {
...
+                  + limit
+                  + ")");
+        }
+
+        // if this fetch returned less than the maximum allowed per request, then that indicates
+        // there are no more
+        if (result.getResources().size() < MAX_RESOURCES_PER_ENUMERATE_REQUEST) {
+          break;
+        }
+      } while (true);
```
jaycarlton, line 411: You don't need a do-while if the condition is always `true`; it degenerates to a while loop I think. I can see do while (resultsReturned < maxPerReq) if that's what you originally meant.

jaycarlton (approved):

src/main/java/bio/terra/cli/command/config/List.java
marikomedlock: This is actually enforcing how many resources are allowed for workspaces where you want to use the CLI. If there are more than that many resources, the CLI will throw an error anytime you try to do anything resource-related (e.g. `list`, `create`, ...).

src/main/java/bio/terra/cli/command/config/List.java
marikomedlock: Changed to `resource-limit`.

src/main/java/bio/terra/cli/command/config/set/Resources.java
marikomedlock: No, the user can't work with a subset of the resources, at least not with the code currently. The reason is that we pass a map of resource -> resource cloud identifier (e.g. `mybucket` -> `gs://my-bucket`) to all tools via environment variables. If we have so many resources (i.e. > max cache size) that they no longer fit comfortably on disk, then that approach seems less viable and we need to consider an overhaul in that area.

src/main/java/bio/terra/cli/command/config/set/Resources.java
marikomedlock: Good catch. Added validation that `maxResources > 0`.

src/main/java/bio/terra/cli/service/utils/WorkspaceManagerService.java
marikomedlock: Changed from `do...while(true)` -> `while(true)`.

marikomedlock: Yes, fixed the `do...while` loop by moving the `break` condition into the `while`. Thanks!

PR #73, opened 2021-05-11
PF-747 Change notebook specific commands for resources

wchamber:
Remove redundant list & describe commands.
Change start and stop to use Google APIs instead of the gcloud commandline.
Add a `notebooks get` for getting information about an instance.

Add CRL dependency for working with notebooks. While using CRL means we need to add more operations to it to use the underlying cloud APIs, it does provide some convenience for calling code and logging support.


build.gradle
```diff
@@ -10,6 +10,8 @@ plugins {
     id 'application'
     id 'com.diffplug.gradle.spotless' version '3.27.2'
     id 'com.github.spotbugs' version '4.2.0'
+    // Fix guava -jre vs -android dependency resolution.
```
marikomedlock, line 13: Is this just for CRL? If so, could you note that in the comment.

src/main/java/bio/terra/cli/command/helperclasses/options/NotebookName.java
```diff
@@ -0,0 +1,65 @@
...
+import bio.terra.cli.context.WorkspaceContext;
+import bio.terra.cli.service.WorkspaceManager;
+import bio.terra.cloudres.google.notebooks.InstanceName;
+import bio.terra.workspace.model.GcpAiNotebookInstanceAttributes;
+import bio.terra.workspace.model.ResourceDescription;
+import picocli.CommandLine;
+
+/**
+ * Command helper class for identifying a notebook by either the resource name of the GCP instance
+ * name in `terra notebooks` commands.
```
marikomedlock, line 14: typo-- or instance id?

src/main/java/bio/terra/cli/command/helperclasses/options/NotebookName.java
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * <p>This class is meant to be used as a @CommandLine.Mixin.
+ */
+public class NotebookName {
```
marikomedlock, line 18: nit: Maybe rename this so it's clearer that this isn't just a name option -- `NotebookNameOrId`? `NotebookInstance`? Or something else you think is better -- it just seems misleading because `ResourceName` is a single option so this one sounds like it is too.

src/main/java/bio/terra/cli/command/notebooks/Get.java
```diff
@@ -0,0 +1,40 @@
...
+import bio.terra.cli.command.helperclasses.options.NotebookName;
+import bio.terra.cli.service.utils.GoogleAiNotebooks;
+import bio.terra.cloudres.google.notebooks.InstanceName;
+import com.google.api.services.notebooks.v1.model.Instance;
+import picocli.CommandLine;
+
+/** This class corresponds to the third-level "terra notebooks get-url" command. */
+@CommandLine.Command(
+    name = "get",
+    description = "Get the GCP data of an AI Notebook instance within your workspace.",
```
marikomedlock, line 14: For notebooks, it's not clear to me when I should use `resources describe` or `notebooks get`. I think the difference is that `resources describe` returns information that WSM knows about, and `notebooks get` returns information that the cloud knows about. If so, that seems confusing to users who aren't familiar with what metadata WSM does/not store. What do you think?

If we were already using CLI-specific objects (i.e. PF-729 were done), then it would make sense to me for this lookup code to live on the CLI Notebook Resource object. Since we are not doing that yet, (brainstorming) perhaps we can put this code in a utility method and call it from the `printText` function in the notebooks case block? With a todo to move it to the object once that ticket is ready.

build.gradle
wchamber: Left a better note. Warning, it's a rabbit hole.

src/main/java/bio/terra/cli/command/helperclasses/options/NotebookName.java
wchamber: Yeah, naming's hard. Went with `NotebookInstance`

src/main/java/bio/terra/cli/command/notebooks/Get.java
wchamber: Agreed it's confusing. There's a split in our API/CLI about workspace operations around resources and resource-type specific operations. I don't know what the right way to resolve it is.

Calling GCP APIs from within a `printText` function seems kind of gross. (It also means we need more agruments for the workspace/global context to make the GCP call to get the notebook info.)
Ok if I just leave a TODO on `printText(ResourceDescription)` for now? If we get more feedback from the solutoins team, we can do more here.

src/main/java/bio/terra/cli/command/helperclasses/options/NotebookName.java
wchamber: `or` (not `of`, thanks) instance name really, cause the GCP instance id isn't enough to specify the notebook, it needs the full "GCP instance name" https://cloud.google.com/ai-platform/notebooks/docs/managed/reference/rest/v1beta1/projects.locations.instances

src/main/java/bio/terra/cli/command/notebooks/Get.java
marikomedlock: I'm hesitant to add a new command without a clear explanation of when to use it, though as with everything UX-related I would certainly defer to the solutions team if their feedback is different.

My preference is for the code to be ugly and the commands to look like we want, so that a future cleanup TODO is opaque to users. From the user's perspective, do you think this additional notebook information belongs in the output of:
1. `resources describe`: This was my first instinct. Agreed the code would be ugly for now, perhaps adding additional logic after `printText` in the `describe` command itself would be slightly less so?
2. `resources resolve`: Probably only relevant for the proxy url.
3. A separate `notebooks` command: 
4. rely on `gcloud`: This information is basically a shortcut for `gcloud notebooks instances describe`, with less-important properties filtered out right? So we could also leave this information out (with a TODO) until the refactoring is done, since users always have a way to look it up via `gcloud`.

More broadly, maybe this would be good to ask about at the next solutions team meeting -- if there's any larger preference around having useful Terra CLI wrappers for `gcloud` commands vs only using the Terra CLI for doing things via Terra services. I'm not sure what the right answer is, but I think the `notebooks start/stop/get` commands are the first examples we have of this.

wchamber: 1. I'll go with that, maybe adding it in the `describe` command.
2. Yeah, only kind of makes sense for the proxy url.
3. I don't have a strong philosophical stance against this.
4. gcloud doesn't seem to be printing the proxy url for these service account proxy notebooks, so its insufficient.

One point for the Terra CLI resource type specific commands is that they let us use the workspace resource name to do the operation instead of only the GCP identifier. That means you don't have to translate the resolved name to the right flags for the gcloud command (`--project --location --instanceId` for notebooks.)

I'm strongly motivated to have `start`, `stop`, and someway to get the proxy URL/state in the CLI because it's just convenient for the manual testing/demoing I'm doing, so looking for some at least short term solution here. While we don't need to recreate every gcloud command available, I think it's nice to have some convenience for the most common ones. If we're strongly opposed to having any Terra CLI wrappers around something that's gcloud feasible, I'll remove `start` & `stop` but keep the uri  & state printing in `describe`.

marikomedlock: Agreed that using the resource name is a nice feature of wrapping them in Terra CLI commands.

RE: a separate `notebooks` command. If you decide this route, maybe a more disambiguated name would help (e.g. `terra notebooks status` for the state, `terra resolve --proxy-url`). That would eliminate confusion I think.

marikomedlock (approved):

PR #75, opened 2021-05-13
PF-763 Add initial noteboks post startup script

wchamber:
Add an initial post startup script for AI Notebooks created by the CLI. This is an initial stub script that installs some python packages and the Terra CLI.
The post startup script runs before the VM has any user credentials, so it can only do public actions.
If there are any bash scripting best practices, please let me know.

In the future, we can set more metadata on the VM instance and pull it in the script from [GCE's metadata server](https://cloud.google.com/compute/docs/storing-retrieving-metadata).
In the long term, we want to move the notebooks configuration out of CLI and to a dedicated Terra notebooks repo/service.


src/main/java/bio/terra/cli/command/resources/create/AiNotebook.java
```diff
@@ -72,6 +72,9 @@
 
   @CommandLine.Option(
       names = "--post-startup-script",
+      // DO NOT SUBMIT do not use dev branch.
+      defaultValue =
+          "https://raw.githubusercontent.com/DataBiosphere/terra-cli/wc-PF-763-post-startup-script/notebooks/post-startup.sh",
```
wchamber, line 77: pointing at this branch for demoing, needs to be updated to point at main.

notebooks/post-startup.sh
```diff
@@ -0,0 +1,24 @@
...
+JUPYTER_USER="jupyter"
+
+# Install common packages in conda environment
+/opt/conda/bin/conda install -y pre-commit nbdime nbstripout pylint pytest
+# Install nbstripout for the jupyter user in all git repositories.
+sudo -u ${JUPYTER_USER} sh -c "/opt/conda/bin/nbstripout --install --global"
+
+# Install & configure the Terra CLI
+cd /tmp || exit
+sudo -u ${JUPYTER_USER} sh -c "curl -L https://github.com/DataBiosphere/terra-cli/releases/latest/download/download-install.sh | bash"
```
marikomedlock, line 21: Just for curiosity, why does the CLI installation need to be done as the jupyter_user?

notebooks/post-startup.sh
```diff
@@ -0,0 +1,24 @@
...
+# Set browser manual login since that's the only login supported from an AI Notebook VM.
+sudo -u ${JUPYTER_USER} sh -c "./terra config set browser manual"
```
marikomedlock, line 23: Is this startup script also the right place to call other config commands? e.g.:
- `terra workspace mount`
- `terra server set`
- `gcloud config set project`

notebooks/post-startup.sh
```diff
@@ -0,0 +1,24 @@
...
+# user space.
+JUPYTER_USER="jupyter"
+
+# Install common packages in conda environment
+/opt/conda/bin/conda install -y pre-commit nbdime nbstripout pylint pytest
+# Install nbstripout for the jupyter user in all git repositories.
+sudo -u ${JUPYTER_USER} sh -c "/opt/conda/bin/nbstripout --install --global"
+
+# Install & configure the Terra CLI
+cd /tmp || exit
```
marikomedlock, line 20: Should this line be moved up before the package installations, since the stdout/err file will not be generated if `/tmp` doesn't exist?

deflaux:
> Add an initial post startup script for AI Notebooks created by the CLI. This is an initial stub script that installs some python packages and the Terra CLI.

This looks like a good start for the setup relevant to all MC Terra users!

> The post startup script runs before the VM has any user credentials, so it can only do public actions.
> If there are any bash scripting best practices, please let me know.

I recommend @mbookman for the bash best practices review.

> In the future, we can set more metadata on the VM instance and pull it in the script from [GCE's metadata server](https://cloud.google.com/compute/docs/storing-retrieving-metadata).

That's great! Configuring the git user name and email doesn't require any credentials, just the correct values to use.

> In the long term, we want to move the notebooks configuration out of CLI and to a dedicated Terra notebooks repo/service.

sound good!

notebooks/post-startup.sh
```diff
@@ -0,0 +1,24 @@
...
+exec >> /tmp/post-startup-output.txt
+exec 2>&1
+set -x
+
+# The linux user that JupyterLab will be running as. It's important to do some parts of setup in the
+# user space.
+JUPYTER_USER="jupyter"
+
+# Install common packages in conda environment
+/opt/conda/bin/conda install -y pre-commit nbdime nbstripout pylint pytest
```
deflaux, line 15: looking forward to seeing in a future PR how you would like to package up the setup needed for specific tutorial notebooks (e.g. there are a default set of tutorials we install, and the setup script for those tutorials first calls this setup script)

notebooks/post-startup.sh
```diff
@@ -0,0 +1,24 @@
...
+# AI Notebook post startup scrips are run only when the instance is first created.
+
+# Send stdout and stderr from this script to a tmp file for debugging.
+exec >> /tmp/post-startup-output.txt
+exec 2>&1
+set -x
+
+# The linux user that JupyterLab will be running as. It's important to do some parts of setup in the
+# user space.
+JUPYTER_USER="jupyter"
```
mbookman, line 12: If a variable is not expected to change, then recommend using "readonly":

```
readonly JUPYTER_USER="jupyter"
```

See https://google.github.io/styleguide/shellguide.html#quoting

notebooks/post-startup.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+# Default post startup script for AI notebooks.
+# AI Notebook post startup scrips are run only when the instance is first created.
+
```
mbookman, line 4: Recommend 

```
set -o errexit
set -o nounset
set -o pipefail
```

As they help catch coding or execution errors closer to the point of failure.

notebooks/post-startup.sh
```diff
@@ -0,0 +1,24 @@
...
+set -x
+
+# The linux user that JupyterLab will be running as. It's important to do some parts of setup in the
+# user space.
+JUPYTER_USER="jupyter"
+
+# Install common packages in conda environment
+/opt/conda/bin/conda install -y pre-commit nbdime nbstripout pylint pytest
+# Install nbstripout for the jupyter user in all git repositories.
+sudo -u ${JUPYTER_USER} sh -c "/opt/conda/bin/nbstripout --install --global"
```
mbookman, line 17: Enclose variable expansion in double-quotes as a general practice.

https://google.github.io/styleguide/shellguide.html#quoting

mbookman (changes requested):

wchamber (commented):
Thanks!

marikomedlock (approved):

notebooks/post-startup.sh
deflaux: we could make a read-only deploy key for the repo and store it in the metadata

deflaux (approved):

PR #77, opened 2021-05-17
PF-718: Remove hardcoded storage class and lifecycle rule.

marikomedlock:
Removed the hard-coded storage class and lifecycle rule when creating a controlled GCS bucket. This is no longer needed now that WSM has implemented defaults. Previously, it threw a 500 error if these were not specified.

** Note: I also synced the `verily-cli` personal environment for solutions' team testing with the latest WSM code, so that this fix will work against that deployment also.


jaycarlton (approved):
FWIW, WSM isn't setting defaults; we just leave the structure empty and let whatever defaults there are on the cloud happen. I need to verify what those are actually.

wchamber (approved):

PR #79, opened 2021-05-19
PF-784: Add env var override for context directory parent.

marikomedlock:
By default, the global context directory is stored in the user's home directory (`$HOME`). Added an option to override this with the `TERRA_CONTEXT_PARENT_DIR ` environment variable.

This will be useful for testing, so that tests don't overwrite an existing installation on the same machine.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,12 +7,14 @@
     * [download-install.sh](#download-installsh)
     * [install.sh](#installsh)
     * [terra](#terra)
-3. [Docker](#docker)
+3. [Testing](#testing)
+    * [Override context directory](#override-context-directory)
+4. [Docker](#docker)
     * [Pull an existing image](#pull-an-existing-image)
     * [Build a new image](#build-a-new-image)
     * [Publish a new image](#publish-a-new-image)
     * [Update the default image](#update-the-default-image)
-4. [Code structure](#code-structure)
+5. [Code structure](#code-structure)
     * [Supported tools](#supported-tools)
         * [Adding a new supported tool](#adding-a-new-supported-tool)
     * [Authentication](#authentication)
@@ -104,6 +106,20 @@ This is the run script that wraps the Java call to the CLI.
 
 It is included in the `terra-cli.tar` install package in each GitHub release.
 This is the script users can add to their `$PATH` to invoke the CLI more easily from another directory.
+
+
+### Testing
+
+#### Override context directory
+The `.terra` context directory is stored in the user's home directory (`$HOME`) by default.
+You can override this default by setting the `TERRA_CONTEXT_DIR` environment variable to a valid directory.
+```
+export TERRA_CONTEXT_DIR="/Desktop/cli-testing"
+terra config list
+```
+If the environment variable override does not point to a valid directory, then the CLI will throw a `SystemException`.
+
+This option is intended for tests, so that they don't overwrite the context for an installation on the same machine.
 
 
 ### Docker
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -106,6 +108,20 @@ It is included in the `terra-cli.tar` install package in each GitHub release.
 This is the script users can add to their `$PATH` to invoke the CLI more easily from another directory.
 
 
+### Testing
+
+#### Override context directory
+The `.terra` context directory is stored in the user's home directory (`$HOME`) by default.
+You can override this default by setting the `TERRA_CONTEXT_DIR` environment variable to a valid directory.
+```
+export TERRA_CONTEXT_DIR="/Desktop/cli-testing"
```
jaycarlton, line 117: This is technically the parent of the context directory. It's not confusing enough that I'd necessarily say to change its behavior. Maybe name it `TERRA_CONTEXT_PARENT_DIR`  for clarity.

src/main/java/bio/terra/cli/context/GlobalContext.java
```diff
@@ -282,8 +285,22 @@ public CommandRunner getRunner(WorkspaceContext workspaceContext) {
...
+
+    // if the override environment variable is set and points to a valid directory, then use it
+    // instead
+    String overrideDirName = System.getenv(CONTEXT_DIR_OVERRIDE_NAME);
+    if (overrideDirName != null && !overrideDirName.isBlank()) {
+      Path overrideDir = Paths.get(overrideDirName).toAbsolutePath();
+      if (overrideDir.toFile().exists() && overrideDir.toFile().isDirectory()) {
+        parentDir = overrideDir;
+      } else {
+        throw new SystemException(
```
jaycarlton, line 299: Maybe `IlilegalArgumentException` would be more specific.

src/main/java/bio/terra/cli/context/GlobalContext.java
```diff
@@ -282,8 +285,22 @@ public CommandRunner getRunner(WorkspaceContext workspaceContext) {
...
+    // default to the user's home directory
     Path parentDir = Paths.get(System.getProperty("user.home"));
+
+    // if the override environment variable is set and points to a valid directory, then use it
+    // instead
+    String overrideDirName = System.getenv(CONTEXT_DIR_OVERRIDE_NAME);
+    if (overrideDirName != null && !overrideDirName.isBlank()) {
+      Path overrideDir = Paths.get(overrideDirName).toAbsolutePath();
+      if (overrideDir.toFile().exists() && overrideDir.toFile().isDirectory()) {
+        parentDir = overrideDir;
```
jaycarlton, line 297: It would be good to check and bail if there's already a `.terra` directory or file under the override. I don't think we want to be merging contexts if we can avoid it.

jaycarlton (approved):
Looks useful. Do you know if there's a plan to test the CLI inside of a container? That would also encapsulate your `gcloud` state, env vars, and stuff.

CONTRIBUTING.md
marikomedlock: Changed to `TERRA_CONTEXT_PARENT_DIR`

src/main/java/bio/terra/cli/context/GlobalContext.java
marikomedlock: On second thought, I decided to change this to a `UserActionableException` because it's an error that the user can fix. As opposed to `SystemException` and all other "unexpected" exceptions that are presumed to be a bug in our code. Those three categories correspond to specific CLI exit codes and output formatting, so I think it's good to pick the appropriate category. Thanks for the push to rethink this one.

src/main/java/bio/terra/cli/context/GlobalContext.java
marikomedlock: Every time you execute a CLI command, it's a new Java process that has to read the context file(s) into memory. So the majority of the time, I expect there will already be a `.terra` directory there and we want to reuse it.

jmczerk (approved):

PR #84, opened 2021-06-01
PF-757: Run tests nightly.

marikomedlock:
- Added a new GitHub action to run tests nightly.
  - Runs unit tests in job 1.
  - Runs integration tests against an installation built from source code via `./gradlew installDist` in job 2.
  - Runs integration tests against an installation built from the latest GitHub release in job 3.
- Each test job uploads a GitHub action archive with the global context file and all log files generated during the test run.

- Two other fixes/changes:
  - Fixed a bug in the Nextflow integration test. It was fine on the Linux GH action runner, but on my Mac `uuidgen` produces uppercase letters, which are [not allowed](https://cloud.google.com/storage/docs/naming-buckets#requirements) for GCS bucket names.
  - Removed the integration test cleanup code that deleted all resources before deleting a workspace. WSM does this for us now, so it's no longer needed. We can just delete the workspace directly.

### DIFF
--- a/.github/workflows/tests-nightly.yml
+++ b/.github/workflows/tests-nightly.yml
@@ -15,7 +15,6 @@ jobs:
         testOptions: [ "-PtestTag=unit", "-PtestTag=integration", "-PtestTag=integration -PtestInstallFromGitHub" ]
       fail-fast: false
     runs-on: ubuntu-latest
-    if: "!contains( github.event.sender.login, 'broadbot')"
     steps:
       - name: Checkout current code
         id: checkout_code
--- a/.github/workflows/tests-nightly.yml
+++ b/.github/workflows/tests-nightly.yml
@@ -1,35 +1,18 @@
-name: Run tests against PRs (pre and post merge)
+name: Run tests nightly
 on:
   workflow_dispatch: {}
-  push:
-    branches:
-      - main
+  schedule:
+    - cron: '0 5 * * *' # 5AM UTC = 12AM EST
+   # TODO: remove this before merging
   pull_request:
     branches:
       - '**'
 
 jobs:
-  lint-and-static-analysis:
-    runs-on: ubuntu-latest
-    if: "!contains( github.event.sender.login, 'broadbot')"
-    steps:
-      - name: Checkout current code
-        id: checkout_code
-        uses: actions/checkout@v2
-        with:
-          token: ${{ secrets.BROADBOT_GITHUB_TOKEN }}
-      - name: Run linter
-        id: run_linter
-        run: |
-          ./gradlew spotlessCheck
-      - name: Run static analysis
-        id: run_static_analysis
-        run: |
-          ./gradlew spotbugsMain spotbugsTest
-  tests-against-source-code:
+  tests-against-source-code-and-latest-install:
     strategy:
       matrix:
-        testTag: [ "unit", "integration" ]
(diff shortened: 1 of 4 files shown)


.github/workflows/tests-nightly.yml
```diff
@@ -0,0 +1,67 @@
+name: Run tests nightly
+on:
+  workflow_dispatch: {}
+  schedule:
+    - cron: '0 5 * * *' # 5AM UTC = 12AM EST
+   # TODO: remove this pull_request section before merging. it's just for testing this GH action pre-merge
```
jaycarlton, line 6: remove

.github/workflows/tests-on-pr-push-and-merge.yml
```diff
@@ -40,10 +40,12 @@ jobs:
         with:
           token: ${{ secrets.BROADBOT_GITHUB_TOKEN }}
       - name: Set up AdoptOpenJDK 11
+        id: setup_jdk
```
jaycarlton, line 43: nit: Are you just putting IDs on everything, or was this needed for something specific? Other workflows are only doing it when it's referenced somewhere.

jaycarlton (approved):

.github/workflows/tests-on-pr-push-and-merge.yml
marikomedlock: Yes, I just put IDs on everything as a best practice. I'm not using it for anything here.

.github/workflows/tests-nightly.yml
```diff
@@ -0,0 +1,67 @@
...
+        run: |
+          # this step does the equivalent of the tools/render-config.sh script.
+          # on local machines, the script fetches a SA from Vault.
+          # in GH actions, the SA key is stored in a GH repo secret.
+          # regardless of how it was fetched, tests expect this
+          # key to be stored in rendered/test-user-account.json
+          mkdir -p rendered
+          echo "$TEST_USER_SA_KEY" > rendered/test-user-account.json
+        env:
+          TEST_USER_SA_KEY: ${{ secrets.TEST_USER_SA_KEY }}
```
wchamber, line 52: not this PR, but in the past dev ops has asked us to keep secrets like test accounts in Vault and only store Vault tokens as github secrets. Then, the GHA uses the vault token to get the service account secret from vault. This helps when rotating credentials. It also makes it easy to access more secrets as needed.
We also get the advantage of being able to run `tools/render-config.sh` script directly. e.g. buffer: https://github.com/DataBiosphere/terra-resource-buffer/blob/df45bd5c4ab163e7651d930529092cfaf07281fc/.github/workflows/test.yml#L35-L49
I found getting a pattern to be a little bit of "which DevOps engineer happened to talk to you" but consider filing a ticket to follow up with the Vault pattern for the CLI.

wchamber (approved):

.github/workflows/tests-nightly.yml
marikomedlock: I specifically asked DevOps about this and they told me to put the secret as a GH secret instead of accessing Vault directly ([slack conversation](https://broadinstitute.slack.com/archives/CADM7MZ35/p1615817519243700)). So definitely seems to be dependent on who you talk to.

I'd prefer to have an agreed upon central approach before making any changes here. Being able to use the same `render-config` script would be good. Also, Dan told me that they recently changed the Artifactory password in Vault, which broke lots of GH actions that had copy/pasted the value into a secret. That's another possible reason to move away from this model, or have a list somewhere of which GH secrets need to be updated, etc.

PR #85, opened 2021-06-04
PF-803: Tests for `workspace` commands.

marikomedlock:
- Changes to test code
	- New tests for the `terra workspace` commands. This includes working with workspaces (create, update, delete, list, set) and workspace users/sharees (add-user, remove-user, list-users).
	- Additional logging for unit tests, so we can see the commands and output as the test runs. e.g.
	```
	    COMMAND: config set image --default
	    Docker image: gcr.io/terra-cli-dev/terra-cli/0.46.0:stable (UNCHANGED)

	    Logging in test user: Ethan.Bonechewer@test.firecloud.org
	    COMMAND: workspace create

	    STDERR --------------
	    [ERROR] Error creating a new workspace: {"message":"User is unauthorized to link spend profile [wm-default-spend-profile]","statusCode":403,"causes":[]}
	    See /Users/marikomedlock/Workspaces/terra-cli/build/test-context/.terra/logs/terra.log for more information
	```
	- Tooling: utility methods for checking the exit code and choosing a test user, new base class for test that can reuse a single workspace for all test methods.

- Changes to application code
	- Added builders to all the classes in the `serialization.userfacing` package to allow deserialization. Previously I didn't include these because there was no use case for deserializing command output in Java. But tests are that use case.
	- Moved 3 inline pojo classes used for `--format=json` command output into separate classes in the `serialization.userfacing` package. These 3 classes are for the commands: `terra status`, `terra auth status`, `terra config get-value logging`. This was just for clearer organization.
	- Fixed a bug when looking up a workspace user/sharee where the email address casing in SAM may not match the case of what is provided by the user.
	- Fixed a bug when removing the last role from a user in the workspace.
	- Removed the list of resources from the workspace `--format=json` command output. Now it only includes workspace metadata.


src/test/java/unit/Workspace.java
```diff
@@ -0,0 +1,261 @@
...
+    // `terra workspace create --format=json`
+    UFWorkspace createWorkspace =
+        TestCommand.runCommandExpectSuccess(
+            UFWorkspace.class, "workspace", "create", "--format=json");
+
+    // check the created workspace has an id and a google project
+    assertNotNull(createWorkspace.id, "create workspace returned a workspace id");
+    assertNotNull(createWorkspace.googleProjectId, "create workspace created a gcp project");
+    assertThat(
+        "workspace email matches test user",
```
wchamber, line 42: optional style nit: I don't think the comments & messages describing *what* the assertions are doing are especially helpful, especially with the message for the assertion. I think `assertNotNull(createWorkspace.id);` on its own is pretty clear that we're checking that there's a workspace id and there's a lot more to read otherwise. Fine to leave when the assertion is fulfilling a non-obvious purpose without a named helper function, but most of these seem self-explanatory. [go/java-practices/comments](go/java-practices/comments)

src/test/java/unit/Workspace.java
```diff
@@ -0,0 +1,261 @@
...
+        TestCommand.runCommandExpectSuccess(
+            new TypeReference<>() {}, "workspace", "list", "--format=json");
+
+    // check the workspace list reflects the update
+    List<UFWorkspace> matchingWorkspaces =
+        listWorkspaces.stream()
+            .filter(workspace -> workspace.id.equals(createWorkspace.id))
+            .collect(Collectors.toList());
+    assertEquals(
+        1, matchingWorkspaces.size(), "updated workspace is included exactly once in list");
```
wchamber, line 182: optional: consider a helper function for `private static UFWorkspace checkOneListedWorkspaceMatching(Predicate<UFWorkspace> predicate)` since this seems to be a common pattern in these tests.

src/test/java/unit/WorkspaceUser.java
```diff
@@ -0,0 +1,191 @@
...
+  protected void setupEachTime() throws IOException {
+    workspaceCreator.login();
+
+    // `terra workspace set --id=$id`
+    TestCommand.runCommandExpectSuccess("workspace", "set", "--id=" + getWorkspaceId());
+
+    // `terra workspace list-users --format=json`
+    List<UFWorkspaceUser> listWorkspaceUsers =
+        TestCommand.runCommandExpectSuccess(
+            new TypeReference<>() {}, "workspace", "list-users", "--format=json");
```
wchamber, line 34: The `new TypeReference<>() {}` looks kind of weird. Maybe we should have a runCommandExpectSuccessList for List<T>, but maybe that's overkill.

src/test/java/unit/WorkspaceUser.java
```diff
@@ -0,0 +1,191 @@
...
+        continue;
+      }
+      for (IamRole role : user.roles) {
+        // `terra workspace remove-user --email=$email --role=$role
+        TestCommand.runCommandExpectSuccess(
+            "workspace", "remove-user", "--email=" + user.email, "--role=" + role);
+      }
+    }
+
+    super.setupEachTime();
```
wchamber, line 47: how do we know if we should do this last or first?

src/test/java/unit/WorkspaceUser.java
```diff
@@ -0,0 +1,191 @@
...
+            new TypeReference<>() {}, "workspace", "list-users", "--format=json");
+
+    // check that the user is in the list as a reader
+    Optional<UFWorkspaceUser> workspaceUser =
+        listWorkspaceUsers.stream()
+            .filter(user -> user.email.equalsIgnoreCase(testUser.email))
+            .findAny();
+    assertTrue(workspaceUser.isPresent(), "test user is in users list");
+    assertEquals(1, workspaceUser.get().roles.size(), "test user has one role");
+    assertTrue(workspaceUser.get().roles.contains(IamRole.READER), "test user has reader role");
```
wchamber, line 107: consider a helper to cut down boilerplate and improve test readability
`private static UFWorkspaceUser expectListedUserWithRoles(String userEmail, IamRoles... roles);`

src/test/java/unit/WorkspaceUser.java
```diff
@@ -0,0 +1,191 @@
...
+            UFWorkspaceUser.class,
+            "workspace",
+            "add-user",
+            "--email=" + testUser.email,
+            "--role=WRITER",
+            "--format=json");
+
+    // check that the user has both the READER and WRITER roles
+    assertTrue(addUserWriter.roles.contains(IamRole.READER), "reader role returned by add-user");
+    assertTrue(addUserWriter.roles.contains(IamRole.WRITER), "writer role returned by add-user");
```
wchamber, line 121: nit: consider collapsing to a single line `contains(IamRole.READER, IamRole.WRITER)`

wchamber (commented):
Looks good, just some suggestions on tests.

Please consider splitting this into multiple PRs next time. Adding more UF classes and deserialization could be separated from adding tests. Two 600 line PRs are easier and faster to review than one 1200 line PR.

src/test/java/harness/TestCommand.java
```diff
@@ -34,19 +37,63 @@ public static Result runCommand(String... args) {
...
     return objectMapper.readValue(cmd.stdOut, objectType);
   }
 
+  /** Convert what's written to standard out into a Java object. */
+  public static <T> T readObjectFromStdOut(Result cmd, TypeReference<T> objectType)
+      throws JsonProcessingException {
+    return objectMapper.readValue(cmd.stdOut, objectType);
+  }
+
+  /** Helper method to run a command and check its exit code is 0=success. */
```
jmczerk, line 68: Big time nit, but is anyone actually using the returned `Result`?  I looked for this in the test code and didn't see any, but might have missed, but if not...

With the three overloads it feels a little unexpected to me that the first would return the local `Result` while the other two throw it away and return something else entirely.  If the return is indeed not used anywhere, maybe consider `void` return, otherwise maybe a different name for the two that return parsed results?

src/main/java/bio/terra/cli/command/workspace/RemoveUser.java
```diff
@@ -21,18 +19,10 @@
       description = "Role to grant: ${COMPLETION-CANDIDATES}")
   private IamRole role;
 
-  @CommandLine.Mixin Format formatOption;
```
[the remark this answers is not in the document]
jmczerk: +1 on this question

jmczerk (commented):
+1'ed on one of Will's questions about Json mixin disappearing, and agree on the two suggestions for helper methods for repeated code in some of the tests.  Also had one minor nit around the `runCommandExpectSuccess()` overloads.  But otherwise looks good.

src/main/java/bio/terra/cli/command/workspace/RemoveUser.java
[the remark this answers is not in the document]
marikomedlock: I thought it was confusing what this returned. Maybe I've just been staring at this too long though. Suggestions welcome.

`remove-user` actually removes a user email + role combination, not an entire user from a workspace. So I could see 3 options for return values here:
- Return the email + role that got removed. That would be consistent with other delete commands that return the deleted thing. But that means we're using `UFWorkspaceUser` to mean different things for different commands. For listing workspace users, it holds email + all roles that email has. For removing a user, it holds email + one role that email had (it may have had others).
- Return the new state of the email + roles in the workspace. That would be inconsistent with other delete commands, but would mean we're using `UFWorkspaceUser` in a consistent way.
- Return just the role that got removed. This seemed useless because the user supplied the role in the command, so they don't need the cli to spit it back out.

Instead of trying to pick a less confusing option, I decided to just remove the json output and let users call `list-users` with JSON format to get the current state of the user they're interested in.

src/test/java/harness/TestCommand.java
```diff
@@ -34,19 +37,63 @@ public static Result runCommand(String... args) {
...
 
-  /** Helper method to convert what's written to standard out into a Java object. */
+  /** Convert what's written to standard out into a Java object. */
   public static <T> T readObjectFromStdOut(Result cmd, Class<T> objectType)
       throws JsonProcessingException {
     return objectMapper.readValue(cmd.stdOut, objectType);
   }
 
+  /** Convert what's written to standard out into a Java object. */
+  public static <T> T readObjectFromStdOut(Result cmd, TypeReference<T> objectType)
```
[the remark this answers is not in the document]
marikomedlock: Good idea. done.

src/test/java/harness/TestUsers.java
```diff
@@ -100,6 +100,21 @@ public void login() throws IOException {
     return dataStoreFactory.getDataStore(StoredCredential.DEFAULT_DATA_STORE_ID);
   }
 
+  /**
+   * Randomly chooses a test user, who is anyone except for the given test user. Helpful e.g.
```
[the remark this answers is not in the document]
marikomedlock: Added a comment to the class, since the randomness is in several other utility methods.

src/test/java/harness/baseclasses/SingleWorkspaceUnit.java
```diff
@@ -0,0 +1,54 @@
...
+import harness.TestCommand;
+import harness.TestContext;
+import harness.TestUsers;
+import java.io.IOException;
+import java.util.UUID;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+
+/**
+ * Base class for unit tests that only need a single workspace for all test methods. This makes the
```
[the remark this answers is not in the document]
marikomedlock: Definitely a tradeoff between requiring more boilerplate code in all classes that need this, vs allowing other setup/cleanup in the future. At the moment I don't have a use case for multiple conditions, so going to err on the side of less boilerplate. If we do end up with lots of inheritance, then agreed composition seems like a good way to address it.

src/test/java/harness/baseclasses/SingleWorkspaceUnit.java
```diff
@@ -0,0 +1,54 @@
...
+
+  @AfterAll
+  static void cleanupOnce() throws IOException {
+    TestContext.clearGlobalContextDir();
+    resetContext();
+
+    // login as the same user that created the workspace
+    workspaceCreator.login();
+
+    // `terra workspace set --id=$id`
```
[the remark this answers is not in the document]
marikomedlock: I prefer using comments for documentation over adding in an extra whitespace parsing step. I also think it'd be easy to forget when to use the non-split version.

src/test/java/unit/Workspace.java
```diff
@@ -0,0 +1,261 @@
...
+import org.hamcrest.CoreMatchers;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+/** Tests for the `terra workspace` commands. */
+@Tag("unit")
+public class Workspace extends ClearContextUnit {
+  @Test
+  @DisplayName("status, workspace list reflect workspace create")
```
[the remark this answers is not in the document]
marikomedlock: This shows up in the JUnit output, as descriptions of the tests that ran. I find that helpful. If we're worried about consistency, I'd advocate for including this in other repos instead of removing it here.

src/test/java/unit/Workspace.java
marikomedlock: These messages show up in the stack trace when an assertion fails. I can certainly just look at the line number and then go to the code to see what failed, but I find it helpful to see a readable error message in stdout, particularly when I'm looking at a GH action output and the line numbers don't match exactly with the code I've got on my machine.

src/test/java/unit/Workspace.java
marikomedlock: Added a helper method to list and filter the workspaces.

src/test/java/unit/WorkspaceUser.java
marikomedlock: It should be last so that the context is cleared just before each test method. I'll add a note on the superclass.

src/test/java/unit/WorkspaceUser.java
marikomedlock: OK going to hold off on this for now.

src/test/java/unit/WorkspaceUser.java
marikomedlock: Added two helper methods.

src/test/java/harness/TestCommand.java
marikomedlock: Good point. Renamed as follows:
- `runAndParseCommandExpectSuccess` (two methods) return the parsed result
- `runCommandExpectSuccess` (one method) has a `void` return
- `runCommand` (one method) returns the `Result` object. This one is used for checking error messages.

src/main/java/bio/terra/cli/command/workspace/RemoveUser.java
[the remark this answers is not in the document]
wchamber: Makes sense. Either returning the new state of email + roles or nothing both sound good to me.

wchamber (approved):

src/main/java/bio/terra/cli/command/workspace/RemoveUser.java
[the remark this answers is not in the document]
jmczerk: Thanks, returning nothing makes sense to me due to potential confusion and the ability to infer the resulting state.

jmczerk (approved):
Looks good.

PR #86, opened 2021-06-08
PF-825: Tests for referenced and controlled GCS bucket resources.

marikomedlock:
- Added tests for referenced and controlled GCS bucket resources.
- The referenced GCS bucket tests generate buckets in an external project using a SA. This project is currently set to `terra-cli-dev`, but should be changed in a future ticket to use different hard-coded project + SA that is generated by Terraform (PF-829).
- Removed some Jackson annotations in the `UFResource` classes so that we don't include the `@class` property in the user-facing command output.

### DIFF
--- a/src/test/java/harness/utils/ExternalGCSBuckets.java
+++ b/src/test/java/harness/utils/ExternalGCSBuckets.java
@@ -1,4 +1,4 @@
-package harness;
+package harness.utils;
 
 import com.google.auth.oauth2.GoogleCredentials;
 import com.google.auth.oauth2.ServiceAccountCredentials;
@@ -19,12 +19,21 @@ import java.util.UUID;
 import java.util.concurrent.TimeUnit;
 
 /** Utility methods for creating external GCS buckets for testing workspace references. */
-public class TestGCSBucketReferences {
+public class ExternalGCSBuckets {
   public static final String gcpProjectId = "terra-cli-dev";
 
   private static final String saKeyFile = "./rendered/ci-account.json";
   private static final List<String> cloudPlatformScope =
       Collections.unmodifiableList(Arrays.asList("https://www.googleapis.com/auth/cloud-platform"));
+
+  /**
+   * Get a bucket. This is helpful for testing controlled GCS bucket resources. It allows tests to
+   * check metadata that is not stored in WSM, only in GCS.
+   */
+  public static Bucket getBucket(String bucketName, GoogleCredentials credentials)
+      throws IOException {
+    return getStorageClient(credentials).get(bucketName);
+  }
 
   /**
    * Create a bucket in an external project. This is helpful for testing referenced GCS bucket
@@ -89,9 +98,14 @@ public class TestGCSBucketReferences {
     GoogleCredentials saCredentials =
         ServiceAccountCredentials.fromStream(new FileInputStream(saKeyFile))
             .createScoped(cloudPlatformScope);
+    return getStorageClient(saCredentials);
+  }
+
+  /** Helper method to build the GCS client object with the given credentials. */
+  private static Storage getStorageClient(GoogleCredentials credentials) throws IOException {
     return StorageOptions.newBuilder()
         .setProjectId(gcpProjectId)
-        .setCredentials(saCredentials)
+        .setCredentials(credentials)
         .build()
         .getService();
   }
--- a/src/test/java/harness/utils/ExternalGCSBuckets.java
+++ b/src/test/java/harness/utils/ExternalGCSBuckets.java
@@ -16,7 +16,6 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.UUID;
-import java.util.concurrent.TimeUnit;
 
 /** Utility methods for creating external GCS buckets for testing workspace references. */
 public class ExternalGCSBuckets {
@@ -90,11 +89,6 @@ public class ExternalGCSBuckets {
                 .addIdentity(StorageRoles.legacyBucketReader(), Identity.user(email))
                 .build());
     getStorageClient().setIamPolicy(bucket.getName(), updatedPolicy);
-    try {
-      TimeUnit.SECONDS.sleep(20);
-    } catch (InterruptedException e) {
-      e.printStackTrace();
-    }
   }
 
   /** Helper method to build the GCS client object with SA credentials for an external project. */
--- a/src/test/java/harness/utils/ExternalGCSBuckets.java
+++ b/src/test/java/harness/utils/ExternalGCSBuckets.java
@@ -28,7 +28,8 @@ public class ExternalGCSBuckets {
 
   /**
    * Get a bucket. This is helpful for testing controlled GCS bucket resources. It allows tests to
-   * check metadata that is not stored in WSM, only in GCS.
+   * check metadata that is not stored in WSM, only in GCS. This method takes in the credentials to
+   * use because tests typically want to check metadata as the test user.
    */
   public static Bucket getBucket(String bucketName, GoogleCredentials credentials)
       throws IOException {
@@ -37,7 +38,7 @@ public class ExternalGCSBuckets {
 
   /**
    * Create a bucket in an external project. This is helpful for testing referenced GCS bucket
-   * resources.
+   * resources. This method uses SA credentials for an external project.
    */
   public static Bucket createBucket() throws IOException {
     String bucketName = UUID.randomUUID().toString();
@@ -66,13 +67,16 @@ public class ExternalGCSBuckets {
 
   /**
    * Delete a bucket in an external project. This is helpful for testing referenced GCS bucket
-   * resources.
+   * resources. This method uses SA credentials for an external project.
    */
   public static void deleteBucket(Bucket bucket) throws IOException {
     getStorageClient().delete(bucket.getName());
   }
 
-  /** Grant a given user object viewer access to a bucket. */
+  /**
+   * Grant a given user object viewer access to a bucket. This method uses SA credentials for an
+   * external project.
+   */
   public static void grantReadAccess(Bucket bucket, String email) throws IOException {
     Storage storage = getStorageClient();
     Policy currentPolicy = storage.getIamPolicy(bucket.getName());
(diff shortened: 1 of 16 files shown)


src/test/java/unit/GcsBucketControlled.java
```diff
@@ -0,0 +1,284 @@
...
+
+    assertEquals(1, matchedResources.size(), "found exactly one resource with this name");
+    return matchedResources.get(0);
+  }
+
+  /**
+   * Helper method to call `terra resources list` and filter the results on the specified resource
+   * name.
+   */
+  static List<UFGcsBucket> listBucketResourceWithName(String resourceName)
```
jaycarlton, line 272: Since there's at most one resource with the name, you could make this return `Optional<UFGcsBucket>`. Just use `findFirst()` instead of `collect()`.

src/test/java/unit/GcsBucketReferenced.java
```diff
@@ -0,0 +1,239 @@
...
+    // check that the name and bucket name match
+    assertEquals(name, addedBucket.name, "add ref output matches name");
+    assertEquals(
+        externalBucket.getName(), addedBucket.bucketName, "add ref output matches bucket name");
+
+    // check that the bucket is in the list
+    UFGcsBucket matchedResource = listOneBucketResourceWithName(name);
+    assertEquals(name, matchedResource.name, "list output matches name");
+    assertEquals(
+        externalBucket.getName(), (matchedResource).bucketName, "list output matches bucket name");
```
jaycarlton, line 71: nit: extra parens

src/test/java/unit/GcsBucketReferenced.java
```diff
@@ -0,0 +1,239 @@
...
+
+    // `terra resources describe --name=$name --format=json`
+    UFGcsBucket describeResource =
+        TestCommand.runAndParseCommandExpectSuccess(
+            UFGcsBucket.class, "resources", "describe", "--name=" + name, "--format=json");
```
jaycarlton, line 76: You could move the `--format=json` into the helper method since you need it every time.

src/test/java/harness/utils/ExternalGCSBuckets.java
```diff
@@ -0,0 +1,110 @@
...
+      Collections.unmodifiableList(Arrays.asList("https://www.googleapis.com/auth/cloud-platform"));
+
+  /**
+   * Get a bucket. This is helpful for testing controlled GCS bucket resources. It allows tests to
+   * check metadata that is not stored in WSM, only in GCS. This method takes in the credentials to
+   * use because tests typically want to check metadata as the test user.
+   */
+  public static Bucket getBucket(String bucketName, GoogleCredentials credentials)
+      throws IOException {
+    return getStorageClient(credentials).get(bucketName);
```
wchamber, line 35: I'm not seeing a lot of value in wrapping the `Storage` class. Why have each of these wrapping helper methods instead of using the `Storage` (or CRL's `StorageCow`) directly? `grantReadAccess` & `getStorageAccess` do some helping, but create/get/delete not as much.

src/test/java/unit/GcsBucketReferenced.java
```diff
@@ -0,0 +1,239 @@
...
+public class GcsBucketReferenced extends SingleWorkspaceUnit {
+
+  // external bucket to use for creating GCS bucket references in the workspace
+  private Bucket externalBucket;
+
+  @BeforeEach
+  @Override
+  protected void setupEachTime() throws IOException {
+    super.setupEachTime();
+    externalBucket = ExternalGCSBuckets.createBucket();
```
wchamber, line 32: do we need a new bucket for every test case or can we share one across all?

jaycarlton (approved):
Very clean and thorough.

src/test/java/unit/GcsBucketReferenced.java
```diff
@@ -0,0 +1,239 @@
...
+    ExternalGCSBuckets.grantReadAccess(externalBucket, workspaceCreator.email);
+  }
+
+  @AfterEach
+  protected void cleanupEachTime() throws IOException {
+    ExternalGCSBuckets.deleteBucket(externalBucket);
```
wchamber, line 38: This test can leak buckets if there's a temporary error trying to delete the bucket, or if the test is interrupted during the middle of execution. CRL + Janitor is exactly for this use case. Consider using the `StorageCow` with the CleanupClient config set to talk to Janitor, and the tools Janitor is given permissions to delete buckets in this google project.

src/test/java/unit/GcsBucketReferenced.java
```diff
@@ -0,0 +1,239 @@
...
+        externalBucket.getName(),
+        describeResource.bucketName,
+        "describe resource output matches bucket name");
+
+    // `terra resources delete --name=$name`
+    TestCommand.runCommandExpectSuccess("resources", "delete", "--name=" + name);
+  }
+
+  @Test
+  @DisplayName("list reflects deleting a referenced bucket")
```
wchamber, line 90: optional, it's reasonable to collapse this with  listDescribeReflectAdd, but up to you. ditto on the controlled buckets test.

src/test/java/unit/GcsBucketReferenced.java
```diff
@@ -0,0 +1,239 @@
...
+    // `terra resources check-access --name=$name
+    TestCommand.runCommandExpectSuccess("resources", "check-access", "--name=" + name);
+
+    // `terra resources delete --name=$name`
+    TestCommand.runCommandExpectSuccess("resources", "delete", "--name=" + name);
+  }
+
+  @Test
+  @DisplayName("add a referenced bucket, specifying all options except lifecycle")
+  void addWithAllOptionsExceptLifecycle() throws IOException {
```
wchamber, line 192: Do referenced buckets have lifecycles? What does a lifecycle mean then?

wchamber (changes requested):

src/test/java/harness/utils/ExternalGCSBuckets.java
```diff
@@ -0,0 +1,110 @@
...
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.UUID;
+
+/** Utility methods for creating external GCS buckets for testing workspace references. */
+public class ExternalGCSBuckets {
+  public static final String gcpProjectId = "terra-cli-dev";
```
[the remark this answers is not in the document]
marikomedlock: Right now there's only going to be a single external-to-WSM project for testing. So I think I'm going to leave it hard-coded for now. If we can optionally use multiple projects in the future, then agreed using a Gradle property or something to let developers choose which one they want to use seems reasonable.

src/test/java/unit/GcsBucketControlled.java
```diff
@@ -0,0 +1,284 @@
...
+            "--format=json");
+
+    // check that the name and bucket name match
+    assertEquals(name, createdBucket.name, "create output matches name");
+    assertEquals(bucketName, createdBucket.bucketName, "create output matches bucket name");
+
+    // check that the bucket is in the list
+    UFGcsBucket matchedResource = listOneBucketResourceWithName(name);
+    assertEquals(name, matchedResource.name, "list output matches name");
+    assertEquals(bucketName, (matchedResource).bucketName, "list output matches bucket name");
```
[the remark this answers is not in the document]
marikomedlock: Huh no idea why I did that! Removed thanks

src/test/java/unit/GcsBucketControlled.java
marikomedlock: Originally I was doing this with `findAny()`, but changed it because I think it's better to assert that there's at most one resource with the name, instead of assuming it.

src/test/java/unit/GcsBucketReferenced.java
marikomedlock: removed

src/test/java/unit/GcsBucketReferenced.java
marikomedlock: Good idea. Done.

src/test/java/unit/GcsBucketReferenced.java
marikomedlock: good question. I'm not sure whether WSM checks for duplicate references or not. will check

src/test/java/unit/GcsBucketReferenced.java
marikomedlock: Good idea. I'll add this to the follow-on ticket to replace this test project with a Terraformed one.

Two questions:
- Is there any difference in telling Janitor to clean up buckets in a project and running a script to wipe all buckets in the project say once a night? Just trying to understand what Janitor gives us here.
- I thought Janitor was being used for project cleanup, which are not as easily contained as buckets in a project. Are there any other places we're using Janitor to cleanup things other than projects?

src/test/java/unit/GcsBucketReferenced.java
marikomedlock: Yeah I had it that way originally, but it seemed like the test method got really long.

src/test/java/unit/GcsBucketReferenced.java
marikomedlock: No, just a copy/paste typo. removed

src/test/java/unit/GcsBucketReferenced.java
wchamber: - You don't have to make sure your script runs successfully and it's called on all the relevant projects and those projects are allowed to have buckets that shouldn't be deleted.
- Yes, for the CRL-tests themselves where there is a terraformed gcp project and resources including buckets are created within them. It's pretty identical to what's going on here.

src/test/java/unit/GcsBucketReferenced.java
marikomedlock: looks like it doesn't check, so we can use the same bucket over and over as long as the resource name is different.

src/test/java/harness/utils/ExternalGCSBuckets.java
wchamber: You can also use CRL to log the request/result. Currently it's at DEBUG, but you can configure the logger for the `StorageCow` class to INFO.

wchamber (approved):

PR #87, opened 2021-06-08
PF-830: Tests for referenced and controlled BQ dataset resources.

marikomedlock:
- Added tests for referenced and controlled BQ dataset resources.
- The referenced BQ dataset tests generate datasets in an external project using a SA. This project is currently set to `terra-cli-dev`, but should be changed in a future ticket to use different hard-coded project + SA that is generated by Terraform (PF-829).
- Pulled out the pointers to the external project and SA into a separate class, for easier reuse.

### DIFF
--- a/src/test/java/harness/utils/ExternalBQDatasets.java
+++ b/src/test/java/harness/utils/ExternalBQDatasets.java
@@ -1,0 +1,39 @@
+package harness.utils;
+
+import com.google.auth.oauth2.GoogleCredentials;
+import com.google.cloud.bigquery.BigQuery;
+import com.google.cloud.bigquery.BigQueryOptions;
+import com.google.cloud.bigquery.Dataset;
+import com.google.cloud.bigquery.DatasetId;
+import harness.TestExternalResources;
+import java.io.IOException;
+import java.util.UUID;
+
+/** Utility methods for creating external GCS buckets for testing workspace references. */
+public class ExternalBQDatasets {
+
+  /** Helper method to generate a random dataset id. */
+  public static String randomDatasetId() {
+    return UUID.randomUUID().toString().replace('-', '_');
+  }
+
+  /**
+   * Get a dataset. This is helpful for testing controlled BQ dataset resources. It allows tests to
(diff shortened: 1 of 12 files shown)


src/test/java/harness/utils/ExternalBQDatasets.java
```diff
@@ -0,0 +1,69 @@
+package harness.utils;
+
+import static harness.TestExternalResources.getProjectId;
+import static harness.TestExternalResources.getSACredentials;
```
wchamber, line 4: optional style nit: I don't think these method names are so unambiguous that staticly importing them helps with readability. `TestExternalResources.getProjectId()` I think improves readability in this file.
Ditto on external buckets file.

wchamber (approved):
Just noticing the test packages now. Not this PR, but I think they should all go under `bio.terra.cli`. Instead of top level `harness` and `unit` packages.

marikomedlock:
> Just noticing the test packages now. Not this PR, but I think they should all go under `bio.terra.cli`. Instead of top level e.g. `import bio.terra.cli.harness` instead of `import harness`.

I thought the reason for putting tests in the same package structure is so that they could optionally access package/protected methods that would otherwise be inaccessible. The CLI tests will always call the the top-level `Main` class or run a bash script in a separate process, so that won't apply here. Are there other reasons for doing this?

jmczerk (approved):

wchamber:
> > Just noticing the test packages now. Not this PR, but I think they should all go under `bio.terra.cli`. Instead of top level e.g. `import bio.terra.cli.harness` instead of `import harness`.
> 
> I thought the reason for putting tests in the same package structure is so that they could optionally access package/protected methods that would otherwise be inaccessible. The CLI tests will always call the the top-level `Main` class or run a bash script in a separate process, so that won't apply here. Are there other reasons for doing this?

General best practice of correct java package scoping. Consistency with other repos.

src/test/java/harness/utils/ExternalBQDatasets.java
marikomedlock: Removed static imports

PR #88, opened 2021-06-09
PF-808 Clone Prototype

jaycarlton:
This is a quick hack to play with clone concepts and possibly additional input modes and possibly do some UX experimentation. It answered a couple of questions and got me familiar with our external API again. Maybe it's useful as sample code.

Things to add if we decide to keep this going would be

- copy BQ dataset data
- copy GCS bucket lifecycle rules and storage class
- conform more fully to the design doc

I want to do any follow-on work we decide to do on other tickets so we can close out PF-808.


marikomedlock (commented):
I'm hesitant to merge POC code in that we're not planning to use or build-on. Perhaps one way to keep this around in a way that's useful to the CLI is as an integration test. e.g. Write a bash script that calls `clone-workspace.py` and then checks that the number of resources match in the source & destination workspaces. What do you think?

jaycarlton:
That could work. I was also thinking maybe a samples directory or
something.

-Jay

marikomedlock:
> That could work. I was also thinking maybe a samples directory or something.

Working sample code is useful. I think it's really easy for it to become out of date if we're not running it regularly. That's the only reason for the testing suggestion. This doesn't cover any code paths that are not already hit by other CLI tests, so it doesn't really add to the CLI test bank in that way, but I think more tests are generally better than less.

If you don't want to get too far into CLI assertions, you could also just put `set -e` at the top of a bash script that calls your python code in the way you want to demonstrate, so that it exits if anything returns an error. And then assert that the script completed successfully (See `integration.Nextflow` for an example of this). That would be my base-level expectation for a samples directory anyway -- things run without error. Next level would be to add assertions that the script does what it's supposed to. Up to you how well-tested you want to keep your example.

jaycarlton:
Not going to have a use for this any time soon. I'll just keep the branch open until it becomes useful.

PR #90, opened 2021-06-10
[no ticket] fix set server command in install script

wchamber:
This is causing some errors in the notebooks post startup script where the CLI install doesn't finish with a successful error code (even though the CLI is actually basically set up by the time it errors here).

This breaks backwards compatibility for installing some past version where the set server didn't use `--name`.


marikomedlock (approved):
Thanks!

PR #91, opened 2021-06-10
[no ticket] Set the terra workspace during AI notebook startup.

wchamber:


wchamber:
We need user credentials to set the workspace id at this time.

PR #92, opened 2021-06-10
PF-513: Add retries to WSM calls.

marikomedlock:
- Add retries for WSM calls and include more retryable status code for SAM calls.
- Both now retry 500, 502, 503, 504 (so all the 5xx except 501 not implemented, 505 http version not supported).
- Also fixed a bug where adding a user to a SAM group failed with a bad request if the user is not already registered. Changed to invite the user if they're not found, and then add them to the group.

### DIFF
--- a/src/main/java/bio/terra/cli/service/SamService.java
+++ b/src/main/java/bio/terra/cli/service/SamService.java
@@ -7,6 +7,7 @@ import bio.terra.cli.service.utils.HttpUtils;
 import com.google.api.client.http.HttpStatusCodes;
 import com.google.auth.oauth2.AccessToken;
 import java.io.IOException;
+import java.net.SocketTimeoutException;
 import java.util.List;
 import org.apache.http.HttpStatus;
 import org.broadinstitute.dsde.workbench.client.sam.ApiClient;
@@ -96,11 +97,12 @@ public class SamService {
 
   /**
    * Call the SAM "/api/users/v1/invite/{inviteeEmail}" endpoint to invite a user and track them.
-   * This is not the same thing as registering a user.
+   * This is not the same thing as registering a user. Unlike other methods in this class, this
+   * method does not include retries.
    *
    * @param userEmail email to invite
    */
-  public void inviteUser(String userEmail) throws ApiException {
+  public void inviteUserNoRetries(String userEmail) throws ApiException {
     logger.info("Inviting new user: {}", userEmail);
     UserStatusDetails userStatusDetails = new UsersApi(apiClient).inviteUser(userEmail);
     logger.info("Invited new user: {}", userStatusDetails);
@@ -290,7 +292,7 @@ public class SamService {
           SamService::isRetryable,
           SamService::isBadRequest,
           () -> {
-            inviteUser(userEmail);
+            inviteUserNoRetries(userEmail);
             return null;
           },
           SamService::isRetryable);
@@ -370,7 +372,7 @@ public class SamService {
           SamService::isRetryable,
           SamService::isBadRequest,
           () -> {
-            inviteUser(userEmail);
+            inviteUserNoRetries(userEmail);
             return null;
           },
           SamService::isRetryable);
@@ -533,7 +535,9 @@ public class SamService {
    * @return true if the exception is retryable
    */
   static boolean isRetryable(Exception ex) {
-    if (!(ex instanceof ApiException)) {
+    if (ex instanceof SocketTimeoutException) {
+      return true;
+    } else if (!(ex instanceof ApiException)) {
       return false;
     }
     int statusCode = ((ApiException) ex).getCode();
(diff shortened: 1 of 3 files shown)


src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
```diff
@@ -332,7 +351,8 @@ public void grantIamRole(UUID workspaceId, String userEmail, IamRole iamRole) {
           () -> {
             new SamService(server, user).inviteUser(userEmail);
             return null;
-          });
+          },
+          (ex) -> isRetryable(ex) || SamService.isRetryable(ex));
```
wchamber, line 355: Isn't the SamService already doing retries internally for `inviteUser`, so why do we need to expose and use SamService.isRetryable?

src/main/java/bio/terra/cli/service/utils/HttpUtils.java
```diff
@@ -294,12 +296,13 @@ public static HttpResponse sendHttpRequest(
       logger.info("Caught possible one-time error: {}", ex);
 
       // handle the one-time error
-      handleOneTimeError.makeRequest();
+      callWithRetries(handleOneTimeError, isRetryable);
```
wchamber, line 299: I don't think we should conflate what's retryable for handleOneTimeError and makeRequest. Arguably, this function shouldn't do retries at all and the retries should be handled in the `SupplierWithCheckedException` if necessary.

marikomedlock: Hmm I'm conflicted on this one. On the one hand, I agree the `handleOneTimeError` pattern is distinct from retries, but everywhere we're using this pattern in the codebase we also need retries. Polling was a similar case, hence the `pollWithRetries` function. So I think it makes sense to keep the retries together with the pattern, until we have a case that needs them separated.

Good point about separating `isRetryable` for the initial request and the one-time error handling functions. Will take a pass at that.

src/main/java/bio/terra/cli/service/utils/HttpUtils.java
```diff
@@ -271,23 +271,28 @@ public static HttpResponse sendHttpRequest(
...
    * @throws E2 if handleOneTimeError throws an exception
    */
-  public static <T, E1 extends Exception, E2 extends Exception> T callAndHandleOneTimeError(
-      SupplierWithCheckedException<T, E1> makeRequest,
-      Predicate<Exception> isOneTimeError,
-      SupplierWithCheckedException<Void, E2> handleOneTimeError,
-      Predicate<Exception> isRetryable)
-      throws E1, E2, InterruptedException {
+  public static <T, E1 extends Exception, E2 extends Exception>
+      T callAndHandleOneTimeErrorWithRetries(
```
jaycarlton, line 286: Since you frequently call this with the same predicates like `samService::isRetryable`, you could make shorter overloads and use them when the defaults apply.

jaycarlton (approved):

src/main/java/bio/terra/cli/service/SamService.java
```diff
@@ -98,22 +99,11 @@ public SystemStatus getStatus() {
...
-      logger.info("Invited new user: {}", userStatusDetails);
-      return userStatusDetails;
-    } catch (ApiException | InterruptedException ex) {
-      throw new SystemException("Error inviting user in SAM.", ex);
-    } finally {
-      closeConnectionPool();
-    }
+  public void inviteUser(String userEmail) throws ApiException {
+    logger.info("Inviting new user: {}", userEmail);
+    UserStatusDetails userStatusDetails = new UsersApi(apiClient).inviteUser(userEmail);
```
wchamber, line 105: up to you: It's still a odd to me that this is the only SamService function that doesn't do its own retrying. I don't think the fact that this is frequently (only?) used with handleOneTimeError to be a compelling reason for the assymetry. I think it'd be preferablle to not retry the SamService.inviteUser (`x -> false`) when it's used in handleOneTimeError and rely on SamService to do the retry. I think then we also don't need to expose SamService::isRetryable and let Sam retry logic be totally encapsulated by the SamService.

wchamber (approved):
thanks!

src/main/java/bio/terra/cli/service/utils/HttpUtils.java
marikomedlock: Good idea. Added an overload just for `SamService` since all 3 occurrences there are using the same `isRetryable` method.

src/main/java/bio/terra/cli/service/SamService.java
marikomedlock: I see your point. I slightly prefer not nesting retries, because I think it makes the code less readable. Going to leave as is for now, but open to changing later if this gets too verbose.

PR #94, opened 2021-06-10
[no-ticket] Move post startup script debug output to user directory

wchamber:
Change from piping outputs to /tmp/ so that they aren't as easily deleted.

### DIFF
--- a/notebooks/post-startup.sh
+++ b/notebooks/post-startup.sh
@@ -8,9 +8,14 @@ set -o nounset
 set -o pipefail
 set -o xtrace
 
-# Send stdout and stderr from this script to a tmp file for debugging.
+# The linux user that JupyterLab will be running as. It's important to do some parts of setup in the
+# user space.
+readonly JUPYTER_USER="jupyter"
+
 cd /tmp || exit
-exec >> /tmp/post-startup-output.txt
+# Send stdout and stderr from this script to a file for debugging.
+mkdir /home/"${JUPYTER_USER}"/.terra-debug
+exec >> /home/"${JUPYTER_USER}"/.terra-debug/post-startup-output.txt
 exec 2>&1
 
 #######################################
--- a/notebooks/post-startup.sh
+++ b/notebooks/post-startup.sh
@@ -9,13 +9,16 @@ set -o pipefail
 set -o xtrace
 
 # The linux user that JupyterLab will be running as. It's important to do some parts of setup in the
-# user space.
+# user space, such as setting Terra CLI settings which are persisted in the user's $HOME.
+# This post startup script is not run by the same user.
 readonly JUPYTER_USER="jupyter"
 
+# Move to the /tmp directory to let any artifacts left behind by this script can be removed.
 cd /tmp || exit
+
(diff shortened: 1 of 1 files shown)


notebooks/post-startup.sh
```diff
@@ -8,9 +8,14 @@ set -o nounset
...
 
-# Send stdout and stderr from this script to a tmp file for debugging.
+# The linux user that JupyterLab will be running as. It's important to do some parts of setup in the
+# user space.
+readonly JUPYTER_USER="jupyter"
+
 cd /tmp || exit
-exec >> /tmp/post-startup-output.txt
+# Send stdout and stderr from this script to a file for debugging.
+mkdir /home/"${JUPYTER_USER}"/.terra-debug
```
marikomedlock, line 17: Do you want `mkdir -p` here in case the user's home directory doesn't exist already?

notebooks/post-startup.sh
```diff
@@ -8,9 +8,14 @@ set -o nounset
 set -o pipefail
 set -o xtrace
 
-# Send stdout and stderr from this script to a tmp file for debugging.
+# The linux user that JupyterLab will be running as. It's important to do some parts of setup in the
+# user space.
+readonly JUPYTER_USER="jupyter"
+
 cd /tmp || exit
```
marikomedlock, line 15: I think you don't need to be in the `/tmp` directory anymore.

notebooks/post-startup.sh
```diff
@@ -8,9 +8,14 @@ set -o nounset
...
-exec >> /tmp/post-startup-output.txt
+# Send stdout and stderr from this script to a file for debugging.
+mkdir /home/"${JUPYTER_USER}"/.terra-debug
+exec >> /home/"${JUPYTER_USER}"/.terra-debug/post-startup-output.txt
```
marikomedlock, line 18: Perhaps you could put this debug output under `$HOME/.terra`. The CLI will put logs in `$HOME/.terra/logs/` and it would be nice to have this consistent. e.g. So we can tell a user with a problem, send me the `$HOME/.terra/....` directory for debugging.

notebooks/post-startup.sh
```diff
@@ -8,9 +8,14 @@ set -o nounset
 set -o pipefail
 set -o xtrace
 
-# Send stdout and stderr from this script to a tmp file for debugging.
+# The linux user that JupyterLab will be running as. It's important to do some parts of setup in the
+# user space.
```
marikomedlock, line 12: nit: Why is that important? A brief explanation here would be helpful.

notebooks/post-startup.sh
wchamber: done

notebooks/post-startup.sh
wchamber: I think it's nice for the terra tar & other stuff from install to get cleaned up eventually, added a comment.

notebooks/post-startup.sh
wchamber: thanks, added more, PTAL

marikomedlock (approved):

PR #98, opened 2021-06-15
PF-796: Workspace config command. Config tests.

marikomedlock:
- Added workspace as a config property.
  - `terra config set workspace` is identical to the existing `terra workspace set`. This is just for convenience, similar to how there are identical `terra config set server` and `terra server set`.
  - `terra config get-value workspace` returns the current workspace id.
  - The current workspace id is now included in the output of `terra config list`.
- Added tests for the `config` group of commands.
- Fixed a bug where the `gcloud` project config property was not being restored if it was previously unset. This happened when executing apps in the LOCAL_PROCESS mode only.

### DIFF
--- a/src/main/java/bio/terra/cli/serialization/userfacing/UFConfig.java
+++ b/src/main/java/bio/terra/cli/serialization/userfacing/UFConfig.java
@@ -74,7 +74,7 @@ public class UFConfig {
             + fileLoggingLevel);
     OUT.println();
     OUT.println("[server] server = " + serverName);
-    OUT.println("[workspace] workspace = " + (workspaceId == null ? "" : workspaceId));
+    OUT.println("[workspace] workspace = " + (workspaceId == null ? "(unset)" : workspaceId));
   }
 
   @JsonPOJOBuilder(buildMethodName = "build", withPrefix = "")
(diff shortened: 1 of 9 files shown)


src/main/java/bio/terra/cli/serialization/userfacing/UFConfig.java
```diff
@@ -67,6 +74,7 @@ public void print() {
             + fileLoggingLevel);
     OUT.println();
     OUT.println("[server] server = " + serverName);
```
jmczerk, line 76: Maybe skip the whole line if `workspaceId == null`?  I feel like this would look strange as written.

jmczerk (approved):

src/main/java/bio/terra/cli/serialization/userfacing/UFConfig.java
marikomedlock: How about `[workspace] workspace = (unset)` instead? Agreed about it looking weird with just a dangling `=`. But I think it's helpful to have all the possible config properties listed out.

jmczerk: Works for me, thanks.

wchamber (approved):
Sorry for the slow turnaround, I must have missed the notification.

PR #102, opened 2021-06-17
PF-797: Tests for `group` and `spend` commands.

marikomedlock:
- Tests for the `group` commands, which manage SAM groups.
- Tests for the `spend` commands, which manage access to the WSM default spend profile.
- For testing the `spend` commands, I didn't want to use any of the test users, because that could cause problems if tests get run simultaneously. Instead, I generate a SAM group, use its accompanying email in the tests, then delete the group. This is just a workaround until we have a way to generate test users and clean them up in SAM afterwards.

### DIFF
--- a/src/test/java/harness/TestUsers.java
+++ b/src/test/java/harness/TestUsers.java
@@ -41,6 +41,9 @@ public enum TestUsers {
   public final String email;
   public final SpendEnabled spendEnabled;
 
+  // name of the group that includes CLI test users and has spend profile access
+  public static final String CLI_TEST_USERS_GROUP_NAME = "cli-test-users";
+
   TestUsers(String email, SpendEnabled spendEnabled) {
     this.email = email;
     this.spendEnabled = spendEnabled;
@@ -132,10 +135,18 @@ public enum TestUsers {
     return chooseTestUser(Set.of(SpendEnabled.values()));
   }
 
-  /** Randomly chooses a test user with spend profile access. */
+  /** Randomly chooses a test user with spend profile access, but without owner privileges. */
   public static TestUsers chooseTestUserWithSpendAccess() {
     return chooseTestUser(
         Set.of(new SpendEnabled[] {SpendEnabled.CLI_TEST_USERS_GROUP, SpendEnabled.DIRECTLY}));
+  }
+
+  /**
+   * Randomly chooses a test user who is a spend profile admin and an admin of the SAM cli-testers
+   * group.
+   */
+  public static TestUsers chooseTestUserWithOwnerAccess() {
+    return chooseTestUser(Set.of(SpendEnabled.OWNER));
   }
 
   /** Randomly chooses a test user without spend profile access. */
--- a/src/test/java/harness/utils/SamGroups.java
+++ b/src/test/java/harness/utils/SamGroups.java
@@ -1,0 +1,46 @@
+package harness.utils;
+
+import harness.TestCommand;
+import harness.TestUsers;
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Random;
+
+/**
+ * Utility class for tracking groups created by a test class, and then trying to delete them all at
+ * the end.
+ */
+public class SamGroups {
+  // keep a map of groups created by tests, so we can try to clean them up
+  // group name -> test user that created it
+  private Map<String, TestUsers> trackedGroups = new HashMap<>();
+
+  private static final Random RANDOM = new Random();
+
+  /** Try to delete each group that was tracked here. */
+  public void deleteAllTrackedGroups() throws IOException {
+    for (Map.Entry<String, TestUsers> groupCreated : trackedGroups.entrySet()) {
+      groupCreated.getValue().login();
+      TestCommand.Result cmd =
+          TestCommand.runCommand("groups", "delete", "--name=" + groupCreated.getKey());
+      if (cmd.exitCode == 0) {
+        System.out.println("group was not cleaned up by test: " + groupCreated.getKey());
+      }
+    }
+  }
+
+  /** Add a group to be tracked. */
+  public void trackGroup(String name, TestUsers admin) {
+    trackedGroups.put(name, admin);
+  }
+
+  /** Helper method to generate a random group name. */
+  public static String randomGroupName() {
+    return "terraCLI_" + RANDOM.nextInt(Integer.MAX_VALUE);
+  }
+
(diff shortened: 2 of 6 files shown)


src/test/java/harness/TestUsers.java
```diff
@@ -41,6 +41,9 @@
   public final String email;
   public final SpendEnabled spendEnabled;
 
+  // name of the group that includes CLI test users and has spend profile access
+  public static final String CLI_TEST_USERS_GROUP_NAME = "cli-test-users";
```
wchamber, line 45: Maybe good to say if this was set up once manually? Will the tests using this fail on non dev Sam?

src/test/java/harness/utils/SamGroups.java
```diff
@@ -0,0 +1,46 @@
...
+  private static final Random RANDOM = new Random();
+
+  /** Try to delete each group that was tracked here. */
+  public void deleteAllTrackedGroups() throws IOException {
+    for (Map.Entry<String, TestUsers> groupCreated : trackedGroups.entrySet()) {
+      groupCreated.getValue().login();
+      TestCommand.Result cmd =
+          TestCommand.runCommand("groups", "delete", "--name=" + groupCreated.getKey());
+      if (cmd.exitCode == 0) {
+        System.out.println("group was not cleaned up by test: " + groupCreated.getKey());
```
wchamber, line 28: is this going to log if the group was previously deleted? maybe worth a comment either way.

src/test/java/unit/Group.java
```diff
@@ -0,0 +1,318 @@
...
+    // check that the group creator is included in the list-users output
+    expectListedMemberWithPolicies(name, groupCreator.email, GroupPolicy.ADMIN);
+
+    // `terra groups delete --name=$name`
+    TestCommand.runCommandExpectSuccess("groups", "delete", "--name=" + name);
+  }
+
+  @Test
+  @DisplayName("cli-testers group includes the appropriate test users")
+  void cliTestersGroupMembership() throws IOException {
```
wchamber, line 262: maybe worth a comment that this isn't about testing the CLI so much as testing the environment setup?

src/test/java/unit/SpendProfileUser.java
```diff
@@ -0,0 +1,197 @@
...
+  // only an owner on the spend profile can disable emails
+  TestUsers spendProfileOwner = TestUsers.chooseTestUserWithOwnerAccess();
+
+  SamGroups trackedGroups = new SamGroups();
+
+  @AfterAll
+  void cleanupOnce() throws IOException, InterruptedException {
+    spendProfileOwner.login();
+
+    // try to disable spend for each group that was created by a method in this class
```
wchamber, line 38: why do we need to do this if we're going to try to delete the groups?

src/test/java/unit/SpendProfileUser.java
```diff
@@ -0,0 +1,197 @@
...
+    // check that the test email is not included in the list-users output
+    Optional<UFSpendProfileUser> listUsersOutput = listUsersWithEmail(testEmail);
+    assertTrue(
+        listUsersOutput.isEmpty(),
+        "user is not included in the test-users output after having their spend access disabled");
+  }
+
+  @Test
+  @DisplayName("only users who are an admin on the spend profile can enable and disable others")
+  void onlyAdminCanEnableDisable() throws IOException {
```
wchamber, line 81: Is this testing CLI functionality or Sam functionality?

src/test/java/unit/SpendProfileUser.java
```diff
@@ -0,0 +1,197 @@
...
+
+    // use a test user that is an owner to remove spend access for the test email
+    spendProfileOwner.login();
+    TestCommand.runCommandExpectSuccess(
+        "spend", "disable", "--email=" + testEmail, "--policy=USER");
+  }
+
+  @Test
+  @DisplayName("spend profile includes cli-testers group and the appropriate test users")
+  void testUsersSpendProfileMembership() throws IOException {
```
wchamber, line 122: ditto comment about testing environment rather than CLI functionality

wchamber (approved):

src/test/java/harness/TestUsers.java
jmczerk: Maybe some elaboration on this in the `CONTRIBUTING.md` "Testing" section? (not nesc this PR)

jmczerk (approved):

src/test/java/harness/TestUsers.java
marikomedlock: RE documentation:
- There's a comment at the top of the class: "These permissions were configured manually, and are not part of the CLI test harness. See CONTRIBUTING.md for more details about the manual setup.".
- `CONTRIBUTING.md` has a [section](https://github.com/DataBiosphere/terra-cli/blob/main/CONTRIBUTING.md#setup-test-users) "Setup test users". I do need to update the command option names there though.

RE tests on non dev SAM:
- Yes, these tests will fail on any SAM instance that hasn't been setup with these test users and group.
- Setting up a new instance should just require re-running the same script. However, if we want to use test users in a different domain (i.e. not `test.firecloud.org`), that would require some refactoring.

src/test/java/harness/utils/SamGroups.java
marikomedlock: It will only log here if the group was not previously deleted. Added a comment.

src/test/java/unit/Group.java
marikomedlock: added

src/test/java/unit/SpendProfileUser.java
marikomedlock: Because if you don't remove spend access first, the group deletion fails.

src/test/java/unit/SpendProfileUser.java
marikomedlock: I guess it's hard to separate the two entirely. I wrote this test thinking about whether the CLI test users were setup correctly. For example, if a user is an owner on the spend profile, when they're supposed to be a user, then the tests that expect spend profile access would still work, but we're not testing what we think we are. I thought this was important to test, because I had trouble with this when I tested briefly against an early Verily SAM (i.e. users that were direct owners on the spend profile worked, but those that were users or members of a group that was a user did not).

src/test/java/unit/SpendProfileUser.java
marikomedlock: added

PR #103, opened 2021-06-18
PF-831: Tests for GCS bucket lifecycle rules.

marikomedlock:
- Tests for setting lifecycle rules on controlled GCS bucket resources.
- Updated examples in README.


src/test/java/harness/utils/ExternalGCSBuckets.java
```diff
@@ -49,18 +50,35 @@ public static Bucket createBucket() throws IOException {
...
+        bucket,
+        Identity.user(email),
+        // TODO (PF-717): revisit this once we're calling WSM endpoints for check-access
+        StorageRoles.objectViewer(),
+        StorageRoles.legacyBucketReader());
+  }
+
+  /**
+   * Grant a given user admin access to a bucket. This method uses SA credentials for an external
+   * project.
```
wchamber, line 63: nit: Do we ever not want to use SA credentials for an external project? I'd think we could use the same methods whether or not the SA is in the same project as the bucket.

src/test/java/unit/GcsBucketLifecycle.java
```diff
@@ -0,0 +1,390 @@
...
+@Tag("unit")
+public class GcsBucketLifecycle extends SingleWorkspaceUnit {
+  // external bucket to use for testing the JSON format against GCS directly
+  private Bucket externalBucket;
+
+  @Override
+  @BeforeAll
+  protected void setupOnce() throws IOException {
+    super.setupOnce();
+    externalBucket = ExternalGCSBuckets.createBucket();
```
wchamber, line 41: add a todo for using CRL + Janitor?

src/test/java/unit/GcsBucketLifecycle.java
```diff
@@ -0,0 +1,390 @@
...
+
+    // `terra auth status --format=json`
+    UFAuthStatus authStatus =
+        TestCommand.runAndParseCommandExpectSuccess(UFAuthStatus.class, "auth", "status");
+
+    // grant the user's proxy group write access to the bucket, so we can test calling `terra gsutil
+    // lifecycle` with the same JSON format used for creating controlled bucket resources with
+    // lifecycle rules
+    ExternalGCSBuckets.grantWriteAccess(externalBucket, Identity.group(authStatus.proxyGroupEmail));
```
wchamber, line 50: nit: since this is only used once in a single test, maybe just do all of this setup/tear down in that test?

src/test/java/unit/GcsBucketLifecycle.java
```diff
@@ -0,0 +1,390 @@
...
+    BucketInfo.LifecycleRule lifecycleRuleFromGCS =
+        createBucketWithOneLifecycleRule(name, name + ".json");
+
+    expectActionDelete(lifecycleRuleFromGCS);
+    assertEquals(365, lifecycleRuleFromGCS.getCondition().getAge(), "condition age matches");
+  }
+
+  @Test
+  @DisplayName("lifecycle action set storage class (condition age)")
+  void setStorageClassAction() throws IOException {
```
wchamber, line 85: nit: It feels like overkill to create a new bucket for every different condition/action. I think this would be fine to have as a unit test on `WorkspaceManagerService#fromCLIObject` that we can translate a CLI input format -> WM API format, but going end-to-end for every variation seems like overkill where we're testing WM functionality more than the CLIs. If we're using the WM API, we don't need to verify that every one of its options does what's promised, one end2end on the lifecycle seems sufficient. Or, why don't we need to verify every condition structure also works as a gsutil format?

wchamber (approved):

src/test/java/unit/GcsBucketLifecycle.java
```diff
@@ -0,0 +1,390 @@
...
+        ruleWithSetStorageClassAction.isPresent(), "one rule has action type = set storage class");
+    expectActionSetStorageClass(ruleWithSetStorageClassAction.get(), StorageClass.COLDLINE);
+    assertFalse(
+        ruleWithSetStorageClassAction.get().getCondition().getIsLive(),
+        "condition is live matches");
+  }
+
+  /** Check that the action is Delete. */
+  private void expectActionDelete(BucketInfo.LifecycleRule rule) {
+    assertEquals("Delete", rule.getAction().getActionType(), "Delete action type matches");
```
jaycarlton, line 315: You can use `DeleteLifecycleAction.TYPE` or some such for this.

jaycarlton (approved):
very thorough. thanks for the coverage

src/test/java/harness/utils/ExternalGCSBuckets.java
marikomedlock: Right, the SA doesn't need to be in the same project as the bucket. But the bucket needs to be in an external (to WSM) project in order for us to be able to set IAM policy on it. Practically, the SA will always be in the same project as the bucket, just because the CLI tests use a single project for testing referenced resources, but that's not required. Maybe this is just poor wording on my part, but if I've misunderstood your question, please restate.

src/test/java/unit/GcsBucketLifecycle.java
marikomedlock: It's noted in the ticket.

src/test/java/unit/GcsBucketLifecycle.java
marikomedlock: I want the teardown to happen regardless of whether the test passes though. I could put just the setup in the test method, and then do the teardown in the `@AfterAll` method -- that's what I do for SAM groups, for example. But since we only need one, it seemed more readable to put setup/cleanup in `@BeforeAll` and `@AfterAll`. I think if we get a test that creates multiple external buckets, then following the pattern of `SamGroups`, which tracks groups and then provides a method for deleting them all at the end, would be good.

src/test/java/unit/GcsBucketLifecycle.java
marikomedlock: I think the CLI test should cover deserializing the lifecycle JSON file into a CLI object and converting that into a WSM object, since those are the parts of the e2e that the CLI owns. The deserializing happens in the command class, so these tests should go through there. From the perspective of testing the CLI, I would be happy with a test that stopped just before making the WSM call, but don't see how to do that without modifying the codebase -- it's simpler to go e2e.

The goal of testing each option wasn't to test WSM (though that did end up being useful -- this test showed that 4 options in WSM's public API are actually not supported yet). It was to test the CLI deserialization and conversion into a WSM object.

I included a single test for `gsutil` just to check that the CLI JSON format was readable by GCS directly. The other tests already covered that the CLI deserialization worked.

src/test/java/harness/utils/ExternalGCSBuckets.java
```diff
@@ -49,18 +50,35 @@ public static Bucket createBucket() throws IOException {
...
-                // TODO (PF-717): revisit this once we're calling WSM endpoints for check-access
-                .addIdentity(StorageRoles.objectViewer(), Identity.user(email))
-                .addIdentity(StorageRoles.legacyBucketReader(), Identity.user(email))
-                .build());
-    getStorageClient().setIamPolicy(bucket.getName(), updatedPolicy);
+    Policy.Builder updatedPolicyBuilder = currentPolicy.toBuilder();
+    for (Role role : roles) {
+      updatedPolicyBuilder.addIdentity(role, user);
+    }
+    Policy updatedPolicy = storage.setIamPolicy(bucket.getName(), updatedPolicyBuilder.build());
```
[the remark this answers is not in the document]
marikomedlock: Good catch. Remove double call.

src/test/java/unit/GcsBucketLifecycle.java
```diff
@@ -0,0 +1,390 @@
...
+
+  @Override
+  @BeforeEach
+  protected void setupEachTime() throws IOException {
+    super.setupEachTime();
+
+    workspaceCreator.login();
+
+    // `terra workspace set --id=$id`
+    TestCommand.runCommandExpectSuccess("workspace", "set", "--id=" + getWorkspaceId());
```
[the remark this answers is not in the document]
marikomedlock: The `login` just sets up credentials for a test user. There are other tests that either login a user that is not the workspace creator, or that don't set a workspace. I note that just to say these calls are not always bundled. For this test they are always done together though, and putting them in the `@BeforeEach` method was to avoid repetition.

src/test/java/unit/GcsBucketLifecycle.java
```diff
@@ -0,0 +1,390 @@
...
+  @DisplayName("lifecycle condition noncurrent time before (action set storage class)")
+  void noncurrentTimeBeforeCondition() throws IOException {
+    String name = "setStorageClass_noncurrentTimeBefore";
+    BucketInfo.LifecycleRule lifecycleRuleFromGCS =
+        createBucketWithOneLifecycleRule(name, name + ".json");
+
+    expectActionSetStorageClass(lifecycleRuleFromGCS, StorageClass.NEARLINE);
+    assertEquals(
+        new DateTime("2014-08-28"),
+        lifecycleRuleFromGCS.getCondition().getNoncurrentTimeBefore(),
```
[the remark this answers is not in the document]
marikomedlock: noncurrent

src/test/java/unit/GcsBucketLifecycle.java
```diff
@@ -0,0 +1,390 @@
...
+        lifecycleRuleFromGCS.getCondition().getNumberOfNewerVersions(),
+        "condition number of newer versions matches");
+  }
+
+  @Test
+  @DisplayName("auto-delete option")
+  void autoDeleteOption() throws IOException {
+    String resourceName = "autodelete";
+    String bucketName = UUID.randomUUID().toString();
+    int autoDelete = 24;
```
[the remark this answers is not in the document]
marikomedlock: changed

src/test/java/unit/GcsBucketLifecycle.java
```diff
@@ -0,0 +1,390 @@
...
+    validateMultipleRules(lifecycleRules);
+  }
+
+  @Test
+  @DisplayName("CLI uses the same format as gsutil for setting lifecycle rules")
+  void sameFormatForExternalBucket() throws IOException {
+    // the CLI mounts the current working directory to the Docker container when running apps
+    // so we need to give it the path to lifecycle JSON file relative to the current working
+    // directory
+    Path lifecyclePathOnHost = TestCommand.getPathForTestInput("gcslifecycle/multipleRules.json");
```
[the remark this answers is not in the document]
marikomedlock: added

src/test/java/unit/GcsBucketLifecycle.java
```diff
@@ -0,0 +1,390 @@
...
+    Path currentDirOnHost = Path.of(System.getProperty("user.dir"));
+    Path lifecycelPathOnContainer = currentDirOnHost.relativize(lifecyclePathOnHost);
+
+    // `terra gsutil lifecycle set $lifecycle gs://$bucketname`
+    TestCommand.runCommandExpectSuccess(
+        "gsutil",
+        "lifecycle",
+        "set",
+        lifecycelPathOnContainer.toString(),
+        "gs://" + externalBucket.getName());
```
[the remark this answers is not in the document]
marikomedlock: Good idea. added.

src/test/java/unit/GcsBucketLifecycle.java
```diff
@@ -0,0 +1,390 @@
...
+
+  @Test
+  @DisplayName("CLI uses the same format as gsutil for setting lifecycle rules")
+  void sameFormatForExternalBucket() throws IOException {
+    // the CLI mounts the current working directory to the Docker container when running apps
+    // so we need to give it the path to lifecycle JSON file relative to the current working
+    // directory
+    Path lifecyclePathOnHost = TestCommand.getPathForTestInput("gcslifecycle/multipleRules.json");
+    Path currentDirOnHost = Path.of(System.getProperty("user.dir"));
+    Path lifecycelPathOnContainer = currentDirOnHost.relativize(lifecyclePathOnHost);
```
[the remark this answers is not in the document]
marikomedlock: done

src/test/java/unit/GcsBucketLifecycle.java
marikomedlock: done.

PR #104, opened 2021-06-18
PF-717: Call WSM for `check-access` command.

marikomedlock:
- Replaced the CLI implementation of `check-access` for referenced resources with a call to the [new WSM endpoint](https://github.com/DataBiosphere/terra-workspace-manager/pull/369). Since the new endpoint is resource type-agnostic, we also don't need resource-type specific implementations of the access check any more.
- Refactored the constructors of the `*Service` classes to use more descriptive factory methods and be more consistent with each other.


zloery (approved):
LGTM

jmczerk (approved):

PR #105, opened 2021-06-21
PF-862, PF-870: Tests for notebooks commands.

marikomedlock:
- Added tests for AI notebooks
  - `resources create/delete/list/describe/resolve/check-access` are very similar to the tests for other controlled resources
  - `notebooks start/stop` are convenience wrappers around `gcloud notebooks instances start/stop`.
- Fixed a bug where the instance lookup fails after deleting a notebook.


src/test/java/unit/AiNotebookControlled.java
```diff
@@ -0,0 +1,312 @@
...
+        "delete either succeeds or times out",
+        cmd.exitCode == 0
+            || (cmd.exitCode == 1
+                && cmd.stdErr.contains(
+                    "CLI timed out waiting for the job to complete. It's still running on the server.")));
+
+    // confirm it no longer appears in the resources list
+    List<UFAiNotebook> listedNotebooks = listNotebookResourcesWithName(name);
+    assertEquals(
+        0, listedNotebooks.size(), "deleted notebook no longer appears in the resources list");
```
wchamber, line 88: nit: `assertThat( "...", listedNotebooks, Matchers.empty());` is a little more idiomatic and will show the contents of listedNotebooks if it's not empty.

src/test/java/unit/AiNotebookControlled.java
```diff
@@ -0,0 +1,312 @@
...
+    assertEquals(location, createdNotebook.location, "create output matches location");
+    assertEquals(instanceId, createdNotebook.instanceId, "create output matches instance id");
+
+    // ai notebooks are always private
+    assertEquals(
+        AccessScope.PRIVATE_ACCESS, createdNotebook.accessScope, "create output matches access");
+    assertEquals(
+        workspaceCreator.email.toLowerCase(),
+        createdNotebook.privateUserName.toLowerCase(),
+        "create output matches private user name");
```
wchamber, line 161: optional: Unless there's a particular reason to think that changing the location and instance id will change the other fields, we could probably skip re-asserting in this test that the name, description, cloning, access, & email match and teset that once in the other test. ditto for description matches.

src/test/java/unit/AiNotebookControlled.java
```diff
@@ -0,0 +1,312 @@
...
+    // TODO (PF-869): change this to expect success once polling notebook operations is fixed
+    TestCommand.runCommand("notebooks", "stop", "--name=" + name);
+    pollDescribeForNotebookState(name, "STOPPED");
+  }
+
+  /**
+   * Helper method to poll `terra resources describe` until the notebook state equals that
+   * specified. Uses the current workspace.
+   */
+  public static void pollDescribeForNotebookState(String resourceName, String notebookState)
```
wchamber, line 215: If we're going to reuse these in multiple tests, make a shared notebook test utilities file.

src/test/java/unit/AiNotebookControlled.java
```diff
@@ -0,0 +1,312 @@
...
+      throws InterruptedException, JsonProcessingException {
+    pollDescribeForNotebookState(resourceName, notebookState, null);
+  }
+
+  /**
+   * Helper method to poll `terra resources describe` until the notebook state equals that
+   * specified. Filters on the specified workspace id; Uses the current workspace if null.
+   */
+  public static void pollDescribeForNotebookState(
```
wchamber, line 224: Why poll through a CLI command instead of using `GoogleAiNotebooks.get`? Would be less verbose and more to the point.

src/test/java/unit/AiNotebookControlled.java
```diff
@@ -0,0 +1,312 @@
...
+                UFAiNotebook.class, "resources", "describe", "--name=" + resourceName)
+            : TestCommand.runAndParseCommandExpectSuccess(
+                UFAiNotebook.class,
+                "resources",
+                "describe",
+                "--name=" + resourceName,
+                "--workspace=" + workspaceId);
+    assertEquals(notebookState, describeNotebook.state, "notebook state matches");
+    if (!notebookState.equals("PROVISIONING")) {
+      assertNotNull(describeNotebook.proxyUri, "proxy url is populated");
```
wchamber, line 255: consider separating the act of polling from a state from asserting what the described notebook should look like. I'm kind of surprised that "STOPPED" has a proxyUri.

src/test/java/unit/AiNotebookControlled.java
```diff
@@ -0,0 +1,312 @@
...
+    }
+  }
+
+  /**
+   * Helper method to call `terra resources list` and expect one resource with this name. Uses the
+   * current workspace.
+   */
+  static UFAiNotebook listOneNotebookResourceWithName(String resourceName)
```
wchamber, line 263: ditto utility file

src/test/java/unit/WorkspaceOverride.java
```diff
@@ -390,4 +395,39 @@ void workspace() throws IOException {
     matchingWorkspaces = listWorkspacesWithId(workspace1.id);
     assertEquals(1, matchingWorkspaces.size(), "workspace 1 is still included in list");
   }
+
+  @Disabled // TODO (PF-869): enable this test once polling notebook operations is fixed
+  @DisplayName("notebook commands respect workspace override")
```
wchamber, line 400: optional: seems kind of like overkill to have to include every workspace id command in this test file. Consider testing that the workspace id overrides alongside the other tests for the same command. Its's still a test of the notebooks commands, rather than a test of the mixin.

wchamber (approved):

src/test/java/unit/GcsBucketControlled.java
```diff
@@ -35,7 +35,7 @@ void listDescribeReflectCreate() throws IOException {
     // `terra workspace set --id=$id`
```
jmczerk, line 35: Out of interest, why drop `--format=json` here?

jmczerk (approved):

src/main/java/bio/terra/cli/serialization/userfacing/resources/UFAiNotebook.java
```diff
@@ -32,11 +33,11 @@ public UFAiNotebook(AiNotebook internalObj) {
     this.instanceId = internalObj.getInstanceId();
     this.location = internalObj.getLocation();
 
-    Instance instance = internalObj.getInstance();
-    this.instanceName = instance.getName();
-    this.state = instance.getState();
-    this.proxyUri = instance.getProxyUri();
-    this.createTime = instance.getCreateTime();
+    Optional<Instance> instance = internalObj.getInstance();
+    this.instanceName = instance.isPresent() ? instance.get().getName() : "";
```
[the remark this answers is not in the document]
marikomedlock: Moved the `null` -> empty string into `print()`.

src/main/java/bio/terra/cli/businessobject/resources/AiNotebook.java
```diff
@@ -118,15 +120,20 @@ public String resolve() {
...
             .location(location)
             .instanceId(instanceId)
             .build();
     GoogleAiNotebooks notebooks = new GoogleAiNotebooks(Context.requireUser().getUserCredentials());
-    return notebooks.get(instanceName);
+    try {
+      return Optional.of(notebooks.get(instanceName));
+    } catch (SystemException ex) {
+      logger.error("Caught exception looking up notebook instance", ex);
+      return Optional.empty();
```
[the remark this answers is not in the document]
marikomedlock: `getInstance` should succeed regardless of whether `notebooks.get` does. Instead of changing `notebooks.get` to handle a specific exception, I'll change `getInstance` to handle all exceptions.

src/test/java/unit/AiNotebookControlled.java
```diff
@@ -0,0 +1,312 @@
...
+/** Tests for the `terra resources` commands that handle controlled AI notebooks. */
+@Tag("unit")
+public class AiNotebookControlled extends SingleWorkspaceUnit {
+  @Test
+  @DisplayName("list and describe reflect creating and deleting a controlled notebook")
+  void listDescribeReflectCreateDelete() throws IOException {
+    workspaceCreator.login();
+
+    // `terra workspace set --id=$id`
+    TestCommand.runCommandExpectSuccess("workspace", "set", "--id=" + getWorkspaceId());
```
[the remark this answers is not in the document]
marikomedlock: I think so. There are some tests that don't do those two steps, but they're the minority and could probably just re-login.

src/test/java/unit/AiNotebookControlled.java
```diff
@@ -0,0 +1,312 @@
...
+    // TODO (PF-616): check the private user roles once WSM returns them
+
+    // `terra notebooks delete --name=$name`
+    TestCommand.Result cmd = TestCommand.runCommand("resources", "delete", "--name=" + name);
+    assertThat(
+        "delete either succeeds or times out",
+        cmd.exitCode == 0
+            || (cmd.exitCode == 1
+                && cmd.stdErr.contains(
+                    "CLI timed out waiting for the job to complete. It's still running on the server.")));
```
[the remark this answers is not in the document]
marikomedlock: Yes, confusingly the hamcrest assertions put the message first.

src/test/java/unit/AiNotebookControlled.java
marikomedlock: Good tip! I didn't know about that one.

src/test/java/unit/AiNotebookControlled.java
marikomedlock: Fair point. Going to keep them since I've already got them and they are testing CLI code, but agreed it's not the most critical part of this test.

src/test/java/unit/AiNotebookControlled.java
marikomedlock: These utility methods are notebooks-related. I think it makes sense to put them in the `AiNotebookControlled` class, where they are used most. If we end up using them in many other classes, then I'd be up for moving them into a `utils` class, but since they're only used in one other place now, I think it's unnecessary.

src/test/java/unit/AiNotebookControlled.java
marikomedlock: I was trying to avoid testing WSM functionality, and only test the CLI. That's not always possible (e.g. for bucket lifecycle rules that WSM doesn't store, I have to query the cloud directly), but where it is, I wanted to take advantage and exercise the CLI more. If we ever add tests for the other notebook creation parameters, I think we will have to query the cloud directly.

src/test/java/unit/AiNotebookControlled.java
marikomedlock: Agreed they're separate operations. But I think it makes sense to bundle them since they're always called together. I included this here because the create notebooks [tutorial](https://cloud.google.com/notebooks/docs/create-new) says to wait until the "open jupyterlab" link becomes active, which I think is equivalent to polling for the proxy url to be defined. Maybe it only makes sense to check the proxy url when it goes from provisioning -> active, but since it's a kind of status indicator, I thought it made sense to check it when polling for state changes.

src/test/java/unit/WorkspaceOverride.java
marikomedlock: You're right that it's testing the notebooks commands, instead of the mixin itself. I grouped these tests together because they have shared setup (i.e. create two workspaces to switch between and clean them up afterwards).

src/test/java/unit/GcsBucketControlled.java
marikomedlock: I only need the `--format=json` flag when I'm parsing the result (i.e. calling `TestCommand.runAndParseCommandExpectSuccess`) into an object. In the two places below, I only need to make sure the command ran successfully so I can use the default `--format=text` flag instead.

PR #107, opened 2021-06-23
PF-860, PF-868: Consolidate handling of SAM, WSM client exceptions.

marikomedlock:
- Consolidated handling of exceptions thrown by the client or retires into one place each for SAM and WSM. This removed a lot of boilerplate try/catch/wrap and re-throw logic in the methods of these classes.
  - Added logic to try and parse the `ErrorReport` returned by SAM and WSM in the response body when an error occurs.
  - Added logging so we always have the status code, response body, and exception message for SAM and WSM client exceptions, even if they're in an unexpected format or property.
  - Wrote two unit tests to make sure the SAM and WSM error message gets written out to the user.
- Overloaded some of the `HttpUtils` methods that do retries to also accept lambdas that don't return a value. This removed boilerplate `() -> { something...; return null; }` when we wanted to do retries around a function with a `void` return.
- Modified the `SupplierWithCheckedException` and `RunnableWithCheckedException` functional interfaces to also throw an `InterruptedException`. This lets us nest calls to the `HttpUtils` methods that do reties/handle one-time errors/poll.

### DIFF
--- a/src/main/java/bio/terra/cli/service/SamService.java
+++ b/src/main/java/bio/terra/cli/service/SamService.java
@@ -12,6 +12,7 @@ import com.google.auth.oauth2.AccessToken;
 import java.io.IOException;
 import java.net.SocketTimeoutException;
 import java.util.List;
+import java.util.function.Predicate;
 import javax.annotation.Nullable;
 import org.apache.http.HttpStatus;
 import org.broadinstitute.dsde.workbench.client.sam.ApiClient;
@@ -88,6 +89,22 @@ public class SamService {
   }
 
   /**
+   * Call the SAM "/register/user/v1" endpoint to register the user who is currently logged in. This
+   * is not the same as inviting a user.
+   */
+  public void registerUser() {
+    callWithRetries(
+        () -> {
+          UserStatus userStatus = new UsersApi(apiClient).createUserV2();
+          logger.info(
+              "User registered in SAM: {}, {}",
+              userStatus.getUserInfo().getUserSubjectId(),
+              userStatus.getUserInfo().getUserEmail());
+        },
+        "Error registering new user in SAM.");
+  }
+
+  /**
    * Call the SAM "/api/users/v1/invite/{inviteeEmail}" endpoint to invite a user and track them.
    * This is not the same thing as registering a user.
    *
@@ -127,21 +144,10 @@ public class SamService {
     // - try to lookup the user
     // - if the user lookup failed with a Not Found error, it means the email is not found
     // - so try to register the user first, then retry looking them up
-    UsersApi samUsersApi = new UsersApi(apiClient);
-    return handleClientExceptions(
-        () ->
-            HttpUtils.callAndHandleOneTimeErrorWithRetries(
-                samUsersApi::getUserStatusInfo,
-                SamService::isRetryable,
-                ex -> isStatusCode(ex, HttpStatusCodes.STATUS_CODE_NOT_FOUND),
-                () -> {
-                  UserStatus userStatus = samUsersApi.createUserV2();
-                  logger.info(
-                      "User registered in SAM: {}, {}",
-                      userStatus.getUserInfo().getUserSubjectId(),
-                      userStatus.getUserInfo().getUserEmail());
-                },
-                SamService::isRetryable),
+    return callAndHandleOneTimeError(
+        new UsersApi(apiClient)::getUserStatusInfo,
+        ex -> isHttpStatusCode(ex, HttpStatusCodes.STATUS_CODE_NOT_FOUND),
+        this::registerUser,
         "Error reading user information from SAM.");
   }
 
@@ -226,16 +232,10 @@ public class SamService {
     // - try to add the email to the group
     // - if this fails with a Bad Request error, it means the email is not found
     // - so try to invite the user first, then retry adding them to the group
-    handleClientExceptions(
-        () ->
-            HttpUtils.callAndHandleOneTimeErrorWithRetries(
-                () ->
-                    new GroupApi(apiClient)
-                        .addEmailToGroup(groupName, policy.getSamPolicy(), userEmail),
-                SamService::isRetryable,
-                (ex) -> isStatusCode(ex, HttpStatusCodes.STATUS_CODE_BAD_REQUEST),
-                () -> inviteUser(userEmail),
-                (ex) -> false), // don't retry because inviteUser already includes retries
+    callAndHandleOneTimeError(
+        () -> new GroupApi(apiClient).addEmailToGroup(groupName, policy.getSamPolicy(), userEmail),
+        (ex) -> isHttpStatusCode(ex, HttpStatusCodes.STATUS_CODE_BAD_REQUEST),
+        () -> inviteUser(userEmail),
         "Error adding user to SAM group.");
   }
 
@@ -284,16 +284,12 @@ public class SamService {
     // - if the add user to policy failed with a Bad Request error, it means the email is not
     // found
     // - so try to invite the user first, then retry adding them to the policy
-    handleClientExceptions(
+    callAndHandleOneTimeError(
         () ->
-            HttpUtils.callAndHandleOneTimeErrorWithRetries(
-                () ->
-                    new ResourcesApi(apiClient)
-                        .addUserToPolicy(resourceType, resourceId, resourcePolicyName, userEmail),
-                SamService::isRetryable,
-                (ex) -> isStatusCode(ex, HttpStatusCodes.STATUS_CODE_BAD_REQUEST),
-                () -> inviteUser(userEmail),
-                (ex) -> false), // don't retry because inviteUser already includes retries
+            new ResourcesApi(apiClient)
+                .addUserToPolicy(resourceType, resourceId, resourcePolicyName, userEmail),
+        (ex) -> isHttpStatusCode(ex, HttpStatusCodes.STATUS_CODE_BAD_REQUEST),
+        () -> inviteUser(userEmail),
         "Error adding user to SAM resource.");
   }
 
@@ -389,7 +385,7 @@ public class SamService {
    * @param ex exception to test
    * @return true if the exception status code matches
    */
-  private static boolean isStatusCode(Exception ex, int statusCode) {
+  private static boolean isHttpStatusCode(Exception ex, int statusCode) {
     if (!(ex instanceof ApiException)) {
       return false;
     }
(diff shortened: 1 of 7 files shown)


src/main/java/bio/terra/cli/service/SamService.java
```diff
@@ -425,20 +309,11 @@ private static boolean isNotFound(Exception ex) {
...
-          () -> {
-            resourcesApi.removeUserFromPolicy(
-                resourceType, resourceId, resourcePolicyName, userEmail);
-            return null;
-          },
-          SamService::isRetryable);
-    } catch (ApiException | InterruptedException ex) {
-      throw new SystemException("Error removing user from SAM resource.", ex);
-    } finally {
-      closeConnectionPool();
```
jaycarlton, line 440: Why don't you need to do this anymore?

jaycarlton, line 440: Why don't you need to do this anymore?

src/main/java/bio/terra/cli/service/SamService.java
```diff
@@ -530,6 +382,21 @@ private void closeConnectionPool() {
...
   }
 
+  /**
+   * Utility method that checks if an exception thrown by the SAM client matches the given HTTP
+   * status code.
+   *
+   * @param ex exception to test
+   * @return true if the exception status code matches
+   */
+  private static boolean isStatusCode(Exception ex, int statusCode) {
```
jaycarlton, line 392: maybe name it `isHttpStatusCode` so it's clear it only works on a subset of exception statuses.

src/main/java/bio/terra/cli/service/SamService.java
```diff
@@ -550,22 +417,108 @@ static boolean isRetryable(Exception ex) {
...
+   * Execute a function that includes hitting SAM endpoints. If an exception is thrown by the SAM
+   * client or the retries, make sure the HTTP status code and error message are logged.
+   *
+   * @param makeRequest function with no return value
+   * @param errorMsg error message for the the {@link SystemException} that wraps any exceptions
+   *     thrown by the SAM client or the retries
+   */
+  private void handleClientExceptions(
+      HttpUtils.RunnableWithCheckedException<ApiException> makeRequest, String errorMsg) {
+    handleClientExceptions(
```
jaycarlton, line 459: you could also overload this to take a runnable, but if this is the only place you need it, it's fine.

src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
```diff
@@ -134,28 +136,15 @@ public WorkspaceManagerService(@Nullable AccessToken accessToken, Server server)
...
-          () -> {
-            unauthenticatedApi.serviceStatus();
-            return null;
-          },
-          WorkspaceManagerService::isRetryable);
-    } catch (ApiException | InterruptedException ex) {
-      throw new SystemException("Error getting Workspace Manager status", ex);
-    }
+    callWithRetries(
+        new UnauthenticatedApi(apiClient)::serviceStatus, "Error getting Workspace Manager status");
```
jaycarlton, line 147: curious: will this create a new client for every retry, or reuse the same one?

src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
```diff
@@ -190,51 +175,51 @@ public WorkspaceDescriptionList listWorkspaces(int offset, int limit) {
...
-          createContextResult.getJobReport(), createContextResult.getErrorReport());
-
-      // call the get workspace endpoint to get the full description object
-      return HttpUtils.callWithRetries(
-          () -> workspaceApi.getWorkspace(workspaceId), WorkspaceManagerService::isRetryable);
-    } catch (ApiException | InterruptedException ex) {
-      throw new SystemException("Error creating a new workspace", ex);
-    }
+    return handleClientExceptions(
+        () -> {
```
jaycarlton, line 179: can you make these big lambdas into helper functions?

src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
```diff
@@ -245,23 +230,17 @@ public WorkspaceDescription createWorkspace(
...
-      return workspaceWithContext;
-    } catch (ApiException | InterruptedException ex) {
-      throw new SystemException("Error fetching workspace", ex);
-    }
+    WorkspaceDescription workspaceWithContext =
+        callWithRetries(
+            () -> new WorkspaceApi(apiClient).getWorkspace(workspaceId),
+            "Error fetching workspace");
+    String googleProjectId =
+        (workspaceWithContext.getGcpContext() == null)
```
jaycarlton, line 238: nit: I like Optional/ofNullable/map for this

src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
```diff
@@ -955,4 +854,102 @@ private static boolean isRetryable(Exception ex) {
...
+        ApiException apiEx = (ApiException) ex;
+        logger.error(
+            "WSM exception status code: {}, response body: {}, message: {}",
+            apiEx.getCode(),
+            apiEx.getResponseBody(),
+            apiEx.getMessage());
+
+        // try to deserialize the response body into an ErrorReport
+        String apiExMsg;
+        try {
```
jaycarlton, line 932: this code seems redundant with what's in SamService. Maybe consolidate?

jaycarlton (approved):
Looks good. Lots of smaller calls.

src/main/java/bio/terra/cli/service/SamService.java
jmczerk: Same question, it's not clear to me that this happens now (but maybe that's OK?)

jmczerk (approved):

src/main/java/bio/terra/cli/service/SamService.java
```diff
@@ -132,27 +124,25 @@ public UserStatusInfo getUserInfo() {
...
-    } finally {
-      closeConnectionPool();
-    }
+    return handleClientExceptions(
+        () ->
+            HttpUtils.callAndHandleOneTimeErrorWithRetries(
+                samUsersApi::getUserStatusInfo,
+                SamService::isRetryable,
+                ex -> isStatusCode(ex, HttpStatusCodes.STATUS_CODE_NOT_FOUND),
+                () -> {
```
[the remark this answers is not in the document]
marikomedlock: Factored this out into a separate `registerUser` method.

src/main/java/bio/terra/cli/service/SamService.java
```diff
@@ -285,26 +223,20 @@ public static GroupPolicy fromSamPolicy(String samPolicyName) {
...
-          });
-    } catch (ApiException | InterruptedException ex) {
-      throw new SystemException("Error adding user to SAM group.", ex);
-    } finally {
-      closeConnectionPool();
-    }
+    // - try to add the email to the group
+    // - if this fails with a Bad Request error, it means the email is not found
+    // - so try to invite the user first, then retry adding them to the group
+    handleClientExceptions(
```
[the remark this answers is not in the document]
marikomedlock: Yeah I didn't do this before because this pattern was only used once or twice, so it didn't save much repetition. But from a code readability standpoint, I see your point. Added helper methods that handle the one-time error and log the client exceptions.

src/main/java/bio/terra/cli/service/SamService.java
```diff
@@ -285,26 +223,20 @@ public static GroupPolicy fromSamPolicy(String samPolicyName) {
...
+        () ->
+            HttpUtils.callAndHandleOneTimeErrorWithRetries(
+                () ->
+                    new GroupApi(apiClient)
+                        .addEmailToGroup(groupName, policy.getSamPolicy(), userEmail),
```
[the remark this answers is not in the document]
marikomedlock: I'll keep this suggestion in mind for future improvements. But I think for now, it's not needed because I've wrapped all the calls to these `HttpUtils` methods in helpers for each client library that take care of their retryable arguments, so it's a smaller arg list in the many places this is actually called.

src/main/java/bio/terra/cli/service/SamService.java
marikomedlock: Because now all methods in `SamService` are now wrapped by the `handleClientExceptions` method, which includes this logic. My goal here was to stop duplicating this logic in each method and instead define it in a single method.

marikomedlock: Because now all methods in `SamService` and `WorkspaceManagerService` are now wrapped by the `handleClientExceptions` method, which includes this logic. My goal here was to stop duplicating this logic in each method and instead define it in a single place.

marikomedlock: This does still need to happen, otherwise the CLI hangs after a direct SAM call. I don't understand why this is required, but left a TODO in the code to revisit later.

src/main/java/bio/terra/cli/service/SamService.java
marikomedlock: done

src/main/java/bio/terra/cli/service/SamService.java
marikomedlock: This method is the overload to take a runnable. Or maybe I've misunderstood your comment. Please restate if so.

src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
marikomedlock: This definitely reuses the same `ApiClient` for every retry, and I think also the same `UnauthenticatedApi`.

src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
marikomedlock: `handleClientExceptions` encapsulates a `try/catch/finally` block. I could make helper function(s) for the logic in the `catch` and `finally` blocks, but then I'd still need the same `try/catch/finally` pattern in every public method of this class. I'm not sure how to encapsulate a `try/catch/finally` block without using a lambda. But open to suggestions!

src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
marikomedlock: Frustratingly, the code is very similar and the classes have similar or identical names, but they are not the same classes. (e.g. `ApiException` is not the same class in WSM client library and SAM client library). If the classes at least extended a common base class, that would probably open this up to sharing logic, but unfortunately our client libraries have nothing in common.

src/main/java/bio/terra/cli/service/SamService.java
jaycarlton: It's fine. I was originally thinking you'd add an overload one level upstream in HttpUtils, but yeah, I read it wrong.

PR #108, opened 2021-06-25
PF-771: `resources update` commands for datasets and buckets.

marikomedlock:
- Added two commands for updating BQ dataset and GCS bucket resources: `terra resources update bq-dataset`, `terra resources update gcs-bucket`.
- Created mixins for options that are reused in the create and update commands.
- Bumped the picocli version to the most recent one.


src/main/java/bio/terra/cli/businessobject/Workspace.java
```diff
@@ -168,6 +168,19 @@ public void delete() {
     return listedWorkspaces.stream().map(Workspace::new).collect(Collectors.toList());
   }
 
+  /**
+   * Get a resource by name and expect a specific type.
+   *
+   * @throws UserActionableException if the resource is not found or is the wrong type
+   */
+  public <T extends Resource> T getResourceOfType(String name, Resource.Type type) {
```
wchamber, line 176: consider making the type casting a method on `Resource` that way it can be used anywhere we're trying to downcast the type instead of only when we're `Workspace.get`.

src/main/java/bio/terra/cli/serialization/userfacing/inputs/UpdateResourceParams.java
```diff
@@ -0,0 +1,46 @@
+package bio.terra.cli.serialization.userfacing.inputs;
+
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+
+/**
+ * Parameters for updating a workspace resource. This class is not currently user-facing, but could
+ * be exposed as a command input format in the future. This class handles properties that are common
+ * to all resource types. Sub-classes include additional resource-type specific properties.
```
wchamber, line 10: fix comment, this doesn't use subclasses.

src/main/java/bio/terra/cli/businessobject/resources/BqDataset.java
```diff
@@ -109,6 +103,26 @@ public static BqDataset createControlled(CreateUpdateBqDataset createParams) {
...
+            .updateReferencedBigQueryDataset(Context.requireWorkspace().getId(), id, updateParams);
+        break;
+      case CONTROLLED:
+        WorkspaceManagerService.fromContext()
+            .updateControlledBigQueryDataset(Context.requireWorkspace().getId(), id, updateParams);
+        break;
+      default:
+        throw new IllegalArgumentException("Unknown stewardship type: " + stewardshipType);
+    }
+    super.update(updateParams);
```
wchamber, line 123: it's a little confusing to me that the public method that makes a workspace manager call & the protected method for modifying the resource & state and caching have the same name. Can we differentiate between updating the workspace manager & updating our cached resources?

src/main/java/bio/terra/cli/businessobject/resources/GcsBucket.java
```diff
@@ -100,6 +94,27 @@ public static GcsBucket createControlled(CreateUpdateGcsBucket createParams) {
...
+  /** Update a GCS bucket resource in the workspace. */
+  public void update(UpdateGcsBucketParams updateParams) {
+    if (updateParams.resourceFields.name != null) {
+      validateEnvironmentVariableName(updateParams.resourceFields.name);
+    }
+    switch (stewardshipType) {
+      case REFERENCED:
+        WorkspaceManagerService.fromContext()
+            .updateReferencedGcsBucket(
+                Context.requireWorkspace().getId(), id, updateParams.resourceFields);
```
wchamber, line 106: this is a little funny in that it throws away a lot of the `UpdateGcsBucketParams` fields. Should we just have 2 methods, one for referenced & one for controlled resources update rather than pretending they take the same arguments?

src/main/java/bio/terra/cli/command/resources/update/GcsBucket.java
```diff
@@ -0,0 +1,69 @@
...
+
+  /** Update a Big Query dataset in the workspace. */
+  @Override
+  protected void execute() {
+    workspaceOption.overrideIfSpecified();
+
+    // all update parameters are optional, but make sure at least one is specified
+    if (!resourceUpdateOptions.isDefined()
+        && !storageClassOption.isDefined()
+        && !lifecycleOptions.isDefined()) {
```
wchamber, line 37: can we do that with an ArgGroup? or does that not work with mixins, I forget.

src/main/java/bio/terra/cli/command/shared/options/ResourceUpdate.java
```diff
@@ -0,0 +1,36 @@
...
+ * Command helper class that defines the relevant options for updating a Terra resource: {@link
+ * ResourceName}, --new-name, and {@link ResourceDescription}.
+ *
+ * <p>This class is meant to be used as a @CommandLine.Mixin.
+ */
+public class ResourceUpdate {
+  @CommandLine.Mixin public ResourceName resourceNameOption;
+
+  @CommandLine.Option(
+      names = "--new-name",
```
wchamber, line 16: consider changing workspace update to also take `--new-name` to match this for user consistency.

src/main/java/bio/terra/cli/command/shared/options/ResourceUpdate.java
```diff
@@ -0,0 +1,36 @@
...
+          "New name of the resource. Only alphanumeric and underscore characters are permitted.")
+  public String newName;
+
+  @CommandLine.Mixin public ResourceDescription resourceDescriptionOption;
+
+  /**
+   * Helper method to return a {@link UpdateResourceParams.Builder} with the resource metadata
+   * fields populated.
+   */
+  public UpdateResourceParams.Builder populateMetadataFields() {
```
wchamber, line 27: why the builder instead of the built object?

wchamber (approved):

src/main/java/bio/terra/cli/businessobject/Workspace.java
marikomedlock: good idea. done.

src/main/java/bio/terra/cli/serialization/userfacing/inputs/UpdateResourceParams.java
marikomedlock: left over from a previous iteration. fixed here and in the create version.

src/main/java/bio/terra/cli/businessobject/resources/BqDataset.java
marikomedlock: renamed the superclass method.

src/main/java/bio/terra/cli/command/resources/update/GcsBucket.java
marikomedlock: You can have an ArgGroup that requires at least one option, though the usage help looks pretty weird. e.g.
`([-a=<a>] [-b=<b>] [-c=<c>]) [[-a=<a>] [-b=<b>] [-c=<c>]] [[-a=<a>] [-b=<b>] [-c=<c>]]`

But it doesn't work with mixins (compile time exception `@Mixins are not supported on @ArgGroups`).

src/main/java/bio/terra/cli/command/shared/options/ResourceUpdate.java
marikomedlock: I think these cases are a little different because `name` is a unique identifier for resources, but not for workspaces. Similar to how both use `--description` instead of `--new-description`.

src/main/java/bio/terra/cli/businessobject/resources/GcsBucket.java
marikomedlock: that's fair. I originally had 2 methods, but thought it was better if the command classes didn't have to do different things depending on the stewardship type. On second thought, the command classes have to do argument validation differently based on stewardship type anyway, so this isn't really encapsulating anything. will change back to 2 separate methods.

src/main/java/bio/terra/cli/command/shared/options/ResourceUpdate.java
marikomedlock: Just a hedge for future flexibility. In case the command classes want to do any additional logic or we want to extend this class in the future.

jmczerk (approved):
Sorry for the delay, LGTM

PR #109, opened 2021-06-29
PF-769: Bump Nextflow version, specify new config properties.

marikomedlock:
- Bumped the Nextflow version to an edge release that includes recent config additions to override the default compute engine service account and the default network/subnetwork.
- Included the pet SA email for the current user + current workspace in the output of `terra auth status`. This is so the user can specify that SA when running a Nextflow workflow.
- Defined a new environment variable `GOOGLE_SERVICE_ACCOUNT_EMAIL` when launching apps. This is set to the pet SA email.


build.gradle
```diff
@@ -25,7 +25,7 @@ if (hasProperty('buildScan')) {
 }
 
 // The tools/publish-release.sh script depends on this version string being of the format "version = '1.2.3'"
-version = '0.66.0'
+version = '0.67.0'
```
marikomedlock, line 28: This manual version bump is so that the tests will run using the new Docker image with the NF version bump. It is a shortcoming of the testing infrastructure that it doesn't handle changes to the Dockerfile. This is tracked in [PF-877](https://broadworkbench.atlassian.net/browse/PF-877).

src/main/java/bio/terra/cli/businessobject/User.java
```diff
@@ -175,6 +175,31 @@ public AccessToken getPetSaAccessToken() {
...
+    if (!jsonKeyPath.toFile().exists()) {
+      throw new SystemException("Pet SA key file not found for current user + workspace");
+    }
+    // create a credentials object from the key
+    try {
+      petSACredentials = GoogleOauth.getServiceAccountCredential(jsonKeyPath.toFile(), SCOPES);
+    } catch (IOException ioEx) {
+      throw new SystemException("Error reading pet SA key file.", ioEx);
+    }
+    return petSACredentials.getClientEmail();
```
wchamber, line 200: this looks a lot like `fetchPetSaCredentials()`. Why don't we call that then check `petSACredentials` for the client email or a helper function used by both?

src/test/resources/testscripts/NextflowRnaseq.sh
```diff
@@ -53,12 +53,18 @@ terra nextflow -version
 #      google.location = 'europe-west2'
 #      google.project = "$GOOGLE_CLOUD_PROJECT"
 #      google.region  = 'europe-west1'
+#      google.lifeSciences.serviceAccountEmail = "$GOOGLE_SERVICE_ACCOUNT_EMAIL"
+#      google.lifeSciences.network = 'network'
+#      google.lifeSciences.subnetwork = 'subnetwork'
```
wchamber, line 58: probably worth a comment that these are special values based on the networks created by RBS today for workspace projects.

wchamber (approved):

src/main/java/bio/terra/cli/service/SamService.java
```diff
@@ -547,16 +547,16 @@ private void handleClientExceptions(
...
-          apiExMsg = apiEx.getResponseBody();
-        }
+        String apiExMsg = apiEx.getResponseBody();
+        if (apiExMsg != null)
+          try {
+            ErrorReport errorReport =
+                JacksonMapper.getMapper().readValue(apiEx.getResponseBody(), ErrorReport.class);
+            apiExMsg = errorReport.getMessage();
+          } catch (JsonProcessingException jsonEx) {
+            logger.debug(
```
jaycarlton, line 557: why is this only a debug message? Is it common to have Sam exceptions that can't be parsed?

src/test/java/unit/AuthStatus.java
```diff
@@ -39,6 +39,40 @@ void authStatusWhenLoggedIn() throws IOException {
...
+
+    // check that it says logged in and includes the user, proxy group, and pet SA emails
+    assertThat(
+        "auth status email matches test user",
+        authStatus.userEmail,
+        equalToIgnoringCase(testUser.email));
+    assertThat(
+        "auth status includes proxy group email",
+        authStatus.proxyGroupEmail,
+        CoreMatchers.not(emptyOrNullString()));
```
jaycarlton, line 71: you could use a pattern that matches emails if you want to be sure it's valid output.

jaycarlton (approved):

src/main/java/bio/terra/cli/businessobject/User.java
marikomedlock: Good idea. Added two helper functions.

src/test/java/unit/AuthStatus.java
marikomedlock: Good idea. Added.

src/main/java/bio/terra/cli/service/SamService.java
marikomedlock: I don't know about common, but I have seen it before. This function is just error handling, and I don't want e.g. a NPE to obscure the actual exception that SAM threw.

PR #111, opened 2021-06-30
PF-821: Script to cleanup test user owned workspaces.

marikomedlock:
- Added a script to cleanup test user owned workspaces.
- Added an on-demand GitHub action to trigger the script. I made this on-demand because it can't be run concurrently with any tests. It will cause tests to fail as workspaces are deleted out from under them. We could run this at night, e.g. just after the nightly tests, but that could cause problems if someone decides to re-run the nightly GH action during the day (e.g. failed overnight, fix went in, and now re-run).

### DIFF
--- a/.github/workflows/cleanup-test-user-workspaces.yml
+++ b/.github/workflows/cleanup-test-user-workspaces.yml
@@ -8,11 +8,7 @@ on:
         default: 'true'
       terraserver:
         description: 'the server to cleanup (e.g. terra-dev)'
-        required: false
-        default: 'terra-dev'
-  push:
-    branches:
-      - '**'
+        required: true
 
 jobs:
   cleanup-test-user-workspaces:
--- a/.github/workflows/cleanup-test-user-workspaces.yml
+++ b/.github/workflows/cleanup-test-user-workspaces.yml
@@ -1,0 +1,72 @@
+name: Run tests nightly
+on:
+  workflow_dispatch: {}
+  push:
+    branches:
+      - '**'
+
+inputs:
+  DRY_RUN:
+    description: 'true to do a dry run (i.e. don't actually try to delete anything), false otherwise.'
+    required: false
+    default: 'true'
+  TERRA_SERVER:
+    description: 'the server to cleanup (e.g. terra-dev)'
+    required: false
+    default: 'terra-dev'
+
+jobs:
+  cleanup-test-user-workspaces:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout current code
+        id: checkout_code
+        uses: actions/checkout@v2
+        with:
+          token: ${{ secrets.BROADBOT_GITHUB_TOKEN }}
+      - name: Set up AdoptOpenJDK 11
+        id: setup_jdk
+        uses: joschi/setup-jdk@v2
+        with:
+          java-version: 11
+      - name: Cache Gradle packages
+        id: cache_gradle
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}-${{ hashFiles('**/*.gradle') }}
+          restore-keys: v1-${{ runner.os }}-gradle-${{ hashfiles('**/gradle-wrapper.properties') }}
+      - name: Grant execute permission for gradlew
+        id: allow_execute_gradle
+        run: chmod +x gradlew
+      - name: Render config
+        id: render_config
+        run: |
+          # this step does the equivalent of the tools/render-config.sh script.
+          # on local machines, the script fetches a SA from Vault.
+          # in GH actions, the SA key is stored in a GH repo secret.
+          # regardless of how it was fetched, tests expect these
+          # keys to be stored in rendered/
+          mkdir -p rendered
+          echo "$TEST_USER_SA_KEY" > rendered/test-user-account.json
+          echo "$DEV_CI_SA_KEY" > rendered/ci-account.json
+        env:
+          TEST_USER_SA_KEY: ${{ secrets.TEST_USER_SA_KEY }}
+          DEV_CI_SA_KEY: ${{ secrets.DEV_CI_SA_KEY }}
+      - name: Run cleanup script
+        id: run_cleanup_script
+        run: |
+          echo "DRY_RUN: ${{ github.event.inputs.DRY_RUN }}"
+          echo "TERRA_SERVER: ${{ github.event.inputs.TERRA_SERVER }}"
+#          ./gradlew cleanupTestUserWorkspaces -Pserver=terra-dev
+      - name: Archive logs and context file
+        id: archive_logs_and_context
+        if: always()
+        uses: actions/upload-artifact@v2
+        with:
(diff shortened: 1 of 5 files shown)


src/test/java/harness/utils/CleanupTestUserWorkspaces.java
```diff
@@ -0,0 +1,115 @@
...
+            "Deleting workspace: id=" + workspace.id + ", testuser=" + testUser.email);
+        if (!isDryRun) {
+          // `terra workspace delete --workspace=$id`
+          TestCommand.runCommandExpectSuccess("workspace", "delete", "--workspace=" + workspace.id);
+          System.out.println(
+              "Cleaned up workspace: id=" + workspace.id + ", testuser=" + testUser.email);
+        }
+      } catch (Throwable ex) {
+        System.out.println(
+            "Error deleting workspace: id=" + workspace.id + ", testuser=" + testUser.email);
```
jaycarlton, line 77: should you maybe log `ex`' s message?

src/test/java/harness/utils/CleanupTestUserWorkspaces.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+    TestCommand.runCommandExpectSuccess("server", "set", "--name", server);
+
+    boolean isDryRun = Boolean.parseBoolean(System.getProperty("DRY_RUN"));
+
+    System.out.println("TERRA_SERVER: " + server);
+    System.out.println("DRY_RUN: " + isDryRun);
+
+    Arrays.stream(TestUsers.values())
+        .filter(TestUsers::hasSpendAccess)
```
jaycarlton, line 105: Why do we not care about test users without spend access?

src/test/java/harness/utils/CleanupTestUserWorkspaces.java
```diff
@@ -0,0 +1,115 @@
...
+        .forEach(
+            testUser -> {
+              try {
+                deleteWorkspaces(testUser, isDryRun);
+              } catch (IOException e) {
+                System.out.println("Error cleaning up workspaces for testuser: " + testUser.email);
```
jaycarlton, line 111: Is it possible to list the workspaces that failed to be deleted? You'd have to keep a running list.

jaycarlton (approved):

.github/workflows/cleanup-test-user-workspaces.yml
```diff
@@ -0,0 +1,74 @@
...
+          DEV_CI_SA_KEY: ${{ secrets.DEV_CI_SA_KEY }}
+      - name: Run cleanup script
+        id: run_cleanup_script
+        run: |
+          echo "dryrun: $DRY_RUN"
+          echo "terraserver: $TERRA_SERVER"
+          if [ "$DRY_RUN" == "true" ]; then
+            ./gradlew cleanupTestUserWorkspaces -Pserver=terra-dev -PdryRun
+          else
+            ./gradlew cleanupTestUserWorkspaces -Pserver=terra-dev
```
wchamber, line 61: should this be `-Pserver=$TERRA_SERVER`?

src/test/java/harness/utils/CleanupTestUserWorkspaces.java
```diff
@@ -0,0 +1,115 @@
...
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * This class implements a script to cleanup workspaces owned by test users. Tests are not supposed
+ * to leak workspaces, but if they do, the logging in this class should help track down any
+ * offenders. During development, in particular, workspace leaks are more likely, so this is a
+ * backup cleanup mechanism for those.
```
wchamber, line 20: This assumes that the test users are only used by the Terra CLI. If the test users create workspaces for other test use cases, those will also be deleted here.
Is that a safe assumption? Should we do some workspace tagging to check that it's a workspace created by the CLI tests?

src/test/java/harness/utils/CleanupTestUserWorkspaces.java
marikomedlock: Good idea. Printed out the stacktrace.

src/test/java/harness/utils/CleanupTestUserWorkspaces.java
marikomedlock: Because they can't create workspaces. I only want to delete workspaces that were created by test users during tests.

src/test/java/harness/utils/CleanupTestUserWorkspaces.java
marikomedlock: Sure. Added a summary printout at the end.

.github/workflows/cleanup-test-user-workspaces.yml
marikomedlock: Yes, thanks. Good catch.

src/test/java/harness/utils/CleanupTestUserWorkspaces.java
marikomedlock: The `TestUsers` enum users are intended only for Terra CLI testing, so we should be able to delete everything they create.

My main concern is not that the test users would create workspaces for other use cases. It's that someone might share their workspace with a test user and it would get deleted. That's the reason for the `OWNER` check, but even that's only a partial safeguard because you could make a test user an `OWNER` on a workspace you manually create. I don't have a solution for this. It's the reason for running this manually and offering a dry-run option.

Tagging sounds great, but I'm not sure how to do that reliably. We can't use the name/description because the CLI tests need to test creating workspaces with neither, one, or both of those fields. Any other ideas/suggestions welcome!

Longer-term, it would be great if janitor could also clean up workspaces. I'm not sure exactly how that would work -- maybe WSM tells janitor about a new workspace if some flag in the request is set. The CLI could also tell the janitor, after a new workspace is created and we have the id.

.github/workflows/cleanup-test-user-workspaces.yml
```diff
@@ -0,0 +1,74 @@
+name: Cleanup test user workspaces
+on:
+  workflow_dispatch:
+    inputs:
+      dryrun:
```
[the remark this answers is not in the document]
marikomedlock: Added an underscore for the word separator.

src/test/java/harness/utils/CleanupTestUserWorkspaces.java
wchamber: If these test users are created and noted to only be for Terra CLI testing, sgtm. If someone manually shares a workspace with a test user that's reserved for Terra CLI testing, that's on them, don't do that.

re: tagging. We'd need a WSM feature request to add metadata to workspaces. And the CLI 

For something to work with Janitor, we need it to be able to uniquely identify the resource before the resource is created. This could be done as a WSM client if WSM allows specifying a workspace UUID in the create request, or with an integration inside of WSM where it registers the to-be-created workspace UUID before starting creation based on a request flag or service config. If you can't identify the resource before it's created, you may get an error between creation and registration and then Janitor will never know something was supposed to be cleaned up.
Janitor also needs permissions to delete the resource. So in this case, we would also register the workspace uuid & the test user creating the workspace and assume we can give the Janitor a service account that can impersonate all test users.

wchamber (approved):

PR #113, opened 2021-07-01
PF-889: Remove cloud platform scope from login flow.

marikomedlock:
- Removed the cloud-platform scope from the CLI login flow. Previously, this was needed to create cloud resources on behalf of the user, while the WSM controlled resource endpoints were under development. Now that the CLI always goes through WSM, we don't want to request this scope any more.
- This additional scope was causing logins with Google/Verily corporate emails (i.e. `@google.com`, `@verily.com`) to fail, because there is a corporate policy against granting the cloud-platform scope. This change fixes this.
- Test users still need the cloud-platform scope when they talk to GCP directly (e.g. to query resource properties that are not stored as WSM metadata). Changed the tests to use the same scopes as CLI users, except when they are talking to the cloud directly. In those cases, they use the CLI scopes + cloud-platform.

See also write-up [here](https://docs.google.com/document/d/1m_-UZCaZlzZ48nUF0BdBXghVaD6aSUqXr9XbioXeTYc/edit?resourcekey=0-2GAc8iE2sRjpb_tW-UBJSQ#).

### DIFF
--- a/src/main/java/bio/terra/cli/businessobject/User.java
+++ b/src/main/java/bio/terra/cli/businessobject/User.java
@@ -54,8 +54,16 @@ public class User {
   private ServiceAccountCredentials petSACredentials;
 
   // these are the same scopes requested by Terra service swagger pages
-  private static final List<String> SCOPES =
+  private static final List<String> USER_SCOPES =
       Collections.unmodifiableList(Arrays.asList("openid", "email", "profile"));
+
+  // these are the same scopes requested by Terra service swagger pages, plus the cloud platform
+  // scope. pet SAs need the cloud platform scope to talk to GCP directly (e.g. to check the status
+  // of an AI notebook)
+  public static final List<String> PET_SA_SCOPES =
+      Collections.unmodifiableList(
+          Arrays.asList(
+              "openid", "email", "profile", "https://www.googleapis.com/auth/cloud-platform"));
 
   // google OAuth client secret file
   // (https://developers.google.com/adwords/api/docs/guides/authentication#create_a_client_id_and_client_secret)
@@ -90,7 +98,10 @@ public class User {
           Context.getConfig().getBrowserLaunchOption().equals(Config.BrowserLaunchOption.AUTO);
       userCredentials =
           GoogleOauth.doLoginAndConsent(
-              SCOPES, inputStream, Context.getContextDir().toFile(), launchBrowserAutomatically);
+              USER_SCOPES,
+              inputStream,
+              Context.getContextDir().toFile(),
+              launchBrowserAutomatically);
     } catch (IOException | GeneralSecurityException ex) {
       throw new SystemException("Error fetching user credentials.", ex);
     }
@@ -124,7 +135,8 @@ public class User {
         User.class.getClassLoader().getResourceAsStream(CLIENT_SECRET_FILENAME)) {
 
       // delete the user credentials
-      GoogleOauth.deleteExistingCredential(SCOPES, inputStream, Context.getContextDir().toFile());
+      GoogleOauth.deleteExistingCredential(
+          USER_SCOPES, inputStream, Context.getContextDir().toFile());
 
       // delete the pet SA credentials
       deletePetSaCredentials();
@@ -144,7 +156,7 @@ public class User {
           User.class.getClassLoader().getResourceAsStream(CLIENT_SECRET_FILENAME)) {
         userCredentials =
             GoogleOauth.getExistingUserCredential(
-                SCOPES, inputStream, Context.getContextDir().toFile());
+                USER_SCOPES, inputStream, Context.getContextDir().toFile());
         if (userCredentials == null) {
           return true;
         }
@@ -253,7 +265,7 @@ public class User {
   /** Create a credentials object for a service account from a key file. */
   private ServiceAccountCredentials createSaCredentials(Path jsonKeyPath) {
     try {
-      return GoogleOauth.getServiceAccountCredential(jsonKeyPath.toFile(), SCOPES);
+      return GoogleOauth.getServiceAccountCredential(jsonKeyPath.toFile(), PET_SA_SCOPES);
     } catch (IOException ioEx) {
       throw new SystemException("Error reading SA key file.", ioEx);
     }
--- a/src/main/java/bio/terra/cli/businessobject/User.java
+++ b/src/main/java/bio/terra/cli/businessobject/User.java
@@ -10,14 +10,14 @@ import com.google.api.client.http.HttpStatusCodes;
 import com.google.auth.oauth2.AccessToken;
 import com.google.auth.oauth2.ServiceAccountCredentials;
 import com.google.auth.oauth2.UserCredentials;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableList;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.file.Path;
 import java.security.GeneralSecurityException;
-import java.util.Arrays;
-import java.util.Collections;
 import java.util.Date;
 import java.util.List;
 import java.util.Optional;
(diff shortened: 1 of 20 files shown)


wchamber:
Isn't this going to prevent any direct cloud API calls from the CLI? e.g. start/stop notebook, check access, anything else using the google credentials of the user and not the pet? Even when we move the check access to workspace manager, it uses the end user credentials.
I think many APIs in Terra assume the user has given the cloud platform scope. I'm not sure which ones that's true on in Sam & WSM, but I don't think it's nothing.

src/test/java/harness/TestUsers.java
```diff
@@ -44,6 +44,14 @@
   // name of the group that includes CLI test users and has spend profile access
   public static final String CLI_TEST_USERS_GROUP_NAME = "cli-test-users";
 
+  // these are the same scopes requested by Terra service swagger pages, plus the cloud platform
+  // scope. test users need the cloud platform scope to talk to GCP directly (e.g. to check the
+  // lifecycle property of a GCS bucket, which is not stored as WSM metadata)
```
jaycarlton, line 49: This makes me think you want a separate test helper user in addition to the test users. That way they'd be the same as in production.

marikomedlock:
> Isn't this going to prevent any direct cloud API calls from the CLI? e.g. start/stop notebook, check access, anything else using the google credentials of the user and not the pet? Even when we move the check access to workspace manager, it uses the end user credentials.
> I think many APIs in Terra assume the user has given the cloud platform scope. I'm not sure which ones that's true on in WSM, but I don't think it's nothing.

Ah I think you're right about the start/stop notebook. Not sure about the check-access. I'll test both once SAM is back up. Perhaps this is worth a post-standup conversation. Points to consider:
- Swagger pages don't request cloud platform scope, so if there are endpoints that require it, then they won't work from the Swagger UI.
- Requesting cloud-platform scope blocks use of corporate google/verily accounts, including any potential users within verily.
- I don't think this scope is required by CWB.

wchamber (commented):
(leaving a review comment, so my review bit can be flipped later)
I think we agreed more changes to be done here once we can run tests and see what works with less scopes. And we'll use pets for the notebooks start/stop.

src/main/java/bio/terra/cli/businessobject/User.java
```diff
@@ -54,9 +54,17 @@
...
   // these are the same scopes requested by Terra service swagger pages
-  private static final List<String> SCOPES =
+  private static final List<String> USER_SCOPES =
       Collections.unmodifiableList(Arrays.asList("openid", "email", "profile"));
 
+  // these are the same scopes requested by Terra service swagger pages, plus the cloud platform
+  // scope. pet SAs need the cloud platform scope to talk to GCP directly (e.g. to check the status
+  // of an AI notebook)
+  public static final List<String> PET_SA_SCOPES =
+      Collections.unmodifiableList(
```
jaycarlton, line 64: What's the effective difference between this and `ImmutableList`?

src/main/java/bio/terra/cli/serialization/userfacing/resources/UFAiNotebook.java
```diff
@@ -59,10 +59,10 @@ public void print() {
     OUT.println("GCP project id:                " + projectId);
     OUT.println("AI Notebook instance location: " + location);
     OUT.println("AI Notebook instance id:       " + instanceId);
-    OUT.println("Instance name: " + instanceName == null ? "(undefined)" : instanceName);
-    OUT.println("State:         " + state == null ? "(undefined)" : state);
-    OUT.println("Proxy URL:     " + proxyUri == null ? "(undefined)" : proxyUri);
-    OUT.println("Create time:   " + createTime == null ? "(undefined)" : createTime);
+    OUT.println("Instance name: " + (instanceName == null ? "(undefined)" : instanceName));
```
jaycarlton, line 62: Optional/ofNullable/orElse works well for this situation.

jaycarlton (approved):

src/test/java/harness/TestUsers.java
marikomedlock: Agreed. We should test with the same scopes we give to users.

src/main/java/bio/terra/cli/businessobject/User.java
marikomedlock: Same thing.

src/main/java/bio/terra/cli/serialization/userfacing/resources/UFAiNotebook.java
marikomedlock: I prefer not creating objects where it's not needed. Just a style preference.

src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
```diff
@@ -121,11 +121,20 @@ public static WorkspaceManagerService fromContext() {
         Context.requireUser().getUserAccessToken(), Context.getServer());
   }
 
+  /**
+   * Factory method for class that talks to WSM. Pulls the current server and user from the context.
+   * Uses the pet SA's credentials.
```
wchamber, line 126: probably worth saying why a client would want to use this instead of `fromContext`

src/test/java/harness/utils/ExternalBQDatasets.java
```diff
@@ -45,10 +45,10 @@ public static Dataset createDataset() throws IOException {
    * Grant a given user reader access to a dataset. This method uses SA credentials to set IAM
```
wchamber, line 45: is this still just for 'users'?

src/test/java/harness/utils/ExternalGCSBuckets.java
```diff
@@ -50,13 +50,8 @@ public static Bucket createBucket() throws IOException {
    * Grant a given user object viewer access to a bucket. This method uses SA credentials to set IAM
```
wchamber, line 50: ditto

src/test/java/unit/WorkspaceOverride.java
```diff
@@ -62,12 +65,16 @@ protected void setupOnce() throws IOException {
...
     resetContext();
 
+    workspaceCreator.login();
+
+    // grant the user's proxy group access to the bucket and dataset so that they will pass WSM's
+    // access check when adding them as referenced resources
     externalBucket = ExternalGCSBuckets.createBucket();
-    ExternalGCSBuckets.grantReadAccess(externalBucket, workspaceCreator.email);
+    ExternalGCSBuckets.grantReadAccess(externalBucket, Identity.user(workspaceCreator.email));
+    ExternalGCSBuckets.grantReadAccess(externalBucket, Identity.group(Auth.getProxyGroupEmail()));
```
wchamber, line 74: consider making the proxy group email a member on TestUsers.

wchamber (approved):

src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
marikomedlock: added

src/test/java/harness/utils/ExternalBQDatasets.java
marikomedlock: users or groups. updated comment.

src/test/java/harness/utils/ExternalGCSBuckets.java
marikomedlock: updated comment.

src/test/java/unit/WorkspaceOverride.java
marikomedlock: Yeah I had it that way originally. I changed it because I want to avoid as much as possible duplicating the `User` business object in the `TestUsers` class. We have to override the `login` method because domain-wide delegation doesn't use the OAuth flow. But anything we can do via CLI commands I think is better for testing.

marikomedlock:
Unit test has one failure: `unit.Config` with a SAM socket timeout error from WSM. I ran this locally and it passed.

PR #114, opened 2021-07-07
PF-435: Confirmation prompt for `delete` commands.

marikomedlock:
- Added a confirmation prompt for all `delete` commands: `groups delete`, `resources delete`, `workspace delete`.
  ```Are you sure you want to delete (y/N)?```
- Added an optional `--quiet` flag to suppress the interactive prompt.
- Updated all existing tests to use the `--quiet` flag. Added new tests for the interactive option.

### DIFF
--- a/src/main/java/bio/terra/cli/command/shared/options/DeletePrompt.java
+++ b/src/main/java/bio/terra/cli/command/shared/options/DeletePrompt.java
@@ -1,7 +1,7 @@
 package bio.terra.cli.command.shared.options;
 
 import bio.terra.cli.exception.UserActionableException;
-import bio.terra.cli.utils.Printer;
+import bio.terra.cli.utils.UserIO;
 import java.io.PrintStream;
 import java.nio.charset.StandardCharsets;
 import java.util.Scanner;
@@ -11,8 +11,7 @@ import picocli.CommandLine;
  * Command helper class that defines the --quiet flag for suppressing interactive user input, and
  * prompting the user for a `Y/N`=proceed/abort confirmation if the flag is not specified.
  *
- * <p>Commands that use this option should call {@link #throwIfConfirmationPromptNegative()} before
- * any business logic.
+ * <p>Commands that use this option should call {@link #confirmOrThrow()} before any business logic.
  *
  * <p>This class is meant to be used as a @CommandLine.Mixin.
  */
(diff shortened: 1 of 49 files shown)


src/main/java/bio/terra/cli/command/shared/options/DeletePrompt.java
```diff
@@ -0,0 +1,43 @@
...
+public class DeletePrompt {
+
+  @CommandLine.Option(names = "--quiet", description = "Suppress interactive prompt and delete.")
+  private boolean quiet;
+
+  /**
+   * Helper method to generate an interactive confirmation prompt if the `--quiet` option is not
+   * specified. Throws a {@link UserActionableException} if the prompt response is negative.
+   */
+  public void throwIfConfirmationPromptNegative() {
```
jaycarlton, line 28: maybe just `confirmOrThrow`

src/main/java/bio/terra/cli/utils/Printer.java
```diff
@@ -20,16 +21,19 @@
...
 
   private final PrintStream out;
   private final PrintStream err;
+  private final InputStream in;
 
   private static Printer printer;
 
   /** Constructor that initializes the printer with the specified output streams. */
-  private Printer(PrintStream out, PrintStream err) {
+  private Printer(PrintStream out, PrintStream err, InputStream in) {
```
jaycarlton, line 33: Maybe need to rename the class now that it's reading in addition to printing.

src/main/java/bio/terra/cli/command/shared/options/DeletePrompt.java
```diff
@@ -0,0 +1,43 @@
...
+    boolean promptReturnedYes = true; // default to Y=yes if user input is suppressed
+    PrintStream OUT = Printer.getOut();
+    if (!quiet) {
+      OUT.print("Are you sure you want to delete (y/N)? ");
```
jaycarlton, line 32: scope creep: It might be nice to display some metadata about the thing to be deleted, such as how many resources are contained within a workspace, objects in a bucket, etc.

jaycarlton (approved):

jmczerk (approved):
LGTM

src/main/java/bio/terra/cli/command/shared/options/DeletePrompt.java
marikomedlock: Changed.

src/main/java/bio/terra/cli/utils/Printer.java
marikomedlock: Renamed to `UserIO`.

src/main/java/bio/terra/cli/command/shared/options/DeletePrompt.java
marikomedlock: It's a good suggestion, but I think out of scope for this PR/ticket which is just to prompt for confirmation before deleting. Filed [PF-891](https://broadworkbench.atlassian.net/browse/PF-891).

marikomedlock:
There was 1 unit test failure with a SAM socket timeout exception: ClientExceptionHandling. I ran it locally and it passed.

PR #115, opened 2021-07-07
PF-839 remove Google JupyterLab extensions metadata.

wchamber:
The Google JupyterLab extensions expect projects editor permissions. We're not granting those to users or SAs within workspace projects. Until there is a fully managed AI Notebook service offered by Google, we probably won't be able to use these extensions. Removing them for now.
See also https://broadworkbench.atlassian.net/browse/PF-839


marikomedlock (approved):
LGTM.

I read briefly through the Google groups thread in the JIRA ticket. Out of curiosity, what was the answer to the least required permissions to use these -- storage admin and bq editor?

wchamber:
> LGTM.
> 
> I read briefly through the Google groups thread in the JIRA ticket. Out of curiosity, what was the answer to the least required permissions to use these -- storage admin and bq editor?

They didn't have a concrete answer less than "project editors is what is expected." So they're not sure, but we could manually search for what's required, but I haven't tried since I don't know what we do with that info at this time.

PR #116, opened 2021-07-07
PF-890: Make WSM default spend profile a property of the server.

marikomedlock:
- Added a property to the server specification for the WSM default spend profile. Previously, this was hard-coded to "wm-default-spend-profile".
- Added a separate server specification for the Verily autopush environment, and updated the one for the Verily dev environment.

Eventually, there will be a spend profile manager service and users will want to specify a spend profile to use (e.g. on workspace create or as a CLI config property, etc.). When that time comes, this property will need to be removed. But for now, there is only a single hard-coded spend profile for each WSM deployment. This PR brings the CLI in-line with this association; the spend profile is now a property of the server.

### DIFF
--- a/src/main/java/bio/terra/cli/businessobject/Server.java
+++ b/src/main/java/bio/terra/cli/businessobject/Server.java
@@ -37,6 +37,7 @@ public class Server {
   // (e.g. URLs, WSM single spend profile)
   private String samUri;
   private String workspaceManagerUri;
+  private String wsmDefaultSpendProfile;
   private String dataRepoUri;
 
   private static final String DEFAULT_SERVER_FILENAME = "verily-cli.json";
@@ -49,6 +50,7 @@ public class Server {
     this.description = configFromDisk.description;
     this.samUri = configFromDisk.samUri;
     this.workspaceManagerUri = configFromDisk.workspaceManagerUri;
+    this.wsmDefaultSpendProfile = configFromDisk.wsmDefaultSpendProfile;
     this.dataRepoUri = configFromDisk.dataRepoUri;
   }
 
@@ -167,6 +169,10 @@ public class Server {
 
   public String getWorkspaceManagerUri() {
     return workspaceManagerUri;
+  }
+
+  public String getWsmDefaultSpendProfile() {
+    return wsmDefaultSpendProfile;
   }
 
   public String getDataRepoUri() {
--- a/src/main/resources/servers/terra-verily-autopush.json
+++ b/src/main/resources/servers/terra-verily-autopush.json
@@ -5,5 +5,5 @@
   "dataRepoUri": "https://jade.datarepo-dev.broadinstitute.org",
   "samUri": "https://terra-autopush-sam.api.verily.com",
   "workspaceManagerUri": "http://terra-autopush-wsm.api.verily.com",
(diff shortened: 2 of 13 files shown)


src/main/resources/servers/terra-verily-dev.json
```diff
@@ -1,8 +1,9 @@
...
   "name": "terra-verily-dev",
-  "description": "Terra development environment deployed at Verily.",
+  "description": "Verily dev environment",
 
   "dataRepoUri": "https://jade.datarepo-dev.broadinstitute.org",
-  "samUri": "https://terra-autopush-sam.api.verily.com",
-  "workspaceManagerUri": "http://terra-autopush-wsm.api.verily.com"
+  "samUri": "https://sam-dev.api.verily.com",
+  "workspaceManagerUri": "https://wsm-dev.api.verily.com",
+  "wsmDefaultSpendProfile": "zloery-spend-profile"
```
marikomedlock, line 8: @zloery Could you confirm these URLs and spend profile name? I think the WSM one, in particular, may be wrong.

zloery: The Sam one is right for Verily dev, but WSM is at `https://terra-dev-workspace.api.verily.com`. no idea what's going on with URL consistency

zloery (approved):

src/main/java/bio/terra/cli/businessobject/Server.java
```diff
@@ -37,6 +37,7 @@
   // (e.g. URLs, WSM single spend profile)
   private String samUri;
   private String workspaceManagerUri;
+  private String wsmDefaultSpendProfile;
```
wchamber, line 40: optional: consider renaming to `defaultSpendProfile`. It's kind of the default spend profile for the whole deployment. But it used only in WSM so far (and in our direct Sam group modifications)

wchamber (approved):
LGTM, but no idea on verily URLs.

src/main/resources/servers/terra-verily-autopush.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "terra-verily-autopush",
+  "description": "Verily autopush environment",
+
+  "dataRepoUri": "https://jade.datarepo-dev.broadinstitute.org",
+  "samUri": "https://terra-autopush-sam.api.verily.com",
+  "workspaceManagerUri": "http://terra-autopush-wsm.api.verily.com",
+  "wsmDefaultSpendProfile": "wm-default-spend-profile"
```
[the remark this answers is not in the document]
marikomedlock: done.

src/main/resources/servers/terra-verily-dev.json
marikomedlock: Updated, thanks.

src/main/java/bio/terra/cli/businessobject/Server.java
marikomedlock: Since it's only used for WSM requests at the moment, I'm going to leave it as is. If this expands to other services (e.g. TDR) before we get a spend profile manager, then I agree with your suggestion and will update it here.

marikomedlock:
There were 2 unit test failures with a SAM socket timeout exception: WorkspaceOverride, GcsBucketReferenced. I ran both locally and they passed.

PR #118, opened 2021-07-13
[PF-741] Run nightly tests against autopush

zloery:
This modifies the nightly tests to run against both dev and Verily's autopush environment.

I'm not sure if we want to change the notification policy (e.g. only notify for `dev` or something) - 6 is a lot of daily slack messages, even if they are going to an alert-only channel.

### DIFF
--- a/.github/workflows/tests-nightly.yml
+++ b/.github/workflows/tests-nightly.yml
@@ -65,7 +65,7 @@ jobs:
             build/test-context/.terra/global-context.json
       - name: Notify PF alerts slack channel
         # don't notify manually triggered runs
-        if: always() && github.event_name != 'workflow_dispatch'
+        if: always() && github.event_name == 'workflow_dispatch'
         uses: broadinstitute/action-slack@v3.8.0
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
@@ -73,6 +73,6 @@ jobs:
           MATRIX_CONTEXT: ${{ toJson(matrix) }}
         with:
           status: ${{ job.status }}
-          channel: "#platform-foundation-alerts"
+          channel: "@zloery"
           username: "CLI nightly test run"
           fields: job,took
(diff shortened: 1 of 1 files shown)


.github/workflows/tests-nightly.yml
```diff
@@ -9,6 +9,7 @@ jobs:
     strategy:
       matrix:
         testOptions: [ "-PtestTag=unit", "-PtestTag=integration", "-PtestTag=integration -PtestInstallFromGitHub" ]
+        testServer: ["", "-Pserver=terra-verily-autopush"]
```
marikomedlock, line 12: Can you change the default option to specify the server also: "-Pserver=terra-dev". I think that makes it clearer what the two options are.

.github/workflows/tests-nightly.yml
```diff
@@ -51,9 +52,9 @@ jobs:
       - name: Run tests
         id: run_tests
         run: |
-          # runs against the default server: terra-dev
-          echo "Running tests with options: ${{ matrix.testOptions }}"
-          ./gradlew runTestsWithTag ${{ matrix.testOptions }}
+          # default server is terra-dev
```
marikomedlock, line 55: Remove this if we're no longer relying on the default.

marikomedlock (commented):
RE notifications: Are there any existing notifications about the Verily deployment? I think it would be good to send these test results to the same place.

zloery:
Per discussion elsewhere, platform-foundations-alerts seems like a decent choice for now, and we'll probably move to Verily slack as that becomes available.
I'm running the tests now to see what the modified alerts look like, expect results in ~30m

zloery:
Screen Shot 2021-07-15 at 2 44 56 PM
The notifications are a little smaller, but 6 of them still fill up a screen. Seems ok to me for now

marikomedlock (approved):
lgtm

PR #120, opened 2021-07-14
PF-574: Write up command style guidelines.

marikomedlock:
Wrote up brief notes on command style guidelines. Goal is to make the usage help and commands look consistent.

Most commands already follow these guidelines. I made some trivial (non-functional) changes to commands (e.g. added periods, fixed capitalization) to make the rest consistent.

As part of this style update, I'm also planning to make the command group names consistently singular
`groups` -> `group`
`notebooks` -> `notebook`
`resources` -> `resource`
to match the existing `app`, `server`, `spend`, `workspace`. But I'm going to do this in a follow-on PR.

Reviewers, feel free to include any other suggestions/comments about general style/consistency while I'm doing this once-over.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,6 +30,7 @@
     * [Always specify a description](#always-specify-a-description)
     * [Alphabetize command lists](#alphabetize-command-lists)
     * [User readable exception messages](#user-readable-exception-messages)
+    * [Singular command group  names](#singular-command-group-names)
 
 -----
 
@@ -393,10 +394,10 @@ instead of `terra workspace add-user user@gmail.com READER`.
 This makes it easier to maintain backwards compatibility when adding new arguments. It also makes it easier to read
 commands with multiple arguments, without having to remember the order.
 
-All option names should start with two dashes. e.g. `--email`
-
 The exception to this rule are the `config get-value` and `config set` commands. These will only ever take one argument
 and use a parameter instead of an option.
+
+All option names should start with two dashes. e.g. `--email`
 
 #### Always specify a description
 Specify a description for all commands and options. Write it like a sentence: end with a period and capitalize the first
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,6 +25,11 @@
     * [Serialization](#serialization)
     * [Terra and cloud services](#terra-and-cloud-services)
     * [Servers](#servers)
+6. [Command style guide](#command-style-guide)
+    * [Options instead of parameters](#options-instead-of-parameters)
+    * [Always specify a description](#always-specify-a-description)
+    * [Alphabetize command lists](#alphabetize-command-lists)
+    * [User readable exception messages](#user-readable-exception-messages)
 
 -----
 
@@ -374,3 +379,42 @@ class here.
 
 To add a new server specification, create a new file in this directory and add the file name to the `all-servers.json` 
 file.
+
+
+### Command style guide
+Below are guidelines for adding or modifying commands. The goal is to have a consistent presentation across commands.
+These are not hard rules, but should apply to most commands. Always choose better usability over following a rule here.
+
+#### Options instead of parameters
+Use options instead of parameters, wherever possible. 
+e.g. `terra workspace add-user --email=user@gmail.com --role=READER`
+instead of `terra workspace add-user user@gmail.com READER`.
+
+This makes it easier to maintain backwards compatibility when adding new arguments. It also makes it easier to read
+commands with multiple arguments, without having to remember the order.
+
+All option names should start with two dashes. e.g. `--email`
+
+The exception to this rule are the `config get-value` and `config set` commands. These will only ever take one argument
+and use a parameter instead of an option.
+
+#### Always specify a description
+Specify a description for all commands and options. Write it like a sentence: end with a period and capitalize the first
+letter of the first word. e.g. `Add a user or group to the workspace.`, `Group name.`
+
+Use a verb for the first word of a command or command group description. e.g. `Manage spend profiles.`
+
+#### Alphabetize command lists
+Alphabetize commands by their name (not their Java class name, though that is usually identical) when specifying
+a list of sub-commands. picocli does not do this automatically.
+
+#### User readable exception messages
+`UserActionableException`s are expected in the course of normal use. Their messages should be readable to the user.
+`SystemException`s and any other exceptions are not expected in the course of normal use. They indicate a bug or error 
+that should be reported, so there's no need to make these messages readable to the user.
+
+#### Singular command group names
+Use singular command group names instead of plural. e.g. `terra resource` instead of `terra resources`.
+
+TODO (PF-802): fix this for existing commands
+
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,7 +30,7 @@
     * [Always specify a description](#always-specify-a-description)
     * [Alphabetize command lists](#alphabetize-command-lists)
     * [User readable exception messages](#user-readable-exception-messages)
-    * [Singular command group  names](#singular-command-group-names)
+    * [Singular command group names](#singular-command-group-names)
 
 -----
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -394,8 +394,7 @@ instead of `terra workspace add-user user@gmail.com READER`.
 This makes it easier to maintain backwards compatibility when adding new arguments. It also makes it easier to read
 commands with multiple arguments, without having to remember the order.
 
-The exception to this rule are the `config get-value` and `config set` commands. These will only ever take one argument
-and use a parameter instead of an option.
+The exception to this rule is the `config set` command, which takes one parameter instead of an option.
 
 All option names should start with two dashes. e.g. `--email`
 
--- a/src/main/java/bio/terra/cli/command/config/set/AppLaunch.java
+++ b/src/main/java/bio/terra/cli/command/config/set/AppLaunch.java
@@ -10,7 +10,7 @@ import picocli.CommandLine.Command;
 @Command(name = "app-launch", description = "Configure the ways apps are launched.")
 public class AppLaunch extends BaseCommand {
 
-  @CommandLine.Parameters(index = "0", description = "App launch mode: ${COMPLETION-CANDIDATES}")
+  @CommandLine.Parameters(index = "0", description = "App launch mode: ${COMPLETION-CANDIDATES}.")
   private Config.CommandRunnerOption mode;
 
   /** Updates the command runner option property of the global context. */
--- a/src/main/java/bio/terra/cli/command/config/set/Image.java
+++ b/src/main/java/bio/terra/cli/command/config/set/Image.java
@@ -14,10 +14,10 @@ public class Image extends BaseCommand {
   ImageIdArgGroup argGroup;
 
   static class ImageIdArgGroup {
-    @CommandLine.Option(names = "--image", description = "image id or tag")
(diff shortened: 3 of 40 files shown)


src/main/java/bio/terra/cli/command/config/set/AppLaunch.java
```diff
@@ -10,7 +10,7 @@
 @Command(name = "app-launch", description = "Configure the ways apps are launched.")
 public class AppLaunch extends BaseCommand {
 
-  @CommandLine.Parameters(index = "0", description = "App launch mode: ${COMPLETION-CANDIDATES}")
+  @CommandLine.Parameters(index = "0", description = "App launch mode: ${COMPLETION-CANDIDATES}.")
```
jaycarlton, line 13: Should this description start with a verb?

src/main/java/bio/terra/cli/command/config/set/Image.java
```diff
@@ -14,10 +14,10 @@
   ImageIdArgGroup argGroup;
 
   static class ImageIdArgGroup {
-    @CommandLine.Option(names = "--image", description = "image id or tag")
+    @CommandLine.Option(names = "--image", description = "Docker image id or tag.")
     private String imageId;
 
-    @CommandLine.Option(names = "--default", description = "use the default image id or tag")
+    @CommandLine.Option(names = "--default", description = "Use the default image id or tag.")
```
jaycarlton, line 20: indicate how these two options work together, e.g. default overrides `--image`.

src/main/java/bio/terra/cli/command/config/set/Logging.java
```diff
@@ -13,10 +13,10 @@
   Logging.LogLevelArgGroup argGroup;
 
   static class LogLevelArgGroup {
-    @CommandLine.Option(names = "--console", description = "console logging level")
+    @CommandLine.Option(names = "--console", description = "Console logging level.")
     private boolean console;
 
-    @CommandLine.Option(names = "--file", description = "file logging level")
+    @CommandLine.Option(names = "--file", description = "File logging level.")
```
jaycarlton, line 19: Maybe indicate where the log files live here.

src/main/java/bio/terra/cli/command/config/set/ResourceLimit.java
```diff
@@ -19,12 +19,13 @@
...
     @CommandLine.Option(
         names = "--max",
-        description = "maximum number to allow before throwing an error")
+        description = "Maximum number to allow before throwing an error.")
     private int max;
 
     @CommandLine.Option(
         names = "--default",
-        description = "use the default number of resources: " + Config.DEFAULT_RESOURCES_CACHE_SIZE)
+        description =
```
jaycarlton, line 27: indicate how this works with `--max`

src/main/java/bio/terra/cli/command/resources/addref/BqDataset.java
```diff
@@ -18,10 +18,13 @@
 public class BqDataset extends BaseCommand {
   @CommandLine.Mixin ResourceCreation resourceCreationOptions;
 
-  @CommandLine.Option(names = "--project-id", required = true, description = "GCP project id")
+  @CommandLine.Option(names = "--project-id", required = true, description = "GCP project id.")
```
jaycarlton, line 21: maybe indicate that this is the id of the project containing the dataset, not necessarily the project ID of the active workspace.

src/main/java/bio/terra/cli/command/resources/create/AiNotebook.java
```diff
@@ -71,7 +71,7 @@
           "https://raw.githubusercontent.com/DataBiosphere/terra-cli/main/notebooks/post-startup.sh",
       description =
           "Path to a Bash script that automatically runs after a notebook instance fully boots up. "
-              + "The path must be a URL or Cloud Storage path, e.g. 'gs://path-to-file/file-name'")
+              + "The path must be a URL or Cloud Storage path, e.g. 'gs://path-to-file/file-name'.")
```
jaycarlton, line 74: URI I think they call it

src/main/java/bio/terra/cli/command/resources/create/AiNotebook.java
```diff
@@ -177,7 +177,7 @@
     @CommandLine.Option(
         names = "--install-gpu-driver",
         description =
-            "If true, the end user authorizes Google Cloud to install a GPU driver on this instance")
+            "If true, the end user authorizes Google Cloud to install a GPU driver on this instance.")
```
jaycarlton, line 180: Maybe say that this is an extra charge, but then you have to indicate costs vs free everywhere. The phrase "authorizes Google Cloud to" is understood, since that happens with every option.

src/main/java/bio/terra/cli/command/resources/create/BqDataset.java
```diff
@@ -18,12 +18,15 @@
 public class BqDataset extends BaseCommand {
   @CommandLine.Mixin ControlledResourceCreation controlledResourceCreationOptions;
 
-  @CommandLine.Option(names = "--dataset-id", required = true, description = "Big Query dataset id")
+  @CommandLine.Option(
+      names = "--dataset-id",
+      required = true,
+      description = "Big Query dataset id.")
```
jaycarlton, line 24: `BigQuery`

src/main/java/bio/terra/cli/command/shared/options/ResourceCreation.java
```diff
@@ -18,7 +18,7 @@
   @CommandLine.Option(
       names = "--cloning",
       description =
-          "Instructions for handling when cloning the workspace: ${COMPLETION-CANDIDATES}")
+          "Instructions for handling when cloning the workspace: ${COMPLETION-CANDIDATES}.")
```
jaycarlton, line 21: also when cloning the resource

src/main/java/bio/terra/cli/command/workspace/Update.java
```diff
@@ -20,13 +20,13 @@
   Update.UpdateArgGroup argGroup;
 
   static class UpdateArgGroup {
-    @CommandLine.Option(names = "--name", required = false, description = "workspace display name")
+    @CommandLine.Option(names = "--name", required = false, description = "Workspace display name.")
```
jaycarlton, line 23: maybe describe uniqueness like above.

jaycarlton (approved):

CONTRIBUTING.md
```diff
@@ -374,3 +380,42 @@ class here.
...
+#### Options instead of parameters
+Use options instead of parameters, wherever possible. 
+e.g. `terra workspace add-user --email=user@gmail.com --role=READER`
+instead of `terra workspace add-user user@gmail.com READER`.
+
+This makes it easier to maintain backwards compatibility when adding new arguments. It also makes it easier to read
+commands with multiple arguments, without having to remember the order.
+
+The exception to this rule are the `config get-value` and `config set` commands. These will only ever take one argument
+and use a parameter instead of an option.
```
wchamber, line 398: no strong opinion here, but if there's not a compelling reason to have a single exception, maybe we just shouldn't have an exception. If there's a general rule about when we think using parameters is preferable to using arguments, might be worth trying to capture that here with config get-value and set as examples. (Doesn't `set` kind of take 1 parameter, but get-value takes 0, but get-value can sometimes have a json format argument?)

aside not necessarily this PR: consider making config `get-value` and `set` symmetrical, e.g. either `get` & `set` or `get-value` & `set-value`.

wchamber (approved):

src/main/java/bio/terra/cli/command/config/set/AppLaunch.java
marikomedlock: Parameter and option descriptions don't need to start with a verb, just the description for commands or command groups.

src/main/java/bio/terra/cli/command/config/set/Image.java
marikomedlock: You can only specify one at a time, so there's no overriding. picocli indicates this relationship for us in the auto-generated usage help.
```
> terra config set image
Error: Missing required argument (specify one of these): (--image=<imageId> | --default)
Usage: terra config set image (--image=<imageId> | --default)
```

src/main/java/bio/terra/cli/command/config/set/Logging.java
marikomedlock: When an error occurs, the log file path is printed out to the console. I could see possibly adding a separate command to get the log file path, but I don't think adding it to the config command for the log level is an intuitive place to look for it.

src/main/java/bio/terra/cli/command/config/set/ResourceLimit.java
marikomedlock: Similar to above, you can only specify one at a time and picocli indicates this relationship for us in the auto-generated usage help:
```
> terra config set resource-limit
Error: Missing required argument (specify one of these): (--max=<max> | --default)
Usage: terra config set resource-limit (--max=<max> | --default)
```

src/main/java/bio/terra/cli/command/resources/addref/BqDataset.java
marikomedlock: ok

src/main/java/bio/terra/cli/command/resources/create/AiNotebook.java
marikomedlock: cloud storage [documentation](https://cloud.google.com/storage/docs/cloud-console#_accessing) calls it URL

src/main/java/bio/terra/cli/command/resources/create/AiNotebook.java
marikomedlock: I think this phrase is just roughly copied from the `gcloud notebooks instances create` command usage help:
```
       --install-gpu-driver
          Whether the end user authorizes Google Cloud to install a GPU driver
          on this instance. If this field is empty or set to false, the GPU
          driver won't be installed. Only applicable to instances with GPUs.
```

src/main/java/bio/terra/cli/command/resources/create/BqDataset.java
marikomedlock: I think I've made that mistake throughout the codebase. Will do a search and replace. Thanks for the catch.

src/main/java/bio/terra/cli/command/shared/options/ResourceCreation.java
marikomedlock: updated description to "cloning the workspace or resource"

src/main/java/bio/terra/cli/command/workspace/Update.java
marikomedlock: Good idea. done.

CONTRIBUTING.md
marikomedlock: Ah you're right that `config get-value` doesn't take any parameters, just the `--format` option. The only exception is `config set` that takes a parameter for the config property value. I thought about changing it to use an option instead e.g. `--value`, but that seems uninformative. I don't feel strongly either way though, so happy to be talked out of it.

RE `get-value` and `set`: I did that to be consistent with `gcloud` which uses these same command names. Maybe that's not a very generic choice though, so we don't gain much by copying them. Thoughts?

wchamber: I don't feel particularly strongly either, sorry.
Could do `--get=key` and `--set=key --value=value`, but I don't know that's much improvement.

Didn't know it came from gcloud, but I'd probably lean towards not copying them for this. Maybe that'll upset some gcloud frequent users.

PR #121, opened 2021-07-14
PF-899: Set workspace id in notebook startup script.

marikomedlock:
- Updated the notebook startup script to set the workspace id from the VM metadata, similar to how the server is set.
- Fixed a bug in the text output of the `terra config list` command, where the `app-launch` and `browser` property values were swapped.

### DIFF
--- a/notebooks/README.md
+++ b/notebooks/README.md
@@ -3,5 +3,5 @@
 Directory for AI Notebook scripts.
 
 For now, we're experimenting with post startup scripts for customizing Terra AI Notebook instances.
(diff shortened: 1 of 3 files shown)


wchamber (approved):
I take it you no longer need to be logged in to set the workspace id?

marikomedlock:
> I take it you no longer need to be logged in to set the workspace id?

Right, provided you specify the `--defer-login` flag. PR #119 just went in. I've been rotating reviewers so no one has to do every one. But feel free to weigh in with suggestions/comments for follow-on work.

PR #123, opened 2021-07-16
PF-802: Change command names (group, notebook, resource, config get).

marikomedlock:
This PR contains no functionality changes, just command renames.

- Made pluralization consistent in the command group names: `groups`->`group`, `notebooks`->`notebook`, `resources`->`resource`
- Made the config sub-commands match: `get-value`->`get` now matches `set`. (I originally did this to be consistent with `gcloud config set`, `gcloud config get-value`. While it's good to try and be consistent with other CLIs, this particular choice is maybe not so widespread. This has come up a couple of times in different PR reviews from different people, so I decided to change it.)

Reviewers, if there are any other renames you think would make the CLI more intuitive/clearer, happy to roll them into this change.


wchamber (approved):

README.md
```diff
@@ -217,7 +217,7 @@ Commands:
   workspace  Setup a Terra workspace.
   resources  Manage resources in the workspace.
```
jaycarlton, line 218: should this be `resource`?

README.md
```diff
@@ -217,7 +217,7 @@ Commands:
...
   app        Run applications in the workspace.
-  notebooks  Use AI Notebooks in the workspace.
+  notebook  Use AI Notebooks in the workspace.
   groups     Manage groups of users.
```
jaycarlton, line 221: `group`?

src/main/java/bio/terra/cli/command/Group.java
```diff
@@ -0,0 +1,28 @@
+package bio.terra.cli.command;
```
jaycarlton, line 1: I'm surprised git detected a delete & replace rather than a rename here. The similarity is really high. If you really want to save the history, I think there's a way to adjust the similarity threshold. Probably not worth it here.

jaycarlton (approved):

README.md
marikomedlock: yep. fixed.

README.md
marikomedlock: fixed also, thanks.

src/main/java/bio/terra/cli/command/Group.java
marikomedlock: Yeah I don't know why it did that either. I'm going to leave it as is for now, but good to know we can adjust it in the future if this keeps happening.

PR #124, opened 2021-07-16
PF-877: Test flag to override default Docker image.

marikomedlock:
Added an optional test flag `-PdockerImage` to override the default Docker image that the tests use. This is useful when testing changes to the `docker/` directory (e.g. `Dockerfile`, `terra_init.sh`).

Changed the on-push-to-pr GitHub action to:
- Check whether there are any changes to the `docker/` directory
- If so, it build a new Docker image locally and point the tests at that

### DIFF
--- a/.github/workflows/tests-on-pr-push-and-merge.yml
+++ b/.github/workflows/tests-on-pr-push-and-merge.yml
@@ -78,7 +78,7 @@ jobs:
           # additionally pull the main branch for on-push-to-PRs, so we can diff and see what changed
           if [ "$GHA_EVENT_NAME" = "pull_request" ]; then
             git fetch --no-tags --depth=1 origin main
-            if [-z "$(git diff --name-status origin/main | grep 'docker/')" ]; then
+            if [ -z "$(git diff --name-status origin/main | grep 'docker/')" ]; then
               echo "No changes to docker/ directory. Using default Docker image."
               exit 0
             fi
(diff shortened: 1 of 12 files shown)


.github/workflows/tests-on-pr-push-and-merge.yml
```diff
@@ -70,12 +70,33 @@ jobs:
         env:
           TEST_USER_SA_KEY: ${{ secrets.TEST_USER_SA_KEY }}
           DEV_CI_SA_KEY: ${{ secrets.DEV_CI_SA_KEY }}
+      - name: Build Docker image
+        id: build_docker_image
+        run: |
+          # additionally pull the main branch for on-push-to-PRs, so we can diff and see what changed
+          if [ "$GHA_EVENT_NAME" = "pull_request" ]; then
+            git fetch --no-tags --depth=1 origin main
```
jmczerk, line 78: You might want to use `grep '^docker/'`

jmczerk (approved):

wchamber (approved):

.github/workflows/tests-on-pr-push-and-merge.yml
marikomedlock: `git diff` starts the line with a letter indicating the type of change (e.g. `M` for modified). e.g.
```
> git diff --name-status origin/main
M       .github/workflows/tests-on-pr-push-and-merge.yml
M       docker/scripts/terra_init.sh
```
I'll change it to `grep '\sdocker'` instead, so it matches on the space before the file path.

jmczerk: ah sorry...  you'd need to also use `--name-only` instead of `--name-status`...  Either way works for me...

marikomedlock: Ah makes sense. I didn't think to check for different output formats. Thanks for the pointer!

PR #125, opened 2021-07-27
PF-922: Re-enable notebook tests.

marikomedlock:
Now that permissions for `notebook start/stop` commands are fixed, re-enabled the tests and assertions for these commands.

### DIFF
--- a/src/test/java/unit/AiNotebookControlled.java
+++ b/src/test/java/unit/AiNotebookControlled.java
@@ -203,13 +203,11 @@ public class AiNotebookControlled extends SingleWorkspaceUnit {
     pollDescribeForNotebookState(name, "PROVISIONING");
 
     // `terra notebook start --name=$name`
-    // TODO (PF-869): change this to expect success once polling notebook operations is fixed
-    TestCommand.runCommand("notebook", "start", "--name=" + name);
+    TestCommand.runCommandExpectSuccess("notebook", "start", "--name=" + name);
     pollDescribeForNotebookState(name, "ACTIVE");
 
     // `terra notebook stop --name=$name`
-    // TODO (PF-869): change this to expect success once polling notebook operations is fixed
-    TestCommand.runCommand("notebook", "stop", "--name=" + name);
+    TestCommand.runCommandExpectSuccess("notebook", "stop", "--name=" + name);
     pollDescribeForNotebookState(name, "STOPPED");
   }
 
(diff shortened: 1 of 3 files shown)


src/test/java/unit/AiNotebookControlled.java
```diff
@@ -203,13 +203,11 @@ void startStop() throws IOException, InterruptedException {
...
-    // TODO (PF-869): change this to expect success once polling notebook operations is fixed
-    TestCommand.runCommand("notebook", "start", "--name=" + name);
+    TestCommand.runCommandExpectSuccess("notebook", "start", "--name=" + name);
     pollDescribeForNotebookState(name, "ACTIVE");
 
     // `terra notebook stop --name=$name`
-    // TODO (PF-869): change this to expect success once polling notebook operations is fixed
-    TestCommand.runCommand("notebook", "stop", "--name=" + name);
+    TestCommand.runCommandExpectSuccess("notebook", "stop", "--name=" + name);
     pollDescribeForNotebookState(name, "STOPPED");
```
wchamber, line 211: I don't think we need to poll for a notebook state anymore - once the command has finished successfully, the notebook should be in that state. (Though I'm not sure how long the cli waits for the notebook to take before timing out. Could be a future source of flaky failures).
Ditto let's match the duplicate workspace override test to do the same thing as this one.

wchamber (approved):

src/test/java/unit/AiNotebookControlled.java
marikomedlock: Good point. I'll change it to check the state has changed after the command finishes. If the CLI command for notebook start/stop times out, then this will fail on the previous line `runCommandExpectSuccess` anyway. If we see flaky failures, then we can see about increasing the timeout.

I don't think we need this in the workspace override test. That's just checking that the flag works (i.e. that notebook start/stop work in a different workspace with the flag). This test should cover that `resource describe` state matches what we expect after start/stop.

marikomedlock: OK, interestingly the state doesn't change after start. If it's `PROVISIONING`, the start completes successfully and the state is still `PROVISIONING`. I changed it to poll after create until the state is `ACTIVE`, then start and make sure it's still `ACTIVE`, then stop and make sure it's `STOPPED`, then start one more time and make sure we get a state change to `ACTIVE`.

PR #126, opened 2021-07-27
PF-920, PF-921: Fixes for nightly GH action.

marikomedlock:
Changed the nightly test GH action to run the three types of tests (unit, integration against source code, integration against GH release) serially instead of in parallel. This means the tests take longer to run (~1hr instead of ~35 min), but this is only for the nightly run, so it shouldn't matter. Tests run for PRs still run in parallel. Also compiled the result summary into a single Slack message per server.

- Fixed test run logs/context file getting overwritten by runs on a second server.
- Fixed link to test run not showing up in the Slack message.

Example Slack notification:
Screen Shot 2021-07-27 at 1 51 25 PM
The failure is artificial just to show what the message looks like. I expect all tests to pass against both environments.

### DIFF
--- a/.github/workflows/tests-nightly.yml
+++ b/.github/workflows/tests-nightly.yml
@@ -107,7 +107,7 @@ jobs:
           echo ::set-output name=status-text::$text
       - name: Notify PF alerts slack channel
         # don't notify manually triggered runs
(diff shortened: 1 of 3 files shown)


zloery (approved):
LGTM!

PR #127, opened 2021-07-30
PF-927: Allow deleting a workspace with no cloud context.

marikomedlock:
If a workspace has no cloud context, then it means something went wrong during `workspace create`. In that case, the user should just delete the workspace and create a new one. Previously, this could be difficult to do because loading the workspace would fail when the pet SA fetch from SAM tried to use a null project id. This change is to skip the pet SA fetch if the workspace project id is undefined. This way, the workspace load will succeed and the user can successfully delete it, without errors.

This bug was causing leaked workspaces in tests. When a cloud context failed to be created, the test cleanup failed to delete the corrupted workspace.

### DIFF
--- a/.github/workflows/cleanup-test-user-workspaces.yml
+++ b/.github/workflows/cleanup-test-user-workspaces.yml
@@ -70,7 +70,7 @@ jobs:
         if: always()
         uses: actions/upload-artifact@v2
         with:
-          name: logs-and-context-[${{ matrix.testOptions }}]
+          name: logs-and-context
           path: |
             build/test-context/.terra/logs/
             build/test-context/.terra/global-context.json
--- a/src/main/java/bio/terra/cli/businessobject/User.java
+++ b/src/main/java/bio/terra/cli/businessobject/User.java
@@ -163,10 +163,19 @@ public class User {
    * global context directory.
    */
   public void fetchPetSaCredentials() {
+    // if the cloud context is undefined, then something went wrong during workspace creation
+    // just log an error here so that login will succeed, and the user can go back and delete the
+    // corrupted workspace
+    String googleProjectId = Context.requireWorkspace().getGoogleProjectId();
+    if (googleProjectId == null) {
+      logger.error(
+          "No Google context for the current workspace. Skip fetching pet SA from SAM. MARIKO");
+      return;
+    }
+
     // ask SAM for the project-specific pet SA key
     HttpUtils.HttpResponse petSaKeySamResponse =
-        new SamService(this, Context.getServer())
-            .getPetSaKeyForProject(Context.requireWorkspace().getGoogleProjectId());
+        new SamService(this, Context.getServer()).getPetSaKeyForProject(googleProjectId);
     logger.debug("SAM response to pet SA key request: {})", petSaKeySamResponse);
     if (!HttpStatusCodes.isSuccess(petSaKeySamResponse.statusCode)) {
       throw new SystemException(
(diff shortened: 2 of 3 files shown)


.github/workflows/cleanup-test-user-workspaces.yml
```diff
@@ -70,7 +70,7 @@ jobs:
         if: always()
         uses: actions/upload-artifact@v2
         with:
-          name: logs-and-context-[${{ matrix.testOptions }}]
+          name: logs-and-context
```
jaycarlton, line 73: what is this change for?

src/main/java/bio/terra/cli/businessobject/User.java
```diff
@@ -163,10 +163,18 @@ public void logout() {
    * global context directory.
    */
   public void fetchPetSaCredentials() {
+    // if the cloud context is undefined, then something went wrong during workspace creation
+    // just log an error here instead of throwing an exception, so that the workspace load will
+    // will succeed and the user can delete the corrupted workspace
+    String googleProjectId = Context.requireWorkspace().getGoogleProjectId();
```
jaycarlton, line 169: aside: Are workspaces without cloud contexts always erroneous? Our model allows them AFAIK, and at least one test script uses this approach to save time, as referenced resources don't require cloud contexts.

src/main/java/bio/terra/cli/businessobject/User.java
```diff
@@ -163,10 +163,18 @@ public void logout() {
...
+    if (googleProjectId == null) {
```
jaycarlton, line 170: might as well check for an empty string

src/test/java/harness/utils/CleanupTestUserWorkspaces.java
```diff
@@ -77,8 +77,8 @@ private static void deleteWorkspaces(TestUsers testUser, boolean isDryRun) throw
               "workspace", "delete", "--workspace=" + workspace.id, "--quiet");
           System.out.println(
               "Cleaned up workspace: id=" + workspace.id + ", testuser=" + testUser.email);
-          deletedWorkspaces.add(workspace.id);
         }
+        deletedWorkspaces.add(workspace.id);
```
jaycarlton, line 81: what's the net effect here?

jaycarlton (approved):

.github/workflows/cleanup-test-user-workspaces.yml
marikomedlock: The nightly tests GH action uses a matrix to run tests against different servers. Then it zips up the logs from each test run and uploads them as GH artifacts with the name `logs-and-context-[server_name]`. This GH action doesn't use a matrix and always runs against the `terra-dev` server, so I just removed the suffix.

src/main/java/bio/terra/cli/businessobject/User.java
marikomedlock: For the CLI, yes they are. `terra workspace create` creates a workspace and a Google cloud context. This may need to change in the multi-cloud future, but for now, the only way you can get a workspace w/o a cloud context from the CLI is if there's a failure.

src/main/java/bio/terra/cli/businessobject/User.java
marikomedlock: added check.

src/test/java/harness/utils/CleanupTestUserWorkspaces.java
marikomedlock: In the script to cleanup test users' workspaces, I report out at the end the list of deleted and failed-to-delete workspaces. This script also has a `dryRun` flag that will not actually do the deletion, but will still print out the lists. Before this change, I wasn't adding "successful" workspaces to the list of deleted ones if the `dryRun` flag was set.

jmczerk (approved):

PR #128, opened 2021-08-03
[PF-834] Implement Commands to Set Default Table and Partition Lifetimes in BQ Datasets

jmczerk:
Add options ` --default-partition-lifetime` and ` --default-table-lifetime` to `terra resource update bq-dataset` command.


src/main/java/bio/terra/cli/command/shared/options/BqDatasetExpirationOptions.java
```diff
@@ -0,0 +1,45 @@
...
+    @CommandLine.Option(
+        names = "--partition-expiration",
+        description =
+            "Number of seconds after which to auto-delete newly created partitions in dataset")
+    private Integer partitionExpiration;
+
+    @CommandLine.Option(
+        names = "--table-expiration",
+        description =
+            "Number of seconds after which to auto-delete newly created tables in dataset")
```
marikomedlock, line 26: nit: Perhaps paraphrase/reuse the `bq` CLI descriptions ([ref](https://cloud.google.com/bigquery/docs/updating-datasets#table-expiration)) here? In particular, I think we should include the magic `0` value that removes an existing time.

src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
```diff
@@ -773,12 +775,26 @@ public void updateReferencedBigQueryDataset(
    * @param updateParams resource properties to update
    */
   public void updateControlledBigQueryDataset(
-      UUID workspaceId, UUID resourceId, UpdateResourceParams updateParams) {
+      UUID workspaceId, UUID resourceId, UpdateBqDatasetParams updateParams) {
+
+    GcpBigQueryDatasetUpdateParameters bigQueryDatasetUpdateParameters =
+        new GcpBigQueryDatasetUpdateParameters();
+
+    if (updateParams.partitionExpirationTime.isPresent())
```
marikomedlock, line 783: nit: please use braces around conditionals, at least for the multi-line ones

src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
```diff
@@ -773,12 +775,26 @@ public void updateReferencedBigQueryDataset(
...
+      bigQueryDatasetUpdateParameters.defaultPartitionLifetime(
+          updateParams.partitionExpirationTime.get());
+
+    if (updateParams.tableExpirationTime.isPresent())
+      bigQueryDatasetUpdateParameters.defaultTableLifetime(updateParams.tableExpirationTime.get());
```
marikomedlock, line 788: I think you can set this property to `null` to indicate it's not being updated. e.g.:
`bigQueryDatasetUpdateParameters.defaultTableLifetime(updateParams.tableExpirationTime.orElse(null));`

marikomedlock (commented):
Looks good so far thanks! Couple of comments inline. Also, I know this is still WIP so apologies if you just haven't gotten to this yet, but I think you should also expose these properties in the create call.

jmczerk:
> Looks good so far thanks! Couple of comments inline. Also, I know this is still WIP so apologies if you just haven't gotten to this yet, but I think you should also expose these properties in the create call.

Thank you, this was very helpful!  Now that I've cleaned things up, I will get this added to the create command and add some tests.

src/main/java/bio/terra/cli/command/shared/options/BqDatasetExpirationOptions.java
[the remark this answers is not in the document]
jmczerk: Good idea, it's a little complicated so the verbose description helps.  I also changed to `--default-*-expiration` to align with the `bq` commands.

src/main/java/bio/terra/cli/command/shared/options/BqDatasetExpirationOptions.java
```diff
@@ -0,0 +1,45 @@
...
+import picocli.CommandLine;
+
+/**
+ * Command helper class that defines the expiration options for `terra resource` commands that
+ * handle Big Query Datasets.
+ *
+ * <p>This class is meant to be used as a @CommandLine.Mixin.
+ */
+public class BqDatasetExpirationOptions {
+  @CommandLine.ArgGroup(exclusive = false, multiplicity = "0..1")
```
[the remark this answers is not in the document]
jmczerk: I realized that there wasn't any good reason to make these an arg group and just made them individual options in the class.  I also removed `Options` from the class name to match the other classes in this directory.

src/main/java/bio/terra/cli/serialization/userfacing/input/UpdateBqDatasetParams.java
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Parameters for updating a BigQuery dataset workspace resource. This class is not currently
+ * user-facing, but could be exposed as a command input format in the future.
+ */
+@JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, include = JsonTypeInfo.As.PROPERTY, property = "@class")
+@JsonDeserialize(builder = UpdateResourceParams.Builder.class)
+public class UpdateBqDatasetParams {
+  public final UpdateResourceParams resourceFields;
+  public final Optional<Integer> partitionExpirationTime;
+  public final Optional<Integer> tableExpirationTime;
```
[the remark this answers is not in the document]
jmczerk: Yes, `Optional` was overdoing it here, just went to null `Integer` here.

src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
jmczerk: Yes, simplified in latest commit.

src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
jmczerk: Will do in the future (this statement got taken out in latest commit).

src/main/java/bio/terra/cli/command/shared/options/BqDatasetLifetime.java
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  public Integer getDefaultPartitionLifetime() {
+    return defaultPartitionLifetime;
+  }
+
+  public Integer getDefaultTableLifetime() {
+    return defaultTableLifetime;
+  }
+}
```
jmczerk, line 36: Help message now looks like:
```
Usage: terra resource update bq-dataset
                                        [--default-partition-lifetime=<defaultPa
                                        rtitionLifetime>]
                                        [--default-table-lifetime=<defaultTableL
                                        ifetime>] [--description=<description>]
                                        [--format=<format>] --name=<name>
                                        [--new-name=<newName>]
                                        [--workspace=<id>]
Update a BigQuery dataset.
      --default-partition-lifetime=<defaultPartitionLifetime>
                             The default lifetime, in seconds, for partitions
                               in newly created partitioned tables. This flag
                               has no minimum value. Specify 0 to remove the
                               existing expiration time. Any partitions in
                               newly created partitioned tables are deleted
                               this many seconds after the partition's UTC
                               date. This value is applied if you do not set a
                               partition expiration on the table when it is
                               created.
      --default-table-lifetime=<defaultTableLifetime>
                             The default lifetime, in seconds, for newly
                               created tables. The minimum value is 3600
                               seconds (one hour). The expiration time
                               evaluates to the current UTC time plus this
                               value. Specify 0 to remove the existing
                               expiration time. Any table created in the
                               dataset is deleted this many seconds after its
                               creation time. This value is applied if you do
                               not set a table expiration when the table is
                               created.
      --description=<description>
                             Description of the resource.
      --format=<format>      Set the format for printing command output: json,
                               text.
                               Default: text
      --name=<name>          Name of the resource, scoped to the workspace.
                               Only alphanumeric and underscore characters are
                               permitted.
      --new-name=<newName>   New name of the resource. Only alphanumeric and
                               underscore characters are permitted.
      --workspace=<id>       Workspace id to use for this command only.
```

src/main/java/bio/terra/cli/command/shared/options/BqDatasetLifetime.java
```diff
@@ -0,0 +1,36 @@
...
+ * handle Big Query Datasets.
+ *
+ * <p>This class is meant to be used as a @CommandLine.Mixin.
+ */
+public class BqDatasetLifetime {
+
+  @CommandLine.Option(
+      names = "--default-partition-lifetime",
+      description =
+          "The default lifetime, in seconds, for partitions in newly created partitioned tables. This flag has no minimum value. Specify 0 to remove the existing expiration time. Any partitions in newly created partitioned tables are deleted this many seconds after the partition's UTC date. This value is applied if you do not set a partition expiration on the table when it is created.")
```
wchamber, line 16: `UTC date` doesn't mean much to me. I see you're copying from the bq documentation, but either link it or explain in more detail.

src/main/java/bio/terra/cli/serialization/userfacing/input/CreateBqDatasetParams.java
```diff
@@ -15,12 +15,16 @@
   public final String projectId;
   public final String datasetId;
   public final String location;
+  public final Integer defaultPartitionLifetime;
+  public final Integer defaultTableLifetime;
```
wchamber, line 19: Integer time durations make me unhappy, but if we're really just using them in the serialization format, I guess it's ok. We could look at jackson java8 time support, but that's probably not helpful to users...
Let's at least add meaningful time unit suffixes to this and the other classes.

src/test/java/unit/BqDatasetLifetime.java
```diff
@@ -0,0 +1,227 @@
...
+
+    // `terra workspace set --id=$id`
+    TestCommand.runCommandExpectSuccess("workspace", "set", "--id=" + getWorkspaceId());
+  }
+
+  @AfterEach
+  protected void cleanupEachTime(TestInfo testInfo) {
+
+    // Assumption is that each test case will create a single dataset, named after the test method
+    // itself.  This will clean up after the test case.
```
wchamber, line 44: consider using a `private String datasetName;`, set it in the test, check if it's non-null here and delete it, instead of assuming a name and using TestInfo.

OTOH, why bother? Can the workspace be deleted regardless of if there are datasets in it?

src/test/java/unit/BqDatasetLifetime.java
```diff
@@ -0,0 +1,227 @@
...
+    // to integers representing seconds to be returned in the map, with 0 representing no lifetime
+    // set.  Both of these methods return milliseconds.
+    Long defaultPartitionExpirationMs = dataset.getDefaultPartitionExpirationMs();
+    Long defaultTableLifetime = dataset.getDefaultTableLifetime();
+
+    return Map.of(
+        DefaultLifetimeEntity.PARTITION,
+        defaultPartitionExpirationMs == null ? 0 : defaultPartitionExpirationMs.intValue() / 1000,
+        DefaultLifetimeEntity.TABLE,
+        defaultTableLifetime == null ? 0 : defaultTableLifetime.intValue() / 1000);
```
wchamber, line 225: optional style: this is a little too much logic in the test for my preference. Fine to have a helper method that retrieves the `Dataset`, but consider just asserting directly against `getDefaultPartionExpirationMs` and `getDefaultTableLifetime`, nulls and all.

src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
```diff
@@ -656,7 +658,9 @@ public GcpBigQueryDatasetResource createControlledBigQueryDataset(
             .dataset(
                 new GcpBigQueryDatasetCreationParameters()
                     .datasetId(createParams.datasetId)
-                    .location(createParams.location));
+                    .location(createParams.location)
+                    .defaultPartitionLifetime(createParams.defaultPartitionLifetime)
```
wchamber, line 662: this is coupling our userfacing serialization very directly to the WSM API. We need their semantics around 0 and nulls to be identical, which makes me a little nervous.

wchamber (approved):

src/main/java/bio/terra/cli/command/shared/options/BqDatasetLifetime.java
```diff
@@ -0,0 +1,36 @@
...
+  private Integer defaultPartitionLifetime;
+
+  @CommandLine.Option(
+      names = "--default-table-lifetime",
+      description =
+          "The default lifetime, in seconds, for newly created tables. The minimum value is 3600 seconds (one hour). The expiration time evaluates to the current UTC time plus this value. Specify 0 to remove the existing expiration time. Any table created in the dataset is deleted this many seconds after its creation time. This value is applied if you do not set a table expiration when the table is created.")
+  private Integer defaultTableLifetime;
```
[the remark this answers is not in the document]
jmczerk: This does work, but it necessitates the user inputting durations using [ISO-8601 duration format](https://en.wikipedia.org/wiki/ISO_8601#Durations), which means that:
```
--default-table-lifetime 3600
```
becomes:
```
--default-table-lifetime PT3600S
```
or:
```
--default-table-lifetime PT1H
```

My initial read is that Jackson will parse to/from this format when ser/des a `Duration`, so that could be made consistent.  But I'm not sure we want to push the user to have to look up ISO-8601 for a case where seconds are probably sufficient.  Interested in others' thoughts.

src/main/java/bio/terra/cli/command/shared/options/BqDatasetLifetime.java
jmczerk: Please take a look at the updated description and see if I've wordsmithed it sufficiently to explain the relevance of UTC to the expiration.

src/main/java/bio/terra/cli/businessobject/resource/BqDataset.java
```diff
@@ -106,26 +107,26 @@ public static BqDataset createControlled(CreateBqDatasetParams createParams) {
...
-      default:
-        throw new IllegalArgumentException("Unknown stewardship type: " + stewardshipType);
-    }
+    WorkspaceManagerService.fromContext()
+        .updateReferencedBigQueryDataset(Context.requireWorkspace().getId(), id, updateParams);
     super.updatePropertiesAndSync(updateParams);
   }
 
+  /** Update a BigQuery dataset controlled resource in the workspace. */
+  public void updateControlled(UpdateBqDatasetParams updateParams) {
```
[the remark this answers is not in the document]
jmczerk: Added "controlled" to class comment of new params classes.

src/main/java/bio/terra/cli/command/shared/options/BqDatasetLifetime.java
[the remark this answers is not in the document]
wchamber: sounds good. I think it's worth suffixing with the 'seconds' or 'sec' or 's' time unit then anywhere we're using an integer for a number of seconds.

src/main/java/bio/terra/cli/command/resource/create/BqDataset.java
```diff
@@ -17,6 +18,7 @@
     showDefaultValues = true)
 public class BqDataset extends BaseCommand {
   @CommandLine.Mixin ControlledResourceCreation controlledResourceCreationOptions;
+  @CommandLine.Mixin BqDatasetLifetime bqDatasetLifetime;
```
marikomedlock, line 21: nit: can you use the `Options` suffix for the variable name to be consistent with the mixin above?

src/main/java/bio/terra/cli/command/resource/update/BqDataset.java
```diff
@@ -17,6 +20,7 @@
     showDefaultValues = true)
 public class BqDataset extends BaseCommand {
   @CommandLine.Mixin ResourceUpdate resourceUpdateOptions;
+  @CommandLine.Mixin BqDatasetLifetime bqDatasetLifetime;
```
marikomedlock, line 23: nit: same as above, can you make the variable name consistent with the above using the `Options` suffix?

src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
jmczerk: Would your concern be addressed by not calling the respective setter on the (WSM client library defined) `GcpBigQueryDatasetCreationParameters` if the corresponding parameter is `null`?

src/main/java/bio/terra/cli/command/shared/options/BqDatasetLifetime.java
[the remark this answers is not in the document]
jmczerk: I take the point on adding seconds suffix.  I feel like `--default-table-lifetime-seconds` is getting a little unwieldy, what do folks think about dropping the `default` in the cli switch (`--table-lifetime-seconds`)?  I'm on the fence because both concepts seem notable.

src/main/java/bio/terra/cli/serialization/userfacing/input/CreateBqDatasetParams.java
marikomedlock: I think it's also relevant that the `bq` CLI uses integer seconds for these two properties. If there were no precedent here, I'd probably choose to use days or Duration, but my preference is to mimic the existing cloud CLI where applicable. This is also similar to how we approached the GCS bucket lifecycle rules (i.e. use exact same JSON format as the `gsutil` CLI).

src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
wchamber: Sorry, I'm not following. If the corresponding parameter, say `x` is null, what's the difference between `wsmParams.x(createParams.x)` and `if createParams.x != null { wsmParams.x.(createParams.x); }`
I'm mostly nervous about us changing the semantics of edge case fields in the userfacing serialization or the WSM api and us not remembering to change this code. But introducing a translation layer is maybe overkill. But magic 0 numbers & Integer null semantics make me a little nervous.

src/main/java/bio/terra/cli/command/shared/options/BqDatasetLifetime.java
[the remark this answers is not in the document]
wchamber: You can leave the command name but change the variable name. I'm fine leaving the command as unspecified units asking the user to read the description (which they're probably doing anyway), but `Integer defaultTableLifetime` is just ambiguous, and not all of them on all the classes are commented.

src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
marikomedlock: I agree that coupling user-facing serialization with the WSM API serialization seems like a bad idea. But I think that's addressed here by having separate classes (`CreateBqDatasetParams` and `GcpBigQueryDatasetCreationParameters`) and an explicit translation between them. Perhaps a comment here would be helpful that both serialization classes (and BQ) have the same semantics around special zeros and nulls, and that's why the translation is pretty trivial? Currently we have examples of both trivial translations (e.g. for [AI notebook creation](https://github.com/DataBiosphere/terra-cli/blob/60bb9b54659a3f0376262f6a58a7b200d5b62860/src/main/java/bio/terra/cli/service/WorkspaceManagerService.java#L521)) and more involved ones (e.g. for [bucket lifecycle rules](https://github.com/DataBiosphere/terra-cli/blob/60bb9b54659a3f0376262f6a58a7b200d5b62860/src/main/java/bio/terra/cli/service/WorkspaceManagerService.java#L597)).

src/test/java/unit/BqDatasetLifetime.java
marikomedlock: Agreed, I think can skip deleting resources after each test as long as they're all named something different. They'll all get cleaned up when the workspace is deleted.

src/test/java/unit/BqDatasetLifetime.java
```diff
@@ -0,0 +1,227 @@
...
+        UFBqDataset.class, arguments.toArray(new String[0]));
+  }
+
+  /** Updates a dataset with a given name using the passed arguments. */
+  private void updateDatasetWithDefaultLifetimes(String name, String... defaultLifetimeArguments) {
+    List<String> arguments =
+        new ArrayList<>(Arrays.asList("resource", "update", "bq-dataset", "--name=" + name));
+
+    arguments.addAll(Arrays.asList(defaultLifetimeArguments));
+    TestCommand.runCommandExpectSuccess(arguments.toArray(new String[0]));
```
marikomedlock, line 190: `resource update` should also return a JSON object. Can you add a check for the fields returned there in one of your tests as well?

src/test/java/unit/BqDatasetLifetime.java
```diff
@@ -0,0 +1,227 @@
...
+
+    Dataset dataset =
+        ExternalBQDatasets.getBQClient(workspaceCreator.getCredentialsWithCloudPlatformScope())
+            .getDataset(datasetId);
+
+    assertNotNull(dataset, "looking up dataset via BQ API succeeded");
+
+    // These methods return null if lifetime is not set, if they are set they need to be converted
+    // to integers representing seconds to be returned in the map, with 0 representing no lifetime
+    // set.  Both of these methods return milliseconds.
```
marikomedlock, line 217: Do you need to convert `null` to `0` here? I'm not sure what you get from that over just `assertNull` in your tests.

marikomedlock (approved):
Thanks for all the work on this!

src/main/java/bio/terra/cli/command/shared/options/BqDatasetLifetime.java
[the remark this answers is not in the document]
marikomedlock: +1 for keeping the command name the same, to match what `bq` CLI does.

src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
jmczerk: OK, I think I get it; the WSM side could offer an API that explicitly says “clear the expiration”, in which case we’d want to map (user input of) zero to “clear”, null to “don’t touch it”, and non-zero to “set”.  Your concern is that we make this change in WSM and it doesn’t propagate here?  I understand the concern but will leave this for the moment as I can’t think of a change that adds value on this front.

src/test/java/unit/BqDatasetLifetime.java
jmczerk: Done

src/test/java/unit/BqDatasetLifetime.java
jmczerk: streamlined this checking by using a `Duration` and minimize hand-rolled conversion logic

PR #129, opened 2021-08-05
PF-901: Use CRL Janitor for external buckets, datasets.

marikomedlock:
- The CLI creates buckets and datasets in an external (to WSM) project for testing referenced resources. This PR changes the setup/cleanup of these buckets and datasets to use CRL with Janitor (in cleanup mode) to make sure they get cleaned up eventually.
- Reading information from the cloud about these buckets and datasets as a user would do continues to use the regular GCS/BQ client library, not the CRL wrapper, because we don't expect CLI users to use CRL.
- Many of the changes in this PR are due to moving from the BQ cloud client library to the api client library, which is what CRL wraps.
- While not part of this PR per se, I also added a new GH secret to this repository for the Janitor client SA key in order to get this working.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -65,6 +65,7 @@ dependencies {
         findbugsAnnotations = '3.0.1u2'
 
         // terra libraries
+        crlPlatform = "0.2.0"
         samClient = "0.1-9435410-SNAP"
         workspaceManagerClient = "0.254.67-SNAPSHOT"
         dataRepoClient = "1.0.155-SNAPSHOT"
@@ -99,11 +100,11 @@ dependencies {
     compileOnly "com.google.code.findbugs:annotations:${findbugsAnnotations}"
     testCompileOnly "com.google.code.findbugs:annotations:${findbugsAnnotations}"
 
-    implementation platform('bio.terra.cloud-resource-lib:platform:0.2.0')
-    implementation group: 'bio.terra.cloud-resource-lib', name: 'google-cloudresourcemanager'
-    implementation group: 'bio.terra.cloud-resource-lib', name: 'google-bigquery'
-    implementation group: 'bio.terra.cloud-resource-lib', name: 'google-notebooks'
-    implementation group: 'bio.terra.cloud-resource-lib', name: 'google-storage'
+    implementation platform("bio.terra.cloud-resource-lib:platform:${crlPlatform}")
+    implementation "bio.terra.cloud-resource-lib:google-cloudresourcemanager"
+    implementation "bio.terra.cloud-resource-lib:google-bigquery"
+    implementation "bio.terra.cloud-resource-lib:google-notebooks"
+    implementation "bio.terra.cloud-resource-lib:google-storage"
 
     implementation "org.broadinstitute.dsde.workbench:sam-client_2.12:${samClient}"
     implementation "bio.terra:datarepo-client:${dataRepoClient}"
--- a/build.gradle
+++ b/build.gradle
@@ -65,7 +65,10 @@ dependencies {
         findbugsAnnotations = '3.0.1u2'
 
         // terra libraries
-        crlGoogleNotebooks="0.5.0"
+        crlGoogleCloudResourcemanager="1.3.0"
+        crlGoogleBigQuery="0.11.0"
+        crlGoogleNotebooks="0.8.0"
+        crlGoogleStorage="0.13.0"
         samClient = "0.1-9435410-SNAP"
         workspaceManagerClient = "0.254.41-SNAPSHOT"
         dataRepoClient = "1.0.155-SNAPSHOT"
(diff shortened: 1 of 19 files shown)


src/test/java/harness/utils/ExternalBQDatasets.java
```diff
@@ -21,44 +29,95 @@ public static String randomDatasetId() {
...
+        .execute();
   }
 
-  /** Helper method to build the BQ client object with the given credentials. */
-  public static BigQuery getBQClient(GoogleCredentials credentials) throws IOException {
+  /**
+   * Helper method to build the CRL wrapper around the BQ client object with SA credentials that
+   * have permissions on the external (to WSM) project.
+   */
+  private static BigQueryCow getBQClientWrapper() throws IOException, GeneralSecurityException {
```
wchamber, line 110: I would not bother throwing these checked exceptions and instead wrap and rethrow as runtime exceptions. In tests, there's nothing we reasonably expect the clients to do about this - the environment is messed up. Checked exceptions are for cases where there might be something to do to recover. Propagating the exception up through create/delete isn't adding much.

wchamber, line 110: nit: consider keeping terminology consistent with CRL `getBigQueryCow` or `getBQCow`.

src/test/java/harness/utils/ExternalBQDatasets.java
```diff
@@ -1,17 +1,25 @@
...
+import java.util.List;
 import java.util.UUID;
 
-/** Utility methods for creating external BQ datasets for testing workspace references. */
+/**
+ * Utility methods for creating external BQ datasets for testing workspace references. Most methods
+ * in the class use the CRL wrapper around the BQ client library to manipulate external datasets.
+ * This class also includes a method to build the unwrapped BQ client object, which is what we would
+ * expect users to call. Setup/cleanup of external datasets should use the CRL wrapper. Fetching
+ * information from the cloud as a user may do should use the unwrapped client object.
```
wchamber, line 21: nit: I don't see what we expect users to use has any effect on what client objects we should use. CRL isn't going to do anything unusual in its GCP calls. In these tests, the reason to use CRL is for a) janitor resource clean up and b) (not for BQ where the cloud library is better) but conveniences over the raw gcp api.services libraries. Let's make it clear when it's important to use CRL wrappers. (it doesn't actually do anything special for deletion)
I think there's a consistency argument to be made for only picking a single type of client object, but if it's much more painful to use the cRL api services, then that's fine, but let's comment that as the reasoning.

ditto other comments.

wchamber (approved):
I'll argue a little for picking a single client library to use for tests. As-is, it isn't clear why some but not all client objects have been changed. I think we could mitigate that by explicitly saying what it's needed for (resource creation), and when we prefer to use something else.
As-is, it's not hard to get a non-CRL client and create a resource, but we can rely on code review to see that it doesn't happen.

src/test/java/harness/utils/ExternalBQDatasets.java
marikomedlock: My reason to use CRL in these tests was strictly a) janitor resource clean up. Can you say more about b) the conveniences? For BQ and GCS, the cloud libraries seem good to me in the limited calls we're using it for here anyway.

In general, I think tests should mimic user behavior as best as possible. I don't expect CRL to make any difference here, but it seems unlikely CLI users would use it. That's the only reason I left it out of everything except setup/cleanup.

That's helpful to know about deletion! So for cleanup purposes, the only call we need is for create?

build.gradle
```diff
@@ -65,7 +65,10 @@ dependencies {
         findbugsAnnotations = '3.0.1u2'
```
jmczerk, line 66: Suggestion, not mandatory...  Since you are using all the latest versions, and unless you have a reason for wanting more granular control over individual library versions, you might consider using the new-ish gradle platform that was added to CRL, so that you can just track a single version. 

Here's an example of how it was applied in buffer:

https://github.com/DataBiosphere/terra-resource-buffer/pull/161/commits/29ffa100198871862ebe5b0f570853da5fa7fd90

Current version of `bio.terra.cloud-resource-lib:platform` is still `0.2.0`

jmczerk (approved):

src/test/java/harness/utils/ExternalBQDatasets.java
wchamber: 1. Things like [polling support](https://github.com/DataBiosphere/terra-cli/blob/60bb9b54659a3f0376262f6a58a7b200d5b62860/src/main/java/bio/terra/cli/service/GoogleAiNotebooks.java#L50-L53) & [value classes for nicer APIs](https://github.com/DataBiosphere/terra-cli/blob/60bb9b54659a3f0376262f6a58a7b200d5b62860/src/main/java/bio/terra/cli/service/GoogleAiNotebooks.java#L24). The cloud libraries are reasonably convenient, but when they don't exist, the raw api services libraries aren't great. We could do more than we're doing now in CRL, but if we're going to do anything to improve the usability, we should do it in a central place i.e. CRL.
2. CRL does some consistent logging forms, which might be helpful for CLI debugging without having to rewrite gcp logging help.
3. Yep, create is the only thing. Janitor needs to handle deleting resources whether or not they exist because we can't guarantee that a resource will be created successfully, or that we'll always publish a creation event if we try to wait until after a resource is created.

src/test/java/harness/utils/ExternalBQDatasets.java
marikomedlock: changed to `getBQCow`

marikomedlock: Agreed and fixed.

build.gradle
marikomedlock: Great suggestion thanks! Changed to use the platform instead of individual library versions. This is much better than me looking through Artifactory to figure out what the latest version of each library was.

src/test/java/harness/utils/ExternalBQDatasets.java
marikomedlock: Gotcha, thanks for the additional information and code links. I think we want all the benefits of CRL in both source code and test infrastructure (e.g. create/delete/set iam policy for buckets/datasets), so I'm going to leave the code that uses CRL for things other than just create, even though it's not required for Janitor.

I slightly prefer using the exact same code users would in the actual tests. In general, I think tests should call code the same way users would -- that was one of my main reasons behind the whole Test Runner effort (e.g. use the client library as a caller would, instead of Spring's integration testing support). Even if it very likely won't make much difference in this case, I think this separation between test and source code is important. Less importantly, I don't think we ever want to be in the situation where we need to add functionality to CRL just to support tests, since that seems an unnecessary overhead.

PR #134, opened 2021-09-08
[PF-765] Call WSM endpoint to enable pet impersonation

zloery:
This change makes the CLI call the [newly added WSM endpoint](https://github.com/DataBiosphere/terra-workspace-manager/pull/427) to allow users to impersonate their own pet service accounts as part of running `terra nextflow`. It also adds a standalone action to call the same WSM endpoint for users who want to use the GCP Pipelines API in other ways.

Once this change is merged, I'll remove the project-level permissions that WSM currently grants to all workspace writers+owners. Users will no longer be able to run jobs using pet SAs that they shouldn't, and pet SAs will no longer have permission to impersonate themselves. I've tested this against my personal env with project-level SA permission removed and the unit tests work correctly (minus existing issues with GCS lifecycle rules). However, removing pet's permission to actAs themselves does break the nextflow integration test, as it currently uses the pet SA to  both make and execute requests.


src/main/java/bio/terra/cli/command/app/passthrough/Nextflow.java
```diff
@@ -16,11 +19,17 @@
...
-  /** Pass the command through to the CLI Docker image. */
+  /**
+   * Enable the user's pet SA in the workspace context, then pass the command through to the CLI
+   * Docker image.
+   */
   @Override
   protected void execute() {
     workspaceOption.overrideIfSpecified();
+    String petEmail = Workspace.enablePet();
+    logger.info("Enabled usage of for pet {} for Nextflow", petEmail);
```
marikomedlock, line 32: Often this WSM call will be a no-op, so I think it would be better to either:
- do this here once and then cache that it's already enabled so we don't tack on an extra call to every `nextflow` call
- do this when we fetch the pet SA, which will be out of band of the `nextflow` calls.

The second option means that all pet SAs would be enabled though, not sure if there's any problem with that. I think it would also mean you wouldn't need the separate command to enable a pet.

marikomedlock, line 32: nit: I think the logging should happen in either the `Workspace` or `WorkspaceManagerService` class instead of here in the command.

marikomedlock (commented):
Right as you say re: integration test failure, the pending follow-on changes in WSM will break the `terra nextflow ...` command because all third-party tools run via the CLI use the pet SA credentials. So while I think this is fine to add to the CLI, it won't really fix the existing problem unless we can run Nextflow using the user's credentials. Perhaps worth a post stand-up discussion on a path forward here.

src/main/java/bio/terra/cli/command/app/passthrough/Nextflow.java
zloery: Makes sense. I don't think there's an additional concern with enabling this for every pet, so moving this call to when. we fetch the pet SA credentials seems cleaner.

zloery:
> Right as you say re: integration test failure, the pending follow-on changes in WSM will break the `terra nextflow ...` command because all third-party tools run via the CLI use the pet SA credentials. So while I think this is fine to add to the CLI, it won't really fix the existing problem unless we can run Nextflow using the user's credentials. Perhaps worth a post stand-up discussion on a path forward here.

Per discussion, the approach for the immediate term will be to fix the project-level SA access in WSM and but track pet self-impersonation separately (PF-991). The CLI will still need to call the WSM endpoint, but I think that should be the only necessary change here.

zloery:
One unit test failure is due to a bug in WSM removing roles (added in https://github.com/DataBiosphere/terra-workspace-manager/pull/431, fixed in https://github.com/DataBiosphere/terra-workspace-manager/pull/433). Another failure is due to the `enablePet` endpoint requiring users have WRITE permission, though it will lowered to READ permission in https://github.com/DataBiosphere/terra-workspace-manager/pull/432. It looks like the other test failures are the known notebook operation and lifecycle rule issues.

src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
```diff
@@ -292,6 +292,18 @@ public WorkspaceDescription updateWorkspace(
         "Error updating workspace");
   }
 
+  /**
+   * Call the Workspace Manager POST "/api/workspaces/v1/{id}/gcp/enablepet" endpoint to grant a
+   * user permission to impersonate their own pet service account in a workspace.
```
marikomedlock, line 297: nit 1: to grant a user and their pet this permission
nit 2: "the currently logged in user" instead of "a user"

[dr seuss rhyme here]

src/main/java/bio/terra/cli/businessobject/User.java
```diff
@@ -192,6 +192,12 @@ public void fetchPetSaCredentials() {
       throw new SystemException("Error writing pet SA key to the global context directory.", ioEx);
     }
     logger.debug("Stored pet SA key file for this user and workspace.");
+
+    // Allow the pet to impersonate itself so that Nextflow and other app calls can run.
+    // TODO(PF-991): This will change in the future when WSM disallows SA self-impersonation
```
marikomedlock, line 197: nit: allow the pet and user to impersonate the pet right?
as an aside, I think we'll probably still have to call the WSM endpoint in the future, it's just the endpoint's behavior will have changed, so this particular call here may stay the same.

src/main/java/bio/terra/cli/businessobject/Workspace.java
```diff
@@ -177,6 +177,15 @@ public void delete() {
     Context.setWorkspace(null);
   }
 
+  /**
+   * Enable the current user to impersonate their pet SA in the current workspace.
```
marikomedlock, line 181: same as above, enables current user and their pet to impersonate the pet, I think

marikomedlock (approved):
Thanks for adding this to the CLI!

marikomedlock:
> One unit test failure is due to a bug in WSM removing roles (added in [DataBiosphere/terra-workspace-manager#431](https://github.com/DataBiosphere/terra-workspace-manager/pull/431), fixed in [DataBiosphere/terra-workspace-manager#433](https://github.com/DataBiosphere/terra-workspace-manager/pull/433)). Another failure is due to the `enablePet` endpoint requiring users have WRITE permission, though it will lowered to READ permission in [DataBiosphere/terra-workspace-manager#432](https://github.com/DataBiosphere/terra-workspace-manager/pull/432). It looks like the other test failures are the known notebook operation and lifecycle rule issues.

SGTM. Let's get a run with just the known notebook/lifecycle failures before merging, just to confirm.

src/main/java/bio/terra/cli/businessobject/User.java
zloery: Fixed comment.
I'm not sure, it depends on what changes. If we add identity/client SAs, I expect either WSM will return something different or the CLI will call a different service, so this spot would change. That depends a lot on future designs, though.

marikomedlock: Good point. Agreed we don't have a clear direction forward on this yet. TBDiscussed after you get back refreshed!

zloery:
Latest test run LGTM, the only failure is due to the change being fixed in https://github.com/DataBiosphere/terra-workspace-manager/pull/432. I'll merge these at the same time so the CLI isn't broken by the project-level SA permission change in that PR.

PR #135, opened 2021-09-15
PF-993: Add server definitions for Verily devel and staging. Update autopush.

marikomedlock:
Update URLs and default WSM spend profile.


jmczerk (approved):

marikomedlock:
One test failure with a SAM sockettimeout exception from WSM.

PR #136, opened 2021-09-15
PF-928: Make notebook tests more tolerant of timeouts and 409s.

marikomedlock:
- Notebook start/stop sometimes fails with a 409: unable to queue operation error. Polling the notebook state to make sure that it is ACTIVE/STOPPED before calling stop/start does not prevent this. There is an [open bug](https://github.com/googleapis/java-notebooks/issues/201) against the GCP notebooks Java client for this. In the meantime, I've added a "wait a few minutes and try again" note to the exception message when this error occurs.
- Updated the README with correct commands for giving a user spend profile access.

Test fixes:
- Changed the delete test for a notebook resource to tolerate timeouts in the CLI waiting for the delete job in WSM to complete. In the future this should use long-running job commands, once they are implemented. Reenabled this test.
- Changed the start/stop test for a notebook to tolerate 409s from GCP if the notebook is not in the right state to issue a change of state command.
- Bumped the GCS library version to get a fix for some GCS lifecycle rule properties that were causing tests to fail.

With the above changes, I have gotten many clean runs locally and one clean nightly test run against the `terra-dev` server. The notebook failures are flakey, so further changes may be needed if they recur. The lifecycle tests were failing reliably, so those should be completely fixed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,11 +77,11 @@ owner access to that spend profile.
 - [Preferred] Add a user to a Terra group that is a user of the spend profile. To also grant permission
 to add new members to the group, use `policy=admin` instead.
 
-`terra group add-user --group=enterprise-pilot-testers --policy=member testuser@gmail.com`
+`terra group add-user --name=enterprise-pilot-testers --policy=MEMBER --email=testuser@gmail.com`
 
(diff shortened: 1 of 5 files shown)


wchamber (approved):
thanks!

jmczerk (approved):

PR #137, opened 2021-09-15
PF-994: Define CLI enum for workspace user roles.

marikomedlock:
Defined a CLI enum for workspace user roles: READER, WRITER, OWNER. It is different from the WSM enum for workspace user roles, which also includes the APPLICATION role.

We don't want to expose the APPLICATION role to CLI users at this time. There are no immediate use cases for it, and possible future use cases (e.g. Leo) may not use the CLI to setup the application user anyway. So for now, I've hidden this role from CLI users to avoid confusion. Previously trying to define an application user resulted in an error.

Also fixed an NPE when listing referenced resources in text format.

### DIFF
--- a/src/main/java/bio/terra/cli/businessobject/WorkspaceUser.java
+++ b/src/main/java/bio/terra/cli/businessobject/WorkspaceUser.java
@@ -22,9 +22,39 @@ public class WorkspaceUser {
   private static final Logger logger = LoggerFactory.getLogger(WorkspaceUser.class);
 
   private String email;
-  private List<IamRole> roles;
+  private List<Role> roles;
 
-  private WorkspaceUser(String email, List<IamRole> roles) {
+  /**
+   * Enum for the workspace user roles supported by the CLI. Each enum value maps to a single WSM
+   * client library ({@link bio.terra.workspace.model.IamRole}) enum value.
+   *
+   * <p>The CLI defines its own enum instead of using the WSM one so that we can restrict the roles
+   * supported (e.g. no applications). It also gives the CLI control over what the enum names are,
+   * which are exposed to users as command options.
+   */
+  public enum Role {
+    READER(IamRole.READER),
+    WRITER(IamRole.WRITER),
+    OWNER(IamRole.OWNER);
+
+    private IamRole wsmRole;
+
+    Role(IamRole wsmRole) {
+      this.wsmRole = wsmRole;
+    }
+
+    /** Get the CLI enum value that maps to the given WSM enum value. */
+    public static Role fromWsmRole(IamRole wsmRole) {
+      return Role.valueOf(wsmRole.getValue());
+    }
+
+    /** Get the WSM client enum value that maps to this CLI enum value. */
+    public IamRole getWsmRole() {
+      return wsmRole;
+    }
+  }
+
+  private WorkspaceUser(String email, List<Role> roles) {
     this.email = email;
     this.roles = roles;
   }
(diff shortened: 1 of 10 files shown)


jmczerk (approved):

src/main/java/bio/terra/cli/serialization/userfacing/UFWorkspaceUser.java
```diff
@@ -36,21 +36,21 @@ private UFWorkspaceUser(Builder builder) {
   /** Print out this object in text format. */
   public void print() {
     PrintStream OUT = UserIO.getOut();
-    List<String> rolesStr = roles.stream().map(IamRole::toString).collect(Collectors.toList());
+    List<String> rolesStr = roles.stream().map(Role::toString).collect(Collectors.toList());
     OUT.println(email + ": " + String.join(",", rolesStr));
```
jaycarlton, line 40: nit: you could do the join in your collect statement. collectors.joining or some such.

src/test/java/harness/utils/CleanupTestUserWorkspaces.java
```diff
@@ -61,7 +61,8 @@ private static void deleteWorkspaces(TestUsers testUser, boolean isDryRun) throw
                 .filter(user -> user.email.equalsIgnoreCase(testUser.email))
                 .findAny();
         // skip deleting if the test user is not an owner
-        if (workspaceUser.isEmpty() || !workspaceUser.get().roles.contains(IamRole.OWNER)) {
+        if (workspaceUser.isEmpty()
+            || !workspaceUser.get().roles.contains(WorkspaceUser.Role.OWNER)) {
```
jaycarlton, line 65: nit. Instead of checking empty first you can do map/contains/orElse(false).

jaycarlton (approved):

src/main/java/bio/terra/cli/businessobject/WorkspaceUser.java
```diff
@@ -22,9 +22,39 @@
...
-  private WorkspaceUser(String email, List<IamRole> roles) {
+  /**
+   * Enum for the workspace user roles supported by the CLI. Each enum value maps to a single WSM
+   * client library ({@link bio.terra.workspace.model.IamRole}) enum value.
+   *
+   * <p>The CLI defines its own enum instead of using the WSM one so that we can restrict the roles
+   * supported (e.g. no applications). It also gives the CLI control over what the enum names are,
+   * which are exposed to users as command options.
+   */
+  public enum Role {
```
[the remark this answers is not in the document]
marikomedlock: The `UF` prefix stands for "user-facing" and is used for serialization POJO classes that are exposed to the CLI users, either as inputs (e.g. GCS bucket lifecycle rules json format) or outputs (e.g. workspace object). Similarly, the `PD` prefix stands for "persisted" and is used for serialization POJO classes that are persisted by the CLI in its context directory. I haven't used prefixes for internal business object classes yet.

src/main/java/bio/terra/cli/serialization/userfacing/UFWorkspaceUser.java
marikomedlock: done.

src/test/java/harness/utils/CleanupTestUserWorkspaces.java
marikomedlock: I find having it written out slightly easier to follow. just personal preference.

PR #138, opened 2021-09-16
PF-891: show info on delete prompt

jaycarlton:
Show describe output before the delete prompt to give the user more of a guilt trip/context for the decision. Add properties for dataset table count, bucket object count, and workspace resource and user count.

```
jaycarlton~/repos/terra-cli (jaycarlton-PF-891) $ terra resource list
bar (BQ_DATASET, CONTROLLED)
baz (AI_NOTEBOOK, CONTROLLED, PRIVATE_ACCESS jay.terra.test1@gmail.com)
foo (GCS_BUCKET, CONTROLLED)
jaycarlton~/repos/terra-cli (jaycarlton-PF-891) $ terra resource delete --name=foo
Name:         foo
Description:  
Stewardship:  CONTROLLED
Cloning:      COPY_NOTHING
Access scope: SHARED_ACCESS
Managed by:   USER
GCS bucket name: jay-bar-foo-bar
Object count: 10

Are you sure you want to delete (y/N)? n
Delete aborted.
jaycarlton~/repos/terra-cli (jaycarlton-PF-891) $ terra resource delete --name=bar
Name:         bar
Description:  
Stewardship:  CONTROLLED
Cloning:      COPY_NOTHING
Access scope: SHARED_ACCESS
Managed by:   USER
GCP project id: terra-wsm-d-cozy-okra-7037
BigQuery dataset id: dataset_1
Table count: 1

Are you sure you want to delete (y/N)? n
Delete aborted.
jaycarlton~/repos/terra-cli (jaycarlton-PF-891) $ terra resource delete --name=baz
Name:         baz
Description:  
Stewardship:  CONTROLLED
Cloning:      COPY_NOTHING
Access scope: PRIVATE_ACCESS
Managed by:   USER
Private user: jay.terra.test1@gmail.com
GCP project id:                terra-wsm-d-cozy-okra-7037
AI Notebook instance location: us-central1-a
AI Notebook instance id:       jayterratest1-20210917-130214
Instance name: projects/terra-wsm-d-cozy-okra-7037/locations/us-central1-a/instances/jayterratest1-20210917-130214
State:         ACTIVE
Proxy URL:     f6c6cf6146448f2-dot-us-central1.notebooks.googleusercontent.com
Create time:   2021-09-17T17:02:25.200400204Z

Are you sure you want to delete (y/N)? n
Delete aborted.
jaycarlton~/repos/terra-cli (jaycarlton-PF-891) $ terra workspace delete
Terra workspace id: fbe0d100-50f2-49a0-9558-f6964804ad3d
Display name:       foo1
Description:        
Resource count:     3
User count:         1
Google project:     terra-wsm-d-cozy-okra-7037
Cloud console:      https://console.cloud.google.com/home/dashboard?project=terra-wsm-d-cozy-okra-7037

Are you sure you want to delete (y/N)? n
Delete aborted.
```


src/main/java/bio/terra/cli/serialization/userfacing/UserFacing.java
```diff
@@ -0,0 +1,6 @@
+package bio.terra.cli.serialization.userfacing;
+
+public interface UserFacing {
```
jaycarlton, line 3: Not all UF classes are printable currently, so arguably there should be a narrower interface like `UserFacingPrintable` that only the printing ones use, and no default impl here. I feel like there should be some kind of interface to tie these classes together, for example if you had a list of them and wanted to print them all in sequence.

src/main/java/bio/terra/cli/serialization/userfacing/UFConfig.java
```diff
@@ -20,7 +20,7 @@
  * <p>See the {@link Config} class for a configuration's internal representation.
  */
 @JsonDeserialize(builder = UFConfig.Builder.class)
-public class UFConfig {
+public class UFConfig implements UserFacing {
   public final Config.BrowserLaunchOption browserLaunchOption;
```
jmczerk, line 24: Missing `@Override` on `print()`?

src/main/java/bio/terra/cli/serialization/userfacing/UFGroup.java
```diff
@@ -18,7 +18,7 @@
  * representation.
  */
 @JsonDeserialize(builder = UFGroup.Builder.class)
-public class UFGroup {
```
jmczerk, line 21: Missing `@Override` on `print()`?

src/main/java/bio/terra/cli/service/GoogleGcsBuckets.java
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  /**
+   * Get an exact (slow) count of the objects in a bucket.
+   *
+   * @param bucketName - name of the bucket
+   * @return count
+   */
+  public long getBucketObjectCount(String bucketName) {
+    BucketCow bucketCow = storageCow.get(bucketName);
```
jaycarlton, line 25: For this, I kinda want to say "100+" items or something instead of keeping on paging through. There's a cost to listing the bucket and it could be significant for very very large buckets IIRC.

jmczerk (approved):

src/main/java/bio/terra/cli/service/GoogleGcsBuckets.java
jmczerk: That seems reasonable, I would think that an unexpected "100+ items" should send someone off to the console or `gsutil` to find out more.

src/main/java/bio/terra/cli/businessobject/resource/BqDataset.java
```diff
@@ -183,4 +185,9 @@ public String getProjectId() {
   public String getDatasetId() {
     return datasetId;
   }
+
+  public int getTableCount() {
```
marikomedlock, line 189: nit: method description that notes this is not a simple getter; rather, it queries the cloud for more info

src/main/java/bio/terra/cli/serialization/userfacing/UserFacing.java
marikomedlock: I agree the `UF` classes that implement a `print()` method are similar in that they are user-facing output classes (the classes that don't have `print()` I think are all user-facing input classes). But I'm on the fence about the usefulness of this interface, because I can't think of a situation when we'd want to use a UF class without knowing what type it was (i.e. and have to rely on the interface to know what its behavior is). There are situations when we have a `UFResource` and don't know which sub-class it is, but that already uses inheritance and is solvable by defining an abstract method on the base class. Thoughts?

src/main/java/bio/terra/cli/service/GoogleBqDatasets.java
```diff
@@ -0,0 +1,27 @@
...
+
+import bio.terra.cli.exception.SystemException;
+import bio.terra.cli.service.utils.CrlUtils;
+import bio.terra.cloudres.google.bigquery.BigQueryCow;
+import com.google.api.services.bigquery.model.TableList;
+import com.google.auth.oauth2.GoogleCredentials;
+import com.google.cloud.bigquery.DatasetId;
+import java.io.IOException;
+
+public class GoogleBqDatasets {
```
marikomedlock, line 11: nit: public class should have a description. Also perhaps rename this class to `GoogleBigQuery` since that's the service name.

src/main/java/bio/terra/cli/service/GoogleGcsBuckets.java
marikomedlock: Agreed there's no need to list huge numbers of items. Perhaps you can add a `limit` argument to this method.

src/main/java/bio/terra/cli/service/GoogleGcsBuckets.java
```diff
@@ -0,0 +1,38 @@
...
+
+import bio.terra.cli.service.utils.CrlUtils;
+import bio.terra.cloudres.google.storage.BlobCow;
+import bio.terra.cloudres.google.storage.BucketCow;
+import bio.terra.cloudres.google.storage.StorageCow;
+import com.google.api.gax.paging.Page;
+import com.google.auth.oauth2.GoogleCredentials;
+import java.util.stream.StreamSupport;
+
+public class GoogleGcsBuckets {
```
marikomedlock, line 11: nit: public class should have a description. Also perhaps rename this class to `GoogleCloudStorage` since that's the service name.

src/main/java/bio/terra/cli/businessobject/resource/BqDataset.java
```diff
@@ -183,4 +185,9 @@ public String getProjectId() {
...
+    GoogleBqDatasets datasets = new GoogleBqDatasets(Context.requireUser().getPetSACredentials());
```
marikomedlock, line 190: could you add a factory method for this, similar to WSM service class `fromContextForPetSa()`?

src/main/java/bio/terra/cli/businessobject/resource/GcsBucket.java
```diff
@@ -150,4 +151,14 @@ public String resolve(boolean includePrefix) {
...
     return bucketName;
   }
+
+  /**
+   * Estimated count of all objects in the bucket
+   *
+   * @return number of objects
+   */
+  public long getObjectCount() {
+    GoogleGcsBuckets buckets = new GoogleGcsBuckets(Context.requireUser().getPetSACredentials());
```
marikomedlock, line 161: same as above: could you add a factory method for this, similar to WSM service class `fromContextForPetSa()`?

src/main/java/bio/terra/cli/command/group/Delete.java
```diff
@@ -19,8 +19,10 @@
   /** Delete an existing Terra group. */
   @Override
   protected void execute() {
-    deletePromptOption.confirmOrThrow();
     Group group = Group.get(groupNameOption.name);
+    String description =
+        String.format("This group contains %d members.", group.getMembers().size());
+    deletePromptOption.confirmOrThrow(description);
```
marikomedlock, line 25: This delete command uses a different pattern than the workspace/resource delete command. I think it would be good to be consistent i.e. include this in the description and print that before the prompt here too.

src/main/java/bio/terra/cli/serialization/userfacing/UserFacing.java
jaycarlton: The initial smell was that we had a bunch of classes implementing the same method with the same semantics, which means there's an unwritten abstraction somewhere.

Maybe an interface like `UserFacingOutput` or something would make sense for those that print, and I'd remove the default implementation. You could make a purely decorative interface for the input classes, but with no methods in it it would be kind of odd.

src/main/java/bio/terra/cli/businessobject/resource/BqDataset.java
jaycarlton: sure

src/main/java/bio/terra/cli/command/group/Delete.java
jaycarlton: good catch

src/main/java/bio/terra/cli/service/GoogleGcsBuckets.java
jaycarlton: good idea

jaycarlton:
Incorporated into a separate PR by  @marikomedlock. Closing this one.

PR #141, opened 2021-09-24
PF-1010: Get pet SA token directly from SAM.

marikomedlock:
- Fetch a pet SA access token directly from SAM instead of creating one from the cached pet SA key file.
- Fetch the pet SA email directly from SAM instead of reading the client email from the pet SA key.
- With this change, the CLI still fetches the pet SA key file from SAM and caches it locally. Instead of using this key file every time pet SA credentials are needed, now the CLI only uses this file for setting application default credentials on the Docker container for running third-party tools.

This is a first step towards removing use of the pet SA key file from the CLI. Remaining work is to figure out how to not require this for running apps on the Docker container. Once we have a solution for that, then we can remove this pet SA key file fetch and caching from the CLI entirely.

### DIFF
--- a/src/main/java/bio/terra/cli/businessobject/User.java
+++ b/src/main/java/bio/terra/cli/businessobject/User.java
@@ -165,13 +165,15 @@ public class User {
 
   /**
    * Fetch the pet SA credentials and email for this user + current workspace from SAM and persist
-   * it in the global context directory.
+   * it in the global context directory. This method assumes that this user is the current user, and
+   * that there is a current workspace specified.
    */
   public void fetchPetSaCredentials() {
     // if the cloud context is undefined, then something went wrong during workspace creation
     // just log an error here instead of throwing an exception, so that the workspace load will
     // will succeed and the user can delete the corrupted workspace
-    String googleProjectId = Context.requireWorkspace().getGoogleProjectId();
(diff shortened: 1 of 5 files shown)


src/main/java/bio/terra/cli/businessobject/User.java
```diff
@@ -172,9 +177,13 @@ public void fetchPetSaCredentials() {
       return;
     }
 
+    // ask SAM for the project-specific pet SA email and persist it on disk
+    SamService samService = new SamService(this, Context.getServer());
+    petSAEmail = samService.getPetSaEmailForProject(googleProjectId);
+    Context.setUser(this);
```
zloery, line 183: Should this method have documentation indicating it sets the current user to active in the global context? I'm not super familiar with the CLI, so that may already be implied.

zloery (approved):

src/main/java/bio/terra/cli/businessobject/User.java
marikomedlock: The `Context.setUser` part is just to persist the pet SA email on disk as part of the cached metadata on the current user. It won't ever set a new user as the current one because this method is only called from places where a current user + workspace are already set, but it's a good idea to document that this method expects that. Added comments.

jmczerk (approved):

PR #142, opened 2021-09-29
PF-1006: Break-glass command.

marikomedlock:
- Add a hidden command `terra workspace break-glass` to grant Editor access to a workspace project.
- Included a script to create a central BQ dataset for cataloging break-glass requests. The command writes to this dataset.
- More details and example commands in `CONTRIBUTING.md`.

### DIFF
--- a/src/main/java/bio/terra/cli/command/workspace/BreakGlass.java
+++ b/src/main/java/bio/terra/cli/command/workspace/BreakGlass.java
@@ -15,11 +15,6 @@ import com.google.cloud.bigquery.InsertAllRequest;
 import com.google.cloud.bigquery.InsertAllResponse;
 import com.google.cloud.bigquery.TableId;
 import com.google.common.collect.ImmutableList;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-import picocli.CommandLine;
-import picocli.CommandLine.Command;
-
 import java.io.IOException;
 import java.nio.file.Path;
 import java.util.Date;
@@ -27,6 +22,10 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import picocli.CommandLine;
+import picocli.CommandLine.Command;
 
 /**
  * This class corresponds to the third-level "terra workspace break-glass" command. This command is
--- a/src/main/java/bio/terra/cli/command/workspace/BreakGlass.java
+++ b/src/main/java/bio/terra/cli/command/workspace/BreakGlass.java
@@ -3,8 +3,30 @@ package bio.terra.cli.command.workspace;
 import bio.terra.cli.businessobject.Context;
 import bio.terra.cli.command.shared.BaseCommand;
 import bio.terra.cli.command.shared.options.WorkspaceOverride;
+import bio.terra.cli.exception.SystemException;
+import bio.terra.cli.exception.UserActionableException;
+import bio.terra.cli.service.GoogleOauth;
+import com.google.api.client.util.DateTime;
+import com.google.auth.oauth2.ServiceAccountCredentials;
+import com.google.cloud.bigquery.BigQuery;
+import com.google.cloud.bigquery.BigQueryError;
+import com.google.cloud.bigquery.BigQueryOptions;
+import com.google.cloud.bigquery.InsertAllRequest;
+import com.google.cloud.bigquery.InsertAllResponse;
+import com.google.cloud.bigquery.TableId;
+import com.google.common.collect.ImmutableList;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import picocli.CommandLine;
 import picocli.CommandLine.Command;
+
+import java.io.IOException;
+import java.nio.file.Path;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
 
 /**
  * This class corresponds to the third-level "terra workspace break-glass" command. This command is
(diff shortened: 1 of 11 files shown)


src/main/java/bio/terra/cli/command/workspace/BreakGlass.java
```diff
@@ -0,0 +1,154 @@
...
+      bigQueryCredentials =
+          GoogleOauth.getServiceAccountCredential(Path.of(bigQuerySAKeyFile).toFile(), SA_SCOPES);
+    } catch (IOException ioEx) {
+      throw new UserActionableException("Error reading break-glass SA key files.", ioEx);
+    }
+
+    // grant the user's proxy group the Editor role on the workspace project
+    String granteeProxyGroupEmail =
+        Context.requireWorkspace().grantBreakGlass(granteeEmail, userProjectsAdminCredentials);
+
```
jmczerk, line 97: I assume that if granting fails we'll throw out here?  It might be good to log failed attempts too?

jmczerk (approved):
This looks good.

One thing we might consider, depending on (1) how urgent delivering this is and (2) whether we want to do this everywhere or just in Verily deployments…  If it’s not super urgent and aimed at only Verily deployments, we could consider adding a devoted “break glass” SA and creating the BQ dataset via the Terra Plato Terraform module.  This could allow us to use a single SA for break glass purposes.

zloery (approved):
LGTM
Does this need any additional guardrails against running this in prod/prod-like environments? OTOH requiring the app SA key (which will probably require a separate breakglass into Vault or similar for prod) might already be enough

src/main/java/bio/terra/cli/command/workspace/BreakGlass.java
marikomedlock: Yes, this will throw an exception if it fails and all exceptions are written to the CLI logs. Or do you mean to log it in the BQ dataset?

marikomedlock:
> One thing we might consider, depending on (1) how urgent delivering this is and (2) whether we want to do this everywhere or just in Verily deployments… If it’s not super urgent and aimed at only Verily deployments, we could consider adding a devoted “break glass” SA and creating the BQ dataset via the Terra Plato Terraform module. This could allow us to use a single SA for break glass purposes.

@jmczerk This is currently only intended for the `verily-cli` deployment at the Broad, which is a "personal environment", though in practice it will work on any WSM that shares the same user projects folder (all personal envs and maybe also dev?). If we plan to eventually transition the solutions team over to Verily's devel env, then agreed we should duplicate this in Verily TF.

RE single SA: Agreed this would make the command cleaner, but would it be a maintenance burden to keep the single SA's permissions in sync with the WSM deployments it supports? (e.g. devel WSM has permissions on folder A, staging WSM on folder B. break-glass SA needs both, not sure how to pipe the folder A & B ids into the break-glass SA TF).

> Does this need any additional guardrails against running this in prod/prod-like environments? OTOH requiring the app SA key (which will probably require a separate breakglass into Vault or similar for prod) might already be enough
@zloery I think we're safe from touching prod because the WSM app SAs should only have access to the folder where their user projects live. Even without break-glass, it seems bad if our dev WSM SA could mess with prod stuff. And if we ever did need break-glass in prod (unlikely to be useful on the Verily side per John C, but my understanding is this has come up in AoU so maybe possible on the Broad side), I think we should transition to building this into WSM rather than relying on Vault/CLI.

src/main/java/bio/terra/cli/command/workspace/BreakGlass.java
jmczerk: Yes, it would probably be good to capture failed attempts in BQ dataset if it's not too much effort.

jmczerk:
> @jmczerk This is currently only intended for the `verily-cli` deployment at the Broad, which is a "personal environment", though in practice it will work on any WSM that shares the same user projects folder (all personal envs and maybe also dev?). If we plan to eventually transition the solutions team over to Verily's devel env, then agreed we should duplicate this in Verily TF.
> 
> RE single SA: Agreed this would make the command cleaner, but would it be a maintenance burden to keep the single SA's permissions in sync with the WSM deployments it supports? (e.g. devel WSM has permissions on folder A, staging WSM on folder B. break-glass SA needs both, not sure how to pipe the folder A & B ids into the break-glass SA TF).

Given we're aiming only at `verily-cli` for now, I think we should stick with what you've done.  If we decide that this is something we want to formalize for non-prod Verily environments, I think that TF-based management would have some up front cost, but shouldn't have an ongoing burden.  We now have [terraform for managing the various Plato-side SA permissions on external `verily-bvdp.com` folders](https://github.com/verily-src/terra-bvdp-terraform)).

src/main/java/bio/terra/cli/command/workspace/BreakGlass.java
marikomedlock: It wouldn't be too much effort, but I guess I'm not sure the benefits of logging client-side CLI errors in a central location. A failed attempt to grant break-glass means that:
  - the requestor is not a workspace owner, in which case we'd tell them to request owner access and then re-run the command to log the grant
  - the script/command is broken (e.g. the SA lost folder permissions somehow), and then we need to go debug it

Maybe I'm looking at this wrong though -- how are you thinking we'd use the central record of failed attempts?

jmczerk: I was just thinking that with the pen test coming up, someone might try to use this command as an attack vector to elevate privilege.   "Requestor not being a workspace owner" might be something we want an audit trail for from this perspective.  Agree there's not much value in logging basic user error or bugs in the code.

marikomedlock: That makes sense. I'll add a failure field to the BQ dataset schema and update it if the command fails because the user isn't a workspace owner.

PR #143, opened 2021-10-01
PF-1017: Run nightly tests against devel instead of autopush.

marikomedlock:
- Added two new admin commands for setting up/deleting the default spend profile: `terra spend create-profile`, `terra spend delete-profile`
- Changed the nightly tests GH action to run against the `terra-verily-devel` server instead of `terra-verily-autopush`. I've got a successful run [here](https://github.com/DataBiosphere/terra-cli/actions/runs/1304423547).
- Also some smaller changes:
  - Improved exception logging when using retries. Previously, we were only logging the last exception message, now all are logged.
  - Separated out documentation for the spend profile and break-glass commands into a new ADMIN.md.
  - Changed the script for setting up CLI test users to not quit if the SAM group already exists.

Not part of this PR, but as part of this ticket, I've setup the spend profile and CLI test users on all three Verily deployments: `terra-verily-devel`, `terra-verily-autopush`, `terra-verily-staging`. This means that:
  - the default WSM spend profile exists and has both owner and user policies
  - the `developer-admins` SAM group is defined and populated with PF team members
  - `developer-admins` is an owner of the default WSM spend profile SAM resource
  - SAM group and spend permissions are setup for the CLI test users
  - `developer-admins` is an admin of the `cli-test-users` SAM group

The first three items above are manual setup steps, and the last two are done by running the `tools/setup-test-users.sh` script.

### DIFF
--- a/ADMIN.md
+++ b/ADMIN.md
@@ -43,8 +43,8 @@ group and grant that group OWNER access to the spend profile. e.g.:
 ```
 terra spend create-profile
 terra group create --name=developer-admins
-terra group add-user --email=admin1@gmail.com --policy=ADMIN
-terra group add-user --email=admin2@gmail.com --policy=ADMIN
+terra group add-user --name=developer-admins --email=admin1@gmail.com --policy=ADMIN
+terra group add-user --name=developer-admins --email=admin2@gmail.com --policy=ADMIN
 terra spend enable --email=developer-admins@gmail.com --policy=OWNER
 ```
 
--- a/ADMIN.md
+++ b/ADMIN.md
@@ -1,0 +1,103 @@
+# terra-cli
+
+1. [Spend profile](#spend-profile)
+    * [Grant spend access](#grant-spend-access)
+    * [Setup spend profile](#setup-spend-profile)
+2. [Break-glass](#break-glass)
+    * [Grant break-glass](#grant-break-glass)
+    * [Requests catalog](#requests-catalog)
+
+-----
+
+### Spend profile
+In order to spend money (e.g. by creating a project and resources within it) in Terra, you need
+access to a billing account via a spend profile. In the future, there will be a dedicated Spend
+Profile Manager service that will allow users to setup different profiles that map to different
+billing accounts, and to grant access to users.
+
+Until that new service is build, Workspace Manager recognizes a single spend profile per deployment.
+This single spend profile corresponds to a SAM resource. The name of this resource is specified in
+the Workspace Manager deployment configuration (e.g. Helm charts) but needs to be setup and managed
+manually in SAM (i.e. there are no WSM endpoints for managing spend profiles, you have to talk to SAM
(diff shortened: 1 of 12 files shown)


src/main/java/bio/terra/cli/command/spend/CreateProfile.java
```diff
@@ -0,0 +1,18 @@
+package bio.terra.cli.command.spend;
+
+import bio.terra.cli.command.shared.BaseCommand;
+import bio.terra.cli.service.SpendProfileManagerService;
+import picocli.CommandLine.Command;
+
+/** This class corresponds to the third-level "terra spend create-profile" command. */
+@Command(
```
zloery, line 8: If this and delete-profile are intended to be admin endpoints, should they be hidden?

zloery (approved):

build.gradle
```diff
@@ -66,7 +66,7 @@ dependencies {
 
         // terra libraries
         crlPlatform = "0.2.0"
-        samClient = "0.1-9435410-SNAP"
+        samClient = "0.1-ffb0a89-SNAP" // "0.1-9435410-SNAP"
```
jmczerk, line 69: Are there API changes that we might be missing in our (pretty stale) Sam fork that might require us to update our Verily Sam repo?

jmczerk (approved):

build.gradle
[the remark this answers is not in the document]
marikomedlock: removed

ADMIN.md
```diff
@@ -0,0 +1,103 @@
...
+
+The user who runs the create command will automatically be added as an OWNER of the spend profile, with
+permissions to add new users. You should also make sure that other people have access to the profile,
+so that changes can be made when you're unavailable. The recommended way to do this is to create an admins
+group and grant that group OWNER access to the spend profile. e.g.:
+```
+terra spend create-profile
+terra group create --name=developer-admins
+terra group add-user --email=admin1@gmail.com --policy=ADMIN
+terra group add-user --email=admin2@gmail.com --policy=ADMIN
```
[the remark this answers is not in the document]
marikomedlock: good catch! added

src/main/java/bio/terra/cli/command/spend/CreateProfile.java
marikomedlock: Hmm not sure. I think all the `terra spend` commands should be part of an admin/devops cut of the CLI. On the other hand, these two could be particularly disruptive if someone runs them by accident, similar to break-glass. I'll make just these two hidden on those grounds.

build.gradle
marikomedlock: I got a clean run of the CLI tests against `terra-verily-devel`, so I think we're fine for now. It's possible there are newer API endpoints that won't work, but the CLI tests aren't hitting any of them apparently.

PR #144, opened 2021-10-04
PF-1028: Disable pet SA key file fetch.

marikomedlock:


src/main/java/bio/terra/cli/app/DockerCommandRunner.java
```diff
@@ -57,16 +64,45 @@ protected String wrapCommandInSetupCleanup(List<String> command) {
...
+      // testing flag is not set, this is normal operation
+      // application default credentials must be set to the user or their pet SA
+      AppDefaultCredentialUtils.throwIfADCDontMatchContext();
+      adcBackingFile = AppDefaultCredentialUtils.getADCBackingFile();
+    }
+
+    // mount the application default credentials file to the container
+    // e.g. ADC file (host) $HOME/pet-sa-key.json -> (container)
+    // CONTAINER_HOME_DIR/.terra/pet-keys/[user id]/application_default_credentials.json
+    if (adcBackingFile.isPresent()) {
```
zloery, line 94: nit: this could be an `else` clause of the `if (adcBackingFile.isEmpty())` block above

zloery (approved):
LGTM

src/main/java/bio/terra/cli/app/DockerCommandRunner.java
jmczerk: Is that true given the block above sets `adcBackingFile`?

jmczerk (approved):
Thank you for doing this!!

PR #146, opened 2021-10-12
PF-1026: Rename AI Platform -> GCP notebook.

marikomedlock:
Renamed in commands, docs, and code. Remaining references to AI Platform notebooks are in calls to WSM API and CRL.

No functionality changes.


marikomedlock:
> LGTM it might be worth reaching out to Nicole about this change, I'm not sure if she has any scripts that call the CLI which would need to be updated

@zloery 
Yes you're right, her scripts will need to be updated. She's the one who requested this change though, so she knows that's coming. I'll make sure she knows this went in at our sync meeting this afternoon.

PR #148, opened 2021-10-18
PF-1057: Add confirmation prompt to `server set` command.

marikomedlock:
- Changed `terra auth login` to always trigger a new login. If there is an existing logged in user, they are logged out first. This makes it easier to switch users (i.e. just `terra auth login` instead of `terra auth revoke` first).
- Changed `terra server set` to clear the current auth and workspace context before switching. A workspace only exists on a single server, so when you switch the workspace context becomes invalid but previously wasn't cleared. Right now, all servers use the same Google OAuth flow to login, but that will likely not be true in the future, so we now force users to login again for each server.
- Added a confirmation prompt to `terra server set`, since it will now clear context, and a `--quiet` flag to suppress the prompt.

### DIFF
--- a/src/main/java/bio/terra/cli/command/auth/Login.java
+++ b/src/main/java/bio/terra/cli/command/auth/Login.java
@@ -24,6 +24,7 @@ public class Login extends BaseCommand {
   }
 
   /** Suppress the login by the super class, so that we can logout the user first, if needed. */
+  @Override
   protected boolean requiresLogin() {
     return false;
   }


src/main/java/bio/terra/cli/command/auth/Login.java
```diff
@@ -13,8 +14,17 @@
...
+    if (Context.getUser().isPresent()) {
+      Context.requireUser().logout();
+    }
+
+    User.login();
     OUT.println("Login successful: " + Context.requireUser().getEmail());
   }
+
+  /** Suppress the login by the super class, so that we can logout the user first, if needed. */
+  protected boolean requiresLogin() {
```
zloery, line 27: add `@Override` annotation

PR #160, opened 2021-11-01
PF-1087: Include location in bucket and dataset output.

marikomedlock:
Included the location in the user-facing output (text and JSON) for both bucket and dataset resources. This applies to both controlled and referenced resources. Location is already included for notebooks, so this brings the other resource types in line with that.

### DIFF
--- a/src/main/java/bio/terra/cli/businessobject/resource/GcsBucket.java
+++ b/src/main/java/bio/terra/cli/businessobject/resource/GcsBucket.java
@@ -8,8 +8,12 @@ import bio.terra.cli.serialization.userfacing.input.UpdateGcsBucketParams;
 import bio.terra.cli.serialization.userfacing.input.UpdateResourceParams;
 import bio.terra.cli.serialization.userfacing.resource.UFGcsBucket;
 import bio.terra.cli.service.WorkspaceManagerService;
+import bio.terra.cli.service.utils.CrlUtils;
+import bio.terra.cloudres.google.storage.BucketCow;
+import bio.terra.cloudres.google.storage.StorageCow;
 import bio.terra.workspace.model.GcpGcsBucketResource;
 import bio.terra.workspace.model.ResourceDescription;
+import java.util.Optional;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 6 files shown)


src/main/java/bio/terra/cli/serialization/userfacing/resource/UFBqDataset.java
```diff
@@ -40,12 +47,14 @@ public void print(String prefix) {
     PrintStream OUT = UserIO.getOut();
     OUT.println(prefix + "GCP project id: " + projectId);
     OUT.println(prefix + "BigQuery dataset id: " + datasetId);
+    OUT.println(prefix + "Location: " + location);
```
zloery, line 50: nit: maybe conditionally print this on a non-null `location`, or provide a non-null default? This should always be populated, but this might look strange if it isn't. Same for buckets

src/main/java/bio/terra/cli/businessobject/resource/GcsBucket.java
```diff
@@ -144,6 +148,18 @@ public String resolve(boolean includePrefix) {
     return includePrefix ? GCS_BUCKET_URL_PREFIX + bucketName : bucketName;
   }
 
+  /** Query the cloud for information about the bucket. */
+  public Optional<BucketCow> getBucket() {
```
zloery, line 152: If this is just returning information about the bucket, IMO this should return an `Optional<BucketInfo>` instead of a BucketCow. If there are other uses for it though, BucketCow is fine.

zloery (approved):

src/main/java/bio/terra/cli/serialization/userfacing/resource/UFBqDataset.java
marikomedlock: Good idea. Added an `(undefined)` null display value for buckets, datasets, and notebooks.

src/main/java/bio/terra/cli/businessobject/resource/GcsBucket.java
marikomedlock: For this PR, we're only using information from the wrapped `BucketInfo` object, but I think in the future we want to also use the `BucketCow.list()` method to get the number of blobs in the bucket to include in the are-you-sure-you-want-to-delete-this prompt. So planning to leave this for now, but will keep in mind as I tackle the delete prompt ticket next.

jmczerk (approved):

PR #161, opened 2021-11-01
PF-1102: Better error message when spend profile access denied.

marikomedlock:
Added a better error message when a command fails because a user does not have access to the spend profile. This is now a `UserActionableException` because there is a way for the user to address it: ask an admin to grant them access. This better message should apply regardless of what command the user runs to hit this error (e.g. `workspace create`, `resource create`, etc.).

Previous:
Screen Shot 2021-11-01 at 11 26 59 AM

New:
now

### DIFF
--- a/src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
+++ b/src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
@@ -1149,6 +1149,9 @@ public class WorkspaceManagerService {
       // if this is a WSM client exception, check for a message in the response body
       if (ex instanceof ApiException) {
         String exceptionErrorMessage = logErrorMessage((ApiException) ex);
+
+        // if this is a spend profile access denied error, then throw a more user-friendly error
+        // message
         if (exceptionErrorMessage.contains("spend profile")
             && ((ApiException) ex).getCode() == HttpStatusCodes.STATUS_CODE_FORBIDDEN) {
           throw new UserActionableException(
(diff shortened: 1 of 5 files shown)


src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
```diff
@@ -1148,7 +1148,13 @@ private void handleClientExceptions(
     } catch (ApiException | InterruptedException ex) {
       // if this is a WSM client exception, check for a message in the response body
       if (ex instanceof ApiException) {
-        errorMsg += ": " + logErrorMessage((ApiException) ex);
+        String exceptionErrorMessage = logErrorMessage((ApiException) ex);
+        if (exceptionErrorMessage.contains("User is unauthorized to link spend profile")) {
```
zloery, line 1152: I'm a little uneasy about string matching on error messages, a change to WSM might accidentally break this without anyone knowing. Would it be possible to match on a smaller phrase (maybe just "spend profile") and check error code (in this case 403) instead?

zloery (approved):

jmczerk (approved):

src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
marikomedlock: Good idea. I'll add a check for the status code. Also added a status code check to the place where we check against a SAM error message string.

I agree in general that matching on error messages is error-prone, but not sure a better approach. Maybe in the future we could match on a specific `Exception` class in the `ApiException`'s `cause` field?

PR #162, opened 2021-11-01
PF-891: Include details in delete prompt.

marikomedlock:


src/main/java/bio/terra/cli/serialization/userfacing/resource/UFBqDataset.java
```diff
@@ -21,15 +22,18 @@
...
   /** Serialize an instance of the internal class to the command format. */
   public UFBqDataset(BqDataset internalObj) {
     super(internalObj);
     this.projectId = internalObj.getProjectId();
     this.datasetId = internalObj.getDatasetId();
 
-    Optional<Dataset> dataset = internalObj.getDataset();
+    GoogleBigQuery bigQuery = GoogleBigQuery.fromContextForPetSa();
+    Optional<Dataset> dataset = bigQuery.getDataset(projectId, datasetId);
     this.location = dataset.isPresent() ? dataset.get().getLocation() : null;
```
jaycarlton, line 35: nit: you can use map/orElse instead of the ternary operator.

src/main/java/bio/terra/cli/service/GoogleBigQuery.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+  }
+
+  /**
+   * Utility method that checks if an exception thrown by the BQ client is retryable.
+   *
+   * @param ex exception to test
+   * @return true if the exception is retryable
+   */
+  static boolean isRetryable(Exception ex) {
```
zloery, line 114: Does this need to handle request timeouts? I'm not exactly sure how those are thrown when you're calling CRL, or if that's already handled somewhere else

src/test/java/unit/BqDatasetNumTables.java
```diff
@@ -0,0 +1,159 @@
...
+  @Override
+  protected void setupOnce() throws Exception {
+    super.setupOnce();
+    externalDataset = ExternalBQDatasets.createDataset();
+    ExternalBQDatasets.grantWriteAccess(
+        externalDataset, workspaceCreator.email, ExternalBQDatasets.IamMemberType.USER);
+
+    // grant the user's proxy group access to the dataset so that it will pass WSM's access check
+    // when adding it as a referenced resource
+    ExternalBQDatasets.grantWriteAccess(
```
zloery, line 40: Do these need to be separate? I think the user should be included in their proxy group

zloery (approved):

src/main/java/bio/terra/cli/serialization/userfacing/UFGroup.java
```diff
@@ -21,18 +21,28 @@
...
 
   public UFGroup(Group internalObj) {
     this.name = internalObj.getName();
     this.email = internalObj.getEmail();
     this.currentUserPolicies = internalObj.getCurrentUserPolicies();
+
+    Integer numMembersNotFinal;
+    try {
+      numMembersNotFinal = internalObj.getMembers().size();
+    } catch (Exception ex) {
```
jmczerk, line 35: Just out of interest, under what circumstances does this throw?  What is the logical difference between `numMembers` being `null` vs zero?

src/main/java/bio/terra/cli/serialization/userfacing/resource/UFGcsBucket.java
```diff
@@ -20,21 +21,30 @@
 public class UFGcsBucket extends UFResource {
   public final String bucketName;
   public final String location;
+  public final Integer numObjects;
+
+  // the maximum number of objects to iterate through in the bucket.
+  // if there are more, we just add a "+" at the end for display
```
jmczerk, line 27: Optional, probably overthinking it, but did you consider using the `LINES` envvar if set, otherwise default to `100`?

jmczerk (approved):

src/main/java/bio/terra/cli/serialization/userfacing/resource/UFBqDataset.java
marikomedlock: done.

src/main/java/bio/terra/cli/service/GoogleBigQuery.java
```diff
@@ -0,0 +1,130 @@
...
+    } catch (Exception ex) {
+      return Optional.empty();
+    }
+  }
+
+  /**
+   * Returns the number of tables in the dataset, or null if there was an error looking it up. This
+   * behavior is useful for display purposes.
+   */
+  public Integer getNumTables(String projectId, String datasetId) {
```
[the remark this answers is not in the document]
marikomedlock: good idea. done.

src/main/java/bio/terra/cli/service/GoogleBigQuery.java
marikomedlock: Do you mean a `SocketTimeoutException`? I'm not sure either, but since those CRL methods declare an `IOException` in their signature, makes sense to me to retry those here.

src/test/java/unit/BqDatasetNumTables.java
marikomedlock: Good point! Removed separate access grant to user email.

src/main/java/bio/terra/cli/serialization/userfacing/UFGroup.java
marikomedlock: This throws if the user does not have permission to list the group members. Only group admins have this permissions, so non-admins can see/describe the group, but they can't see how many people are in it. The text display value will be `(unknown)` if `numMembers` is null. Will add a comment.

src/main/java/bio/terra/cli/serialization/userfacing/resource/UFGcsBucket.java
marikomedlock: I thought the `LINES` is typically used for terminal/window size. That doesn't seem to apply here, since we're not displaying the bucket objects, just the count. Have you seen it used in other circumstances?

jmczerk: Ah yes, I misinterpreted that we were going to print object ID's.  Disregard.

PR #165, opened 2021-11-11
[PF-1103] delete empty workspace when creating cloud context fails

jaycarlton:
Previously, when context creation failed, we simply aborted the operation, leaving a new stub workspace in the database. With this change, we'll delete such workspaces in this scenario.

### DIFF
--- a/src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
+++ b/src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
@@ -61,6 +61,7 @@ import bio.terra.workspace.model.GrantRoleRequestBody;
 import bio.terra.workspace.model.IamRole;
 import bio.terra.workspace.model.JobControl;
 import bio.terra.workspace.model.JobReport;
+import bio.terra.workspace.model.JobReport.StatusEnum;
 import bio.terra.workspace.model.ManagedBy;
 import bio.terra.workspace.model.PrivateResourceIamRoles;
 import bio.terra.workspace.model.PrivateResourceUser;
@@ -242,6 +243,13 @@ public class WorkspaceManagerService {
                   CREATE_WORKSPACE_MAXIMUM_RETRIES,
                   CREATE_WORKSPACE_DURATION_SLEEP_FOR_RETRY);
           logger.debug("create workspace context result: {}", createContextResult);
+          StatusEnum status = createContextResult.getJobReport().getStatus();
+          if (StatusEnum.FAILED == status || StatusEnum.RUNNING == status) {
+            // need to delete the empty workspace before bailing below
+            HttpUtils.callWithRetries(
+                () -> workspaceApi.deleteWorkspace(workspaceId),
+                WorkspaceManagerService::isRetryable);
+          }
           throwIfJobNotCompleted(
               createContextResult.getJobReport(), createContextResult.getErrorReport());
 
--- a/src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
+++ b/src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
@@ -246,17 +246,28 @@ public class WorkspaceManagerService {
           StatusEnum status = createContextResult.getJobReport().getStatus();
           if (StatusEnum.FAILED == status) {
             // need to delete the empty workspace before continuing
-            HttpUtils.callWithRetries(
-                () -> workspaceApi.deleteWorkspace(workspaceId),
-                WorkspaceManagerService::isRetryable);
-
+            boolean workspaceSuccessfullyDeleted = false;
+            try {
+              deleteWorkspace(workspaceId);
+              workspaceSuccessfullyDeleted = true;
+            } catch (SystemException ex) {
+              logger.error("Failed to delete workspace {} when cleaning up failed creation of cloud context. "
+                  + "Exception: {}", workspaceApi, ex.getMessage());
+            }
             // if this is a spend profile access denied error, then throw a more user-friendly error
             // message
             if (createContextResult.getErrorReport().getMessage().contains("spend profile")
                 && createContextResult.getErrorReport().getStatusCode()
                     == HttpStatusCodes.STATUS_CODE_FORBIDDEN) {
-              throw new UserActionableException(
-                  "Accessing the spend profile failed. Ask an administrator to grant you access.");
+              final String errorMessage;
+              if (workspaceSuccessfullyDeleted) {
+                errorMessage = "Accessing the spend profile failed. Ask an administrator to grant you access.";
+              } else {
+                errorMessage = String.format("Accessing the spend profile failed. Ask an administrator to grant you access. "
+                + "Additionally, there was a problem cleaning up the workspace, and now workspace ID %s is left behind "
+                + "but won't function.", workspaceId);
+              }
+              throw new UserActionableException(errorMessage);
             }
           }
           // handle non-spend-profile-related failures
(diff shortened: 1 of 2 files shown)


src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
```diff
@@ -242,6 +243,13 @@ public WorkspaceDescription createWorkspace(
                   CREATE_WORKSPACE_MAXIMUM_RETRIES,
                   CREATE_WORKSPACE_DURATION_SLEEP_FOR_RETRY);
           logger.debug("create workspace context result: {}", createContextResult);
+          StatusEnum status = createContextResult.getJobReport().getStatus();
+          if (StatusEnum.FAILED == status || StatusEnum.RUNNING == status) {
```
zloery, line 247: Just to confirm: we're polling the create context call above, so if it's still in `RUNNING` status at this point we consider it failed?

zloery (approved):

src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
jaycarlton: That's what `throwIfJobNotCompleted` does right below. `"CLI timed out waiting for the job to complete. It's still running on the server."`

marikomedlock: I think we should only delete the workspace if the context creation fails, not if it times out. 2 reasons:
- If e.g. RBS gets overloaded and creating a context takes a long time, then we'd have no way of creating workspaces. If instead we just let the job keep running, the user will get a timed out message but can still use the workspace after the job completes.
- The main use case for this change at the moment is when a user has no spend profile access. That's an early step in the create context flight, so I wouldn't expect timeouts.

Longer term, we need to add CLI commands for managing long-running jobs (PF-745), including some kind of callback on success/failure. That would cover the case when the job times out and ultimately fails. For now, let's assume that if the job times out it will ultimately succeed, because I think that will be more common.

marikomedlock (changes requested):
Please also add a test for this change. You may be able to modify an existing method that uses a user without spend profile access `createFailsWithoutSpendAccess`.

src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
jaycarlton: done

src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
```diff
@@ -242,17 +243,23 @@ public WorkspaceDescription createWorkspace(
...
-              && createContextResult.getErrorReport().getStatusCode()
-                  == HttpStatusCodes.STATUS_CODE_FORBIDDEN) {
-            throw new UserActionableException(
-                "Accessing the spend profile failed. Ask an administrator to grant you access.");
+          StatusEnum status = createContextResult.getJobReport().getStatus();
+          if (StatusEnum.FAILED == status) {
+            // need to delete the empty workspace before continuing
+            HttpUtils.callWithRetries(
+                () -> workspaceApi.deleteWorkspace(workspaceId),
+                WorkspaceManagerService::isRetryable);
```
marikomedlock, line 251: I think we should handle failures in this delete call a little differently. I'm thinking that if this workspace delete failed, but the original error was no spend profile access, then it would be confusing to get a "failed to delete workspace" error.

Not sure the best way to implement. Maybe just a `logger.error` for errors here and then fall through to throwing the original exception below?

jaycarlton: Are you thinking we need to put a try/catch around the delete call and swallow the exception and just log it? That would be a bit tricky to read.

marikomedlock: I don't think it matters so much which error/exception you throw -- the one from creation or deletion. As long as both are logged we should be able to debug.

It's more that a custom error message would be helpful e.g. "Workspace creation failed and there was a problem cleaning up. You may see a partially created workspace in your list of available ones.". Otherwise, I'm thinking we'll get reports like "I tried to create a workspace, but got a message about delete failing.", which would be confusing to me as a user.

So yes one approach to doing that could be to `try/catch` around this delete, wrap that exception in a custom error message and then re-throw. Or you could `try/catch` around this delete, log that exception, wrap the original create exception in a custom error message and throw that. I was thinking the latter is slightly better because you're not double-wrapping -- i.e. the `deleteWorkspace` call already wraps in a CLI exception I think.

jaycarlton: I think I have a solution with two error messages, one for the spend profile, and another with that message and some text about the deletion failure.

src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
```diff
@@ -242,17 +243,34 @@ public WorkspaceDescription createWorkspace(
...
-            throw new UserActionableException(
-                "Accessing the spend profile failed. Ask an administrator to grant you access.");
+          StatusEnum status = createContextResult.getJobReport().getStatus();
+          if (StatusEnum.FAILED == status) {
+            // need to delete the empty workspace before continuing
+            boolean workspaceSuccessfullyDeleted = false;
+            try {
+              deleteWorkspace(workspaceId);
+              workspaceSuccessfullyDeleted = true;
+            } catch (SystemException ex) {
```
marikomedlock, line 253: I think you should catch all unchecked exceptions here (i.e. catch `RuntimeException`).

src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
```diff
@@ -242,17 +243,34 @@ public WorkspaceDescription createWorkspace(
...
+              logger.error("Failed to delete workspace {} when cleaning up failed creation of cloud context. "
```
marikomedlock, line 254: `logger.error` optionally takes a `Throwable` as its last argument, which will also log the stack trace. This is helpful in case the message is not helpful or undefined.

src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
```diff
@@ -242,17 +243,34 @@ public WorkspaceDescription createWorkspace(
...
+                  + "Exception: {}", workspaceApi, ex.getMessage());
```
marikomedlock, line 255: Do you want `workspaceId` here instead of `workspaceApi`?

src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
```diff
@@ -242,17 +243,34 @@ public WorkspaceDescription createWorkspace(
...
+            // message
+            if (createContextResult.getErrorReport().getMessage().contains("spend profile")
+                && createContextResult.getErrorReport().getStatusCode()
+                    == HttpStatusCodes.STATUS_CODE_FORBIDDEN) {
+              final String errorMessage;
+              if (workspaceSuccessfullyDeleted) {
+                errorMessage = "Accessing the spend profile failed. Ask an administrator to grant you access.";
+              } else {
+                errorMessage = String.format("Accessing the spend profile failed. Ask an administrator to grant you access. "
+                + "Additionally, there was a problem cleaning up the workspace, and now workspace ID %s is left behind "
```
marikomedlock, line 267: nit: To the user, if I don't have spend profile access, then there shouldn't be a workspace to clean up. Maybe rephrase slightly to "there was a problem cleaning up the partially created workspace", or just "there was a problem cleaning up" or something else?

src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
jaycarlton: good call

marikomedlock (approved):

PR #167, opened 2021-11-16
[PF-1075] use case-insensitive enum flag

jaycarlton:
Allow any case enum options, e.g. `terra workspace list --format=JSon`

### DIFF
--- a/src/test/java/unit/Syntax.java
+++ b/src/test/java/unit/Syntax.java
@@ -1,0 +1,41 @@
+package unit;
+
+import static org.hamcrest.CoreMatchers.containsString;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+import harness.TestCommand;
+import harness.TestCommand.Result;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+/** Tests for syntax variations */
+@Tag("unit")
+public class Syntax {
+
+  @Test
+  @DisplayName("format options are accepted in any case")
+  void formatOptions() {
+    Result resultLowercase = TestCommand.runCommand("status", "--format=json");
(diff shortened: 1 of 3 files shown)


src/test/java/unit/Workspace.java
```diff
@@ -136,10 +136,11 @@ void statusDescribeListReflectUpdate() throws IOException {
     // `terra workspace create --format=json --name=$newName --description=$newDescription`
     String newName = "NEW_statusDescribeListReflectUpdate";
     String newDescription = "NEW status describe list reflect update";
+    // verify mixed case enum works
     TestCommand.runCommandExpectSuccess(
         "workspace",
         "update",
-        "--format=json",
+        "--format=jSON",
```
marikomedlock, line 143: Please add a new test for the case insensitive behavior instead of modifying this existing unrelated one. Perhaps new method(s) in the `unit.Config` class, or a new `unit.Syntax` class altogether.

Suggest including tests for options & parameters, lower/upper/mixed case.

marikomedlock (changes requested):

src/test/java/unit/Workspace.java
jaycarlton: Done.

zloery (approved):

src/test/java/unit/Syntax.java
```diff
@@ -0,0 +1,41 @@
...
+
+import harness.TestCommand;
+import harness.TestCommand.Result;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+/** Tests for syntax variations */
+@Tag("unit")
+public class Syntax {
```
marikomedlock, line 14: This should extend one of the classes in `harness.baseclasses`, probably `ClearContextUnit`.

jaycarlton: done

marikomedlock (approved):

PR #170, opened 2021-11-24
PF-1223: Add  --scan so that the test failure is more readable. 

yuhuyoyo:


marikomedlock (approved):
Thanks for this! Could you add it to the nightly tests also please?

PR #174, opened 2021-12-01
PF-1171: Minor nit improvement for CLI command for BqTable

yuhuyoyo:
Follow up of https://github.com/DataBiosphere/terra-cli/pull/169, addressing Mariko's comments.


src/main/java/bio/terra/cli/businessobject/resource/BigQueryResolveOptions.java
```diff
@@ -0,0 +1,17 @@
+package bio.terra.cli.businessobject.resource;
+
+/** This enum specifies the possible ways to resolve a BigQuery dataset resource. */
+public enum BigQueryResolveOptions {
```
marikomedlock, line 4: nit: `BqResolveOptions` to be consistent with `BqDataset`, `BqTable` classes in the same package.

src/main/java/bio/terra/cli/serialization/userfacing/UFResource.java
```diff
@@ -33,7 +33,7 @@
     property = "resourceType")
 @JsonSubTypes({
   @Type(value = UFGcpNotebook.class, name = "AI_NOTEBOOK"),
-  @Type(value = UFBqDataTable.class, name = "BQ_DATA_TABLE"),
+  @Type(value = UFBqTable.class, name = "BQ_DATA_TABLE"),
```
marikomedlock, line 36: Should this be `BQ_TABLE`?

src/main/java/bio/terra/cli/serialization/userfacing/resource/UFBqTable.java
```diff
@@ -57,12 +56,13 @@ public void print(String prefix) {
...
-    OUT.println(prefix + "BigQuery dataset id: " + datasetId + " table id: " + dataTableId);
+    OUT.println(prefix + "BigQuery dataset id: " + datasetId);
+    OUT.println(prefix + "BigQuery table id: " + dataTableId);
 
     if (tableDescription != null) {
-      OUT.println(prefix + "table description: " + tableDescription);
+      OUT.println(prefix + "Table description: " + tableDescription);
     }
-    OUT.println(prefix + "table has row #: " + numRows.toString());
+    OUT.println(prefix + "# Rows: " + numRows.toString());
```
marikomedlock, line 65: I think we should have nice translations of `null` here (e.g. `(undefined)` or `(unknown)`). You can see an example in the other `UF` resource class for BQ datasets

src/test/java/unit/BqTableReferenced.java
```diff
@@ -429,20 +432,71 @@ void updateMultipleOrNoProperties() throws IOException {
...
+    UFBqTable describeDataTable =
+        TestCommand.runAndParseCommandExpectSuccess(
+            UFBqTable.class, "resource", "describe", "--name=" + name);
+
+    // the external dataset created in the beforeall method should have 1 table in it, but the
+    // sharee user doesn't have read access to the table so they can't know that
+    assertEquals(
+        BigInteger.ZERO,
+        describeDataTable.numRows,
+        "referenced dataset with no access contains NULL numRows");
```
marikomedlock, line 471: The actual assertion is fine, but I don't think this comment and assertion message make sense. Here you're checking that the CLI reports zero rows for a user that does have access.

marikomedlock (commented):
Thank you very much for these updates!

src/main/java/bio/terra/cli/businessobject/resource/BqResolvedOptions.java
```diff
@@ -0,0 +1,17 @@
+package bio.terra.cli.businessobject.resource;
+
+/** This enum specifies the possible ways to resolve a BigQuery resource. */
+public enum BqResolvedOptions {
+  FULL_PATH, // For data table: [project id].[dataset id].[data table id if applicable]
+  TABLE_ID_ONLY, // [data table id]
+  DATASET_ID_ONLY, // [dataset id]
+  PROJECT_ID_ONLY; // [project id]
```
zloery, line 8: I missed the earlier PR so I might not have context but I'm a little surprised here - how do you resolve a BQ resource from just a project ID?

zloery (approved):
LGTM, one question outside the scope of this PR

marikomedlock (approved):

src/main/java/bio/terra/cli/businessobject/resource/BqResolvedOptions.java
marikomedlock: Maybe the command name is a little misleading. `terra resource resolve` returns the cloud identifier for a particular workspace resource. For different resources, we have flags that can tweak the behavior (e.g. `--exclude-bucket-prefix` drops the `gs://` from the bucket url and just gives the bucket name). The solutions team mentioned that sometimes they need the BQ project/dataset id separately, and so we added an option to return just parts of the BQ `FULL_PATH` to make those use cases easier. Before they were getting `[project id].[dataset id]` and parsing it into pieces in order to use it in some places.

PR #179, opened 2021-12-09
PF-1170: Add debug logs to all the commands

yuhuyoyo:


marikomedlock (changes requested):
I think it would be good to log all the arguments also. Ideally, we could do this in a central place (e.g. the `BaseCommand` class that the commands extend) instead of having to modify every command class.

There may be some trickiness around where to get the original list of arguments and where you can log it. The original list of arguments is definitely available in `Main`, which accepts an array of `String`s, but you can't log anything there because the logging levels get setup in the `BaseCommand` class which is called after `Main`.

So I'd first check if you can get the full arg list in `BaseCommand` and log there after the `Logger.setupLogging` line. If you can't do that, then perhaps you can save the list in `Main` and have `BaseCommand` ask `Main` what the arg list is before logging.

PR #184, opened 2021-12-13
[PF-1096] Split client log files by date

gjuggler:
A small quality-of-life improvement for client debugging, this change causes the CLI to create log files with a date-specific pattern. We retain the same per-file and total history filesize limit, as well as a new limit on the total # of archive files to retain.

Before:

Screen Shot 2021-12-13 at 4 26 59 PM

After:

Screen Shot 2021-12-13 at 4 23 49 PM


yuhuyoyo (approved):

zloery (approved):
LGTM, pending Yu's question

PR #185, opened 2021-12-14
PF-1178: support updating reference target.

yuhuyoyo:
Before this change, the resource update command only support updating name and description of a resource. This change allows updating referencing target of a reference resources (e.g. pointing a reference resource from bq dataset A to bq dataset B).

Updating reference to a different resource type is not supported.


src/main/java/bio/terra/cli/serialization/userfacing/input/referenced/UpdateReferencedBqTableParams.java
```diff
@@ -0,0 +1,122 @@
...
+  public BqTable getOriginalResource() {
+    return originalResource;
+  }
+
+  /**
+   * Whether to update the target that the referenced resource is pointing to.
+   *
+   * @return
+   */
+  public boolean updateReferenceTarget() {
```
ddietterich, line 75: That call sounds like it is doing an update. Maybe `shouldUpdateReferenceTarget()`?

src/main/java/bio/terra/cli/businessobject/resource/BqDataset.java
```diff
@@ -112,13 +112,19 @@ public static BqDataset createControlled(CreateBqDatasetParams createParams) {
...
 
   /** Update a BigQuery dataset referenced resource in the workspace. */
-  public void updateReferenced(UpdateResourceParams updateParams) {
-    if (updateParams.name != null) {
-      validateEnvironmentVariableName(updateParams.name);
+  public void updateReferenced(UpdateReferencedBqDatasetParams updateParams) {
+    if (updateParams.getResourceParams().name != null) {
+      validateEnvironmentVariableName(updateParams.getResourceParams().name);
+    }
+    if (updateParams.getProjectId() != null) {
```
ddietterich, line 119: I don't understand the purpose of this test. If `updateParams.getProjectId()` is null, you would be assigning null to the projectId. But isn't the projectId null by construction?

Said another way, why wouldn't you always: `this.projectId = updateParams.getProjectId()` regardless of the null state?

ddietterich (approved):
LGTM - a couple of comments
I don't know the CLI code, so it is not a deep review...

src/main/java/bio/terra/cli/serialization/userfacing/input/referenced/UpdateReferencedBqDatasetParams.java
```diff
@@ -0,0 +1,108 @@
...
+ * Parameters for updating a BigQuery dataset workspace referenced resource. This class is not
+ * currently user-facing, but could be exposed as a command input format in the future.
+ */
+@JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, include = JsonTypeInfo.As.PROPERTY, property = "@class")
+@JsonDeserialize(builder = UpdateResourceParams.Builder.class)
+public class UpdateReferencedBqDatasetParams {
+  private final UpdateResourceParams resourceParams;
+  private final @Nullable String datasetId;
+  private final @Nullable String projectId;
+  private final BqDataset originalResource;
```
marikomedlock, line 20: The `serialization` package classes should all be POJOs, and not include classes from other CLI packages, with the  exception of enums. You can replace this with the individual properties that you need (e.g. `originalProjectId`)

ditto for the other classes that do this (`UpdateReferencedGcsObjectParams` etc.)

src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
```diff
@@ -983,16 +991,30 @@ public void updateControlledGcsBucket(
...
     // convert the CLI object to a WSM request object
-    UpdateDataReferenceRequestBody updateRequest =
-        new UpdateDataReferenceRequestBody()
-            .name(updateParams.name)
-            .description(updateParams.description);
+    UpdateBigQueryDataTableReferenceRequestBody updateRequest =
+        new UpdateBigQueryDataTableReferenceRequestBody()
+            .name(updateParams.getResourceParams().name)
+            .description(updateParams.getResourceParams().description);
+    if (updateParams.updateReferenceTarget()) {
```
marikomedlock, line 1000: nit: I think this method name is slightly misleading, since it's really just checking to see if any resource-specific properties were set. Maybe `hasNewReferenceTargetFields` or something else?

src/test/java/harness/utils/ExternalBQDatasets.java
```diff
@@ -106,6 +113,33 @@ public static void grantWriteAccess(
...
+    bigQuery
+        .tables()
+        .setIamPolicy(
+            projectId,
+            datasetId,
+            tableId,
+            new SetIamPolicyRequest().setPolicy(policy.setBindings(updatedBindings)))
+        .execute();
+    System.out.println(
+        "Grant dataViewer access to table " + tableId + " for proxy group: " + groupEmail);
```
marikomedlock, line 140: nit: "group email" instead of "proxy group", since this is not necessarily a proxy group.

src/test/java/unit/BqDatasetNumTables.java
```diff
@@ -153,5 +176,167 @@ void numTablesForReferencedWithNoAccess() throws IOException {
     // the external dataset created in the beforeall method should have 1 table in it, but the
     // sharee user doesn't have read access to the dataset so they can't know that
     assertNull(describeDataset.numTables, "referenced dataset with no access contains NULL tables");
+
+    // Clean up.
+    workspaceCreator.login();
+    // `terra workspace set --id=$id`
+    TestCommand.runCommandExpectSuccess("workspace", "set", "--id=" + createWorkspace.id);
+    TestCommand.runCommandExpectSuccess("resource", "delete", "--name=" + name, "--quiet");
+    TestCommand.runCommandExpectSuccess("workspace", "delete", "--quiet");
```
marikomedlock, line 185: these tests that create/delete a workspace in each method will leak workspaces if they fail. suggest looking at other classes that create multiple workspaces (e.g. `WorkspaceOverride`). I'm also not sure you need to create a new workspace every time. if the `shareeUser` is the same and has the same role every time, then you should be able to just reuse the same workspace and create resources with names that are unique to each test method.

src/test/java/unit/BqDatasetNumTables.java
```diff
@@ -153,5 +176,167 @@ void numTablesForReferencedWithNoAccess() throws IOException {
...
+            UFBqTable.class,
+            "resource",
+            "update",
+            "bq-table",
+            "--name=" + name,
+            "--new-name=" + newName);
+    assertEquals(newName, updateTable.name);
+
+    TestCommand.runCommandExpectExitCode(
+        1,
```
marikomedlock, line 270: why is this a different exit code that the other failures? perhaps you can assert against the exception message to make that clearer

src/test/java/unit/BqDatasetNumTables.java
```diff
@@ -153,5 +176,167 @@ void numTablesForReferencedWithNoAccess() throws IOException {
...
+    workspaceCreator.login();
+    // `terra workspace set --id=$id`
+    TestCommand.runCommandExpectSuccess("workspace", "set", "--id=" + createWorkspace.id);
+    TestCommand.runCommandExpectSuccess("resource", "delete", "--name=" + name, "--quiet");
+    TestCommand.runCommandExpectSuccess("workspace", "delete", "--quiet");
+  }
+
+  @Test
+  @DisplayName("Attempt to update table reference but the user has no access.")
+  void updateTableReferenceWithNoAccess() throws IOException {
```
marikomedlock, line 190: I don't think these tests for updating the reference belong in this test class `BqDatasetNumTables`. Suggest moving them to a new class e.g. `BqDatasetReferencedUpdate`.

src/test/java/unit/BqDatasetReferenced.java
```diff
@@ -339,6 +348,24 @@ void updateIndividualProperties() throws IOException {
...
+    updateDataset =
+        TestCommand.runAndParseCommandExpectSuccess(
+            UFBqDataset.class,
+            "resource",
+            "update",
+            "bq-dataset",
+            "--name=" + newName,
+            "--new-dataset-id=" + externalDataset2.getDatasetId());
+    assertEquals(externalDataset2.getDatasetId(), updateDataset.datasetId);
+    assertEquals(externalDataset2.getProjectId(), updateDataset.projectId);
```
marikomedlock, line 361: how come the project id also changed -- isn't the command above just updating the dataset id?

src/test/java/unit/BqTableReferenced.java
```diff
@@ -381,6 +398,30 @@ void updateIndividualProperties() throws IOException {
...
+        TestCommand.runAndParseCommandExpectSuccess(
+            UFBqTable.class,
+            "resource",
+            "update",
+            "bq-table",
+            "--name=" + newName,
+            "--new-dataset-id=" + externalDataset2.getDatasetId());
+    assertEquals(externalDataTableName2, updateDataTable.dataTableId);
+    assertEquals(externalDataset2.getDatasetId(), updateDataTable.datasetId);
+    assertEquals(externalDataset2.getProjectId(), updateDataTable.projectId);
```
marikomedlock, line 424: same question as above -- why did the project id here also change, shouldn't it just be the dataset id?

src/test/java/unit/GcsBucketReferenced.java
```diff
@@ -27,26 +30,42 @@
...
     super.setupOnce();
-    externalBucket = ExternalGCSBuckets.createBucketWithUniformAccess();
-    ExternalGCSBuckets.grantReadAccess(externalBucket, Identity.user(workspaceCreator.email));
+    externalSharedBucket = ExternalGCSBuckets.createBucketWithUniformAccess();
+    externalPrivateBucket = ExternalGCSBuckets.createBucketWithUniformAccess();
+
+    ExternalGCSBuckets.grantWriteAccess(
+        externalSharedBucket, Identity.user(workspaceCreator.email));
+    ExternalGCSBuckets.grantReadAccess(
+        externalPrivateBucket, Identity.user(workspaceCreator.email));
```
marikomedlock, line 48: Since you are granting access to the proxy group below, I don't think you need to grant permission to the user's email, since that should be included in their proxy group.

src/test/java/unit/GcsBucketReferenced.java
```diff
@@ -27,26 +30,42 @@
...
+    ExternalGCSBuckets.grantReadAccess(
+        externalSharedBucket, Identity.group(Auth.getProxyGroupEmail()));
   }
 
   @AfterAll
   @Override
   protected void cleanupOnce() throws Exception {
     super.cleanupOnce();
-    ExternalGCSBuckets.deleteBucket(externalBucket);
-    externalBucket = null;
```
marikomedlock, line 49: both buckets should be deleted in cleanup, not just the shared one.

src/test/java/unit/GcsBucketReferenced.java
```diff
@@ -345,5 +391,143 @@ void updateMultipleOrNoProperties() throws IOException {
...
+  }
+
+  @Test
+  @DisplayName(
+      "Attempt to update the gcs bucket while the user only have access to the externalSharedBucket")
+  void updateGcsBucketWithPartialAccess() throws IOException {
+    workspaceCreator.login();
+
+    UFWorkspace createWorkspace =
+        TestCommand.runAndParseCommandExpectSuccess(UFWorkspace.class, "workspace", "create");
```
marikomedlock, line 404: same comment as above. this test will leak workspaces when it fails.

src/test/java/unit/GcsBucketReferenced.java
```diff
@@ -345,5 +391,143 @@ void updateMultipleOrNoProperties() throws IOException {
...
+            UFGcsBucket.class,
+            "resource",
+            "update",
+            "gcs-bucket",
+            "--name=" + name,
+            "--new-name=" + newName);
+    assertEquals(newName, updateBucket.name);
+
+    TestCommand.runCommandExpectExitCode(
+        1,
```
marikomedlock, line 435: does this fail because the user is a READER or because they don't have access to the new bucket?

src/test/java/unit/GcsBucketReferenced.java
```diff
@@ -345,5 +391,143 @@ void updateMultipleOrNoProperties() throws IOException {
...
+    TestCommand.runCommandExpectSuccess("workspace", "delete", "--quiet");
+  }
+
+  @Test
+  @DisplayName(
+      "Attempt to descrone the gcs bucket while the user does not have access to the externalPrivateBucket")
+  void describeBucketReferenceWithoutAccess() throws IOException {
+    workspaceCreator.login();
+
+    UFWorkspace createWorkspace =
```
marikomedlock, line 456: ditto above, don't create workspaces in test methods because they won't be cleaned up on failure. if the different workspace roles are important here (I'm not totally clear on whether the READER/WRITER distinction is what's causing failures or the private/shared bucket access is), then I suggest making a new test class that creates 2 workspaces in setup (e.g. `workspaceSharedWithReader`, `workspaceSharedWithWriter`) and shares them with a user.

src/test/java/unit/GcsBucketReferenced.java
```diff
@@ -345,5 +391,143 @@ void updateMultipleOrNoProperties() throws IOException {
...
+    TestCommand.runCommandExpectSuccess("resource", "delete", "--name=" + newName, "--quiet");
+    workspaceCreator.login();
+    // `terra workspace set --id=$id`
+    TestCommand.runCommandExpectSuccess("workspace", "set", "--id=" + createWorkspace.id);
+    TestCommand.runCommandExpectSuccess("workspace", "delete", "--quiet");
+  }
+
+  @Test
+  @DisplayName(
+      "Attempt to descrone the gcs bucket while the user does not have access to the externalPrivateBucket")
```
marikomedlock, line 452: Is there already a test for this case (e.g. describe succeeds even though the user doesn't have access to the underlying resource)? Maybe check `GcsBucketNumObjects` to see if you can modify an existing test instead.

src/test/java/unit/GcsBucketReferenced.java
```diff
@@ -345,5 +391,143 @@ void updateMultipleOrNoProperties() throws IOException {
...
+    TestCommand.runCommandExpectSuccess("workspace", "set", "--id=" + createWorkspace.id);
+    TestCommand.runCommandExpectSuccess("workspace", "delete", "--quiet");
+  }
+
+  @Test
+  @DisplayName(
+      "Attempt to add reference to buckets while the user only have access to externalSharedbucket")
+  void addBucketRefWithPartialAccess() throws IOException {
+    workspaceCreator.login();
+    UFWorkspace createWorkspace =
```
marikomedlock, line 503: ditto above, don't create workspaces in test methods

src/main/java/bio/terra/cli/serialization/userfacing/input/controlled/CreateBqDatasetParams.java
```diff
@@ -1,5 +1,6 @@
-package bio.terra.cli.serialization.userfacing.input;
+package bio.terra.cli.serialization.userfacing.input.controlled;
```
marikomedlock, line 1: This package reorganization means we now have 2 indications about whether a class pertains to a controlled/referenced resource: the package name (`input.controlled` or `input.referenced`) and the class name (`Create...Params` or `Add...Params`). I don't think this repetition is useful. Either we should differentiate by class name, as it already is I think, or reorganize into separate packages and remove the class name prefixes.

src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
```diff
@@ -983,16 +991,30 @@ public void updateControlledGcsBucket(
...
+      updateRequest.resourceAttributes(
+          new GcpBigQueryDataTableAttributes()
+              .projectId(
+                  Optional.ofNullable(updateParams.getProjectId())
+                      .orElse(updateParams.getOriginalResource().getProjectId()))
+              .datasetId(
+                  Optional.ofNullable(updateParams.getDatasetId())
+                      .orElse(updateParams.getOriginalResource().getDatasetId()))
```
marikomedlock, line 1008: I think WSM API should be changed to accept `null` to mean no-change for the update endpoints. How does WSM team feel about making this change, or is there already a ticket to do that?

marikomedlock (changes requested):

src/test/java/unit/BqDatasetReferenced.java
yuhuyoyo: it didn't change. externalDataset1 and 2 are both in the same project. just asserting it here to make sure that by not passing projectId here, the updateDatset don't have projectId null.

src/test/java/unit/BqDatasetNumTables.java
yuhuyoyo: in some test, the sharee user needs to have write access to perform add-ref and update. but if i moved it to another class, this won't be an issue.

src/main/java/bio/terra/cli/businessobject/resource/BqDataset.java
yuhuyoyo: because this is used for updatereference. if the user doesn't specify an update param, we don't want to override the existing param to null.

src/test/java/unit/BqTableReferenced.java
yuhuyoyo: same as above.

src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
yuhuyoyo: filed a ticket: https://broadworkbench.atlassian.net/browse/PF-1271

src/main/java/bio/terra/cli/serialization/userfacing/input/UpdateReferencedBqDatasetParams.java
```diff
@@ -0,0 +1,111 @@
...
+import javax.annotation.Nullable;
+
+/**
+ * Parameters for updating a BigQuery dataset workspace referenced resource. This class is not
+ * currently user-facing, but could be exposed as a command input format in the future.
+ */
+@JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, include = JsonTypeInfo.As.PROPERTY, property = "@class")
+@JsonDeserialize(builder = UpdateResourceParams.Builder.class)
+public class UpdateReferencedBqDatasetParams {
+  private final UpdateResourceParams resourceParams;
```
marikomedlock, line 15: nit: The other params classes have `public final` fields instead of getters, which makes the class a little less verbose. It would be nice to be consistent in this class. Ditto for the other `Update*Params` classes below.

src/main/java/bio/terra/cli/command/resource/update/GcsObject.java
```diff
@@ -54,7 +54,8 @@ protected void execute() {
             .resourceFields(updateResourceParams)
             .bucketName(newBucketName.getNewBucketName())
             .objectName(newObjectName)
-            .originalResource(resource)
+            .originalObjectName(resource.getObjectName())
```
marikomedlock, line 57: nit: Add a `TODO (PF-...)` comment here about removing the original value fields in the future. Ditto for the other update command classes.

src/test/java/unit/BqDatasetNumTables.java
```diff
@@ -115,8 +117,10 @@ void numTablesForReferenced() throws IOException {
             "--project-id=" + externalDataset.getProjectId(),
             "--dataset-id=" + externalDataset.getDatasetId());
 
-    // the external dataset created in the beforeall method should have 1 table in it
+    // the external dataset created in the beforeall method should have 2 table in it
```
marikomedlock, line 120: nit: comment looks incorrect based on assertion below.

src/test/java/unit/BqDatasetNumTables.java
```diff
@@ -24,8 +24,8 @@
   // external dataset to use for creating BQ dataset references in the workspace
   private DatasetReference externalDataset;
 
-  // name of table in external dataset
-  private String externalDatasetTableName = "testTable";
+  // name of tables in external dataset
+  private String sharedExternalTable = "testTable2";
```
marikomedlock, line 28: nit: does "shared" mean anything specific here? or is this just leftover from a test restructuring?

src/test/java/unit/BqDatasetReferenced.java
```diff
@@ -385,14 +408,20 @@ void updateMultipleOrNoProperties() throws IOException {
             "bq-dataset",
             "--name=" + name,
             "--new-name=" + newName,
-            "--description=" + newDescription);
+            "--description=" + newDescription,
+            "--new-project-id=" + externalDataset2.getProjectId(),
```
marikomedlock, line 412: is this actually changing the project id? it may be difficult to test that case with the `ExternalBqDatasets` utilities, since those will always create resources in the same project, but you could e.g. create a controlled bq dataset in a workspace (even in this one) and then update the reference to that, which should have a different project id than the "external" project we use. suggest doing that in a separate test method if you do take that approach.

src/test/java/unit/BqTableReferenced.java
```diff
@@ -427,9 +468,15 @@ void updateMultipleOrNoProperties() throws IOException {
             "bq-table",
             "--name=" + name,
             "--new-name=" + newName,
-            "--description=" + newDescription);
+            "--description=" + newDescription,
+            "--new-project-id=" + externalDataset2.getProjectId(),
```
marikomedlock, line 472: same comment as above about whether this project id is actually changing here?

src/test/java/unit/BqTableReferencedUpdate.java
```diff
@@ -0,0 +1,199 @@
...
+        privateExternalTable);
+    ExternalBQDatasets.createTable(
+        workspaceCreator.getCredentialsWithCloudPlatformScope(),
+        externalDataset.getProjectId(),
+        externalDataset.getDatasetId(),
+        sharedExternalTable);
+
+    // `terra workspace set --id=$id`
+    TestCommand.runCommandExpectSuccess("workspace", "set", "--id=" + getWorkspaceId());
+    // `terra workspace add-user --email=$email --role=READER`
```
marikomedlock, line 54: nit: comment doesn't match below

marikomedlock (approved):
This looks great, thanks! Some nit comments, otherwise looks good. Also can you please update the README with your new commands and this PR description?

src/test/java/unit/BqDatasetReferenced.java
yuhuyoyo: the two project has the same id. i added line 412 just because the test is to specify all the attributes.

yuhuyoyo: i took the approach you suggested in this test because this test is supposed to test for changing all the attributes. same for the table reference test below.

PR #186, opened 2022-01-06
PF-1274: Make resource name field as not required

yuhuyoyo:
WSM allows not specifying resource name when creating controlled resources in PF-1201. This PR is a follow-up on that. CLI is already auto-generating an instance-id for a controlled notebook resources so no change is needed for the notebook resource.

### DIFF
--- a/src/main/java/bio/terra/cli/command/resource/create/BqDataset.java
+++ b/src/main/java/bio/terra/cli/command/resource/create/BqDataset.java
@@ -20,7 +20,7 @@ public class BqDataset extends BaseCommand {
   @CommandLine.Mixin ControlledResourceCreation controlledResourceCreationOptions;
   @CommandLine.Mixin BqDatasetLifetime bqDatasetLifetimeOptions;
 
-  @CommandLine.Option(names = "--dataset-id", required = true, description = "BigQuery dataset id.")
+  @CommandLine.Option(names = "--dataset-id", description = "BigQuery dataset id.")
   private String bigQueryDatasetId;
 
   @CommandLine.Option(
--- a/src/main/java/bio/terra/cli/command/resource/create/BqDataset.java
+++ b/src/main/java/bio/terra/cli/command/resource/create/BqDataset.java
@@ -20,7 +20,10 @@ public class BqDataset extends BaseCommand {
   @CommandLine.Mixin ControlledResourceCreation controlledResourceCreationOptions;
   @CommandLine.Mixin BqDatasetLifetime bqDatasetLifetimeOptions;
 
-  @CommandLine.Option(names = "--dataset-id", description = "BigQuery dataset id.")
+  @CommandLine.Option(
+      names = "--dataset-id",
+      description =
+          "BigQuery dataset id. If not provided, the resource name will be used as the dataset id.")
   private String bigQueryDatasetId;
 
   @CommandLine.Option(
--- a/src/main/java/bio/terra/cli/command/resource/create/GcpNotebook.java
+++ b/src/main/java/bio/terra/cli/command/resource/create/GcpNotebook.java
@@ -1,7 +1,6 @@
 package bio.terra.cli.command.resource.create;
 
 import bio.terra.cli.businessobject.Context;
-import bio.terra.cli.businessobject.User;
 import bio.terra.cli.command.shared.BaseCommand;
 import bio.terra.cli.command.shared.options.Format;
 import bio.terra.cli.command.shared.options.ResourceCreation;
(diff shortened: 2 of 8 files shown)


src/main/java/bio/terra/cli/command/resource/create/BqDataset.java
```diff
@@ -20,7 +20,7 @@
   @CommandLine.Mixin ControlledResourceCreation controlledResourceCreationOptions;
   @CommandLine.Mixin BqDatasetLifetime bqDatasetLifetimeOptions;
 
-  @CommandLine.Option(names = "--dataset-id", required = true, description = "BigQuery dataset id.")
+  @CommandLine.Option(names = "--dataset-id", description = "BigQuery dataset id.")
```
marikomedlock, line 23: nit: include a note here about how we will generate a name if not provided, similar to the `create gcp-notebook` command. ditto other command classes

src/main/java/bio/terra/cli/serialization/userfacing/input/CreateBqDatasetParams.java
```diff
@@ -13,7 +14,7 @@
 public class CreateBqDatasetParams {
   public final CreateResourceParams resourceFields;
   public final String projectId;
-  public final String datasetId;
+  public final @Nullable String datasetId;
```
marikomedlock, line 17: nit: I don't think you need to include `@Nullable` in these serialization classes, because all fields are nullable. ditto below. alternatively, we could always include these annotations in the serialization classes as well, but I think it should be consistent. it seems easier to just maintain these requirements on the `businessobject` package classes instead of also on their serialization POJO counterparts.

marikomedlock (commented):
RE notebook default naming: Would it make sense to move the name auto-generation logic from the CLI into WSM so that the UI can also use it? (doesn't have to be this PR)

zloery (approved):
LGTM

yuhuyoyo:
about the notebook, WSM does generate a notebook name in the format of terra-uuid-notebook if an instanceId is not provided. This is different from the auto generation logic terra CLI have. Is it OK that I just remove the default auto gen name from the CLI logic?

marikomedlock:
> about the notebook, WSM does generate a notebook name in the format of terra-uuid-notebook if an instanceId is not provided. This is different from the auto generation logic terra CLI have. Is it OK that I just remove the default auto gen name from the CLI logic?

I don't know what the desired behavior is (cli logic or wsm logic or something else) -- suggest checking with the solutions team for their opinion. But I was just thinking we may want the same auto-generation behavior for all clients, in which case it would make sense to put whatever that is in wsm rather than implementing it once per client.

yuhuyoyo:
> > about the notebook, WSM does generate a notebook name in the format of terra-uuid-notebook if an instanceId is not provided. This is different from the auto generation logic terra CLI have. Is it OK that I just remove the default auto gen name from the CLI logic?
> 
> I don't know what the desired behavior is (cli logic or wsm logic or something else) -- suggest checking with the solutions team for their opinion. But I was just thinking we may want the same auto-generation behavior for all clients, in which case it would make sense to put whatever that is in wsm rather than implementing it once per client.

Removed the auto-gen logic here so we can just rely on the one in WSM

src/main/java/bio/terra/cli/command/resource/create/GcpNotebook.java
```diff
@@ -47,8 +39,7 @@
               + "The instance name must be 1 to 63 characters long and contain only lowercase "
               + "letters, numeric characters, and dashes. The first character must be a lowercase "
               + "letter and the last character cannot be a dash. If not specified, an "
-              + "auto-generated name based on your email address and time will be used.",
-      defaultValue = AUTO_GENERATE_NAME)
+              + "auto-generated name based on your email address and time will be used.")
```
marikomedlock, line 42: Is this true? you earlier comment said that wsm generates an instance id like terra-uuid-notebook.

Also, is the solutions team ok with this change in behavior? Will wrote this logic and I don't know whether he did that based on input from them or not.

yuhuyoyo: i'm copying this logic to WSM side in a separate PR.

marikomedlock: ah ok. are you planning to merge the wsm PR before this one then? or I guess you could move the notebook part of this to a separate PR and just merge the other stuff until the wsm change is ready.

marikomedlock (approved):
LGTM. per comment, consider whether to merge the notebook changes before/after the accompanying wsm changes.

src/main/java/bio/terra/cli/command/resource/create/GcpNotebook.java
yuhuyoyo: https://github.com/DataBiosphere/terra-workspace-manager/pull/525 I merged the WSM change. I will update the wsm version of this PR in the gradle.

PR #188, opened 2022-01-14
PF-1271: Update WorkspaceManagerService API call based on WSM API change

yuhuyoyo:
WSM no longer requires the resource attributes field when updating targeting reference. 

e.g. when updating bucket object from foo to bar within the same object, the bucket name field can be set to null.


marikomedlock (approved):
Looks good, thanks!

zloery (approved):

PR #189, opened 2022-01-14
[PF-1253] Invalidate refresh/access token when user runs terra auth revoke

melissachang:
Before this PR: After `terra auth revoke`, still able to make calls using access token, such as calling WSM api/workspaces/v1 via swagger

After this PR: After `terra auth revoke`, calling api/workspaces/v1 returns 401. Note that it can take up to a minute for the revoke to take effect.

### DIFF
--- a/src/main/java/bio/terra/cli/service/GoogleOauth.java
+++ b/src/main/java/bio/terra/cli/service/GoogleOauth.java
@@ -284,7 +284,7 @@ public final class GoogleOauth {
     try {
       HttpUtils.sendHttpRequest("https://oauth2.googleapis.com/revoke", "POST", headers, params);
     } catch (IOException ioEx) {
-      throw new SystemException("Unable to revoke token");
+      throw new SystemException("Unable to revoke token", ioEx);
     }
   }
 }
--- a/src/main/java/bio/terra/cli/service/GoogleOauth.java
+++ b/src/main/java/bio/terra/cli/service/GoogleOauth.java
@@ -1,5 +1,7 @@
 package bio.terra.cli.service;
 
+import bio.terra.cli.exception.SystemException;
+import bio.terra.cli.service.utils.HttpUtils;
 import bio.terra.cli.utils.UserIO;
 import com.google.api.client.auth.oauth2.Credential;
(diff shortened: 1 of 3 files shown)


src/main/java/bio/terra/cli/service/GoogleOauth.java
```diff
@@ -261,4 +265,26 @@ public static AccessToken getAccessToken(GoogleCredentials credential) {
...
+    String endpoint = "https://oauth2.googleapis.com/revoke";
+    Map<String, String> headers =
+        ImmutableMap.of("Content-type", "application/x-www-form-urlencoded");
+    Map<String, String> params =
+        ImmutableMap.of("token", credential.getAccessToken().getTokenValue());
+
+    try {
+      HttpUtils.sendHttpRequest("https://oauth2.googleapis.com/revoke", "POST", headers, params);
+    } catch (IOException ioEx) {
+      throw new SystemException("Unable to revoke token");
```
marikomedlock, line 287: nit: wrap the `ioEx` in the `SystemException`

src/main/java/bio/terra/cli/service/utils/HttpUtils.java
```diff
@@ -54,6 +55,27 @@ private HttpUtils() {}
   public static HttpResponse sendHttpRequest(
       String urlStr, String requestType, String accessToken, Map<String, String> params)
       throws IOException {
+    Map<String, String> headers = ImmutableMap.of("accept", "*/*");
+    if (accessToken != null) {
+      headers.put("Authorization", "Bearer " + accessToken);
+    }
```
marikomedlock, line 61: I think this needs to be a mutable map in order to do a `put`, or you can use a builder.

melissachang: Oops, done :)

src/main/java/bio/terra/cli/service/GoogleOauth.java
melissachang: Done

jmczerk (approved):

marikomedlock (approved):

PR #191, opened 2022-01-24
[PF-1269] installation without Docker support

jaycarlton:
Since we have a config option to run apps locally instead of in the docker container, we technically can install the CLI without docker. This changes mainly the bash installation scripts to make this possible.

Also changed some bash code based on IDE suggestions, primarily around double quoting expanded variables.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -38,10 +38,13 @@
 -----
 
 ### Setup development environment
-From the top-level directory:
+The TERRA_CLI_DOCKER_MODE environment variable controls Docker support. Set it to
+   DOCKER_NOT_AVAILABLE (default) to skip pulling the Docker image
+   or DOCKER_AVAILABLE to pull the image (requires Docker to be installed and running).
+Then, from the top-level directory, run:
 ```
 source tools/local-dev.sh
-terra
+build/install/terra-cli/bin/terra
 ```
 
 #### Logging
@@ -157,6 +160,9 @@ CLI installation on the same machine.
 
 - Run a single test by specifying the `--tests` option:
 `./gradlew runTestsWithTag -PtestTag=unit --tests "unit.Workspace.createFailsWithoutSpendAccess" --info`
+
+#### Docker and Tests
+The tests require the Docker daemon to be running (install mode DOCKER_AVAILABLE).
 
 #### Override default server
 The tests run against the `broad-dev` server by default. You can run them against a different server
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,6 +10,7 @@
 3. [Testing](#testing)
     * [Two types of tests](#two-types-of-tests)
     * [Run tests](#run-tests)
+    * [Docker and Tests](#docker-and-tests)
     * [Override default server](#override-default-server)
     * [Override default Docker image](#override-default-docker-image)
     * [Override context directory](#override-context-directory)
@@ -44,7 +45,7 @@ The TERRA_CLI_DOCKER_MODE environment variable controls Docker support. Set it t
 Then, from the top-level directory, run:
 ```
 source tools/local-dev.sh
-build/install/terra-cli/bin/terra
+terra
 ```
 
 #### Logging
--- a/README.md
+++ b/README.md
@@ -387,10 +387,16 @@ For example, a bucket within the workspace Google project. You can create these 
 
 A referenced resource is a cloud resource that is NOT managed by Terra. It exists outside the current workspace
 context. For example, a BigQuery dataset hosted outside of Terra or in another workspace. You can add these with the
-`add-ref` command.
+`add-ref` command. The workspace currently supports the following referenced resource: 
+- `gcs-bucket`
+- `gcs-object`
+- `bq-dataset`
+- `bq-table`
+- `git-repo`
 
 The `check-access` command lets you see whether you have access to a particular resource. This is useful when a
-different user created or added the resource and subsequently shared the workspace with you.
+different user created or added the resource and subsequently shared the workspace with you. `check-access` currently always 
+returns true for `git-repo` reference type because workspace doesn't support authentication to external git services yet.
 
 The list of resources in a workspace is maintained on the Terra Workspace Manager server. The CLI caches this list
 of resources locally. Third-party tools can access resource details via environment variables (e.g. $TERRA_mybucket
--- a/README.md
+++ b/README.md
@@ -48,6 +48,11 @@ curl -L https://github.com/DataBiosphere/terra-cli/releases/latest/download/down
 ./terra
 ```
 
+By default, the CLI will be installed without support for Docker (i.e. it won't pull the Docker image).
+The TERRA_CLI_DOCKER_MODE environment variable controls Docker support. Set it to
+* DOCKER_NOT_AVAILABLE (default) to skip pulling the Docker image
+* DOCKER_AVAILABLE to pull the image (requires Docker to be installed and running).
+
 This will install the Terra CLI in the current directory. Afterwards, you may want to add it to your `$PATH` directly
 or move it to a place that is already on your `$PATH` (e.g. `/usr/local/bin`).
 
@@ -56,7 +61,7 @@ modify the `$PATH`. So if you have added it to your `$PATH`, that step needs to 
 
 #### Requirements
 1. Java 11
-2. Docker 20.10.2 (Must be running)
+2. Docker 20.10.2 (Must be running if installing in DOCKER_AVAILABLE mode)
 3. `curl`, `tar`, `gcloud` (For install only)
 
 Note: The CLI doesn't use `gcloud` directly either during install or normal operation.
--- a/README.md
+++ b/README.md
@@ -387,16 +387,10 @@ For example, a bucket within the workspace Google project. You can create these 
 
 A referenced resource is a cloud resource that is NOT managed by Terra. It exists outside the current workspace
 context. For example, a BigQuery dataset hosted outside of Terra or in another workspace. You can add these with the
-`add-ref` command. The workspace currently supports the following referenced resource: 
-- `gcs-bucket`
-- `gcs-object`
-- `bq-dataset`
-- `bq-table`
-- `git-repo`
+`add-ref` command.
 
 The `check-access` command lets you see whether you have access to a particular resource. This is useful when a
-different user created or added the resource and subsequently shared the workspace with you. `check-access` currently always 
-returns true for `git-repo` reference type because workspace doesn't support authentication to external git services yet.
+different user created or added the resource and subsequently shared the workspace with you.
 
 The list of resources in a workspace is maintained on the Terra Workspace Manager server. The CLI caches this list
 of resources locally. Third-party tools can access resource details via environment variables (e.g. $TERRA_mybucket
--- a/README.md
+++ b/README.md
@@ -150,7 +150,9 @@ gcloud auth login
 gcloud auth application-default login
 ```
 
-Run a Nextflow hello world example.
+#### Nextflow Examples
+Run a Nextflow hello world example (requires Docker image set and container running, or Nextflow to
+be installed locally. For Docker support, `export TERRA_CLI_DOCKER_MODE=DOCKER_AVAILABLE` before installing `terra`):
 ```
 terra nextflow run hello
 ```
@@ -208,18 +210,23 @@ access token. Then specify the `-with-tower` flag when kicking off the workflow.
 - Call the Gcloud CLI tools in the current workspace context.
 This means that Gcloud is configured with the backing Google project and environment variables are defined that
 contain workspace and resource properties (e.g. bucket names, pet service account email).
-```
-terra gcloud config get-value project
-terra gsutil ls
-terra bq version
-```
+    ```
+    terra gcloud config get-value project
+    terra gsutil ls
+    terra bq version
+    ```
 
 - See the list of supported third-party tools.
-The CLI runs these tools in a Docker image. Print the image tag that the CLI is currently using.
-```
-terra app list
-terra config get image
-```
+The CLI runs these tools in a Docker image, if `app-launch` mode is `DOCKER_CONTAINER`. If the
+`app-launch` mode is `LOCAL_PROCESS`, the CLI will assume the tools are available in the current
+shell environment and launch them there.
+    ```
+    terra app list
+    ```
+- Print the image tag that the CLI is currently using.
+    ```
+    terra config get image
+    ```
 
 ### Commands description
 ```
--- a/src/main/java/bio/terra/cli/app/utils/DockerClientWrapper.java
+++ b/src/main/java/bio/terra/cli/app/utils/DockerClientWrapper.java
@@ -243,11 +243,12 @@ public class DockerClientWrapper {
         ex.getCause() != null
             && ex.getCause() instanceof IOException
             && ex.getCause().getMessage() != null
-            && (ex.getCause().getMessage().contains("Connection refused")
+            && (ex.getCause().getMessage().toLowerCase().contains("connection refused")
                 || ex.getCause().getMessage().contains("native connect() failed"));
     if (isDockerConnectionFailed) {
       return new UserActionableException(
-          "Connecting to Docker daemon failed. Check that Docker is installed and running.", ex);
+          "Connecting to Docker daemon failed. Check that Docker is installed and running. "
+              + "To run apps without Docker, use the LOCAL_PROCESS app launch mode (terra config set app-launch LOCAL_PROCESS).", ex);
     } else {
       return ex;
     }
--- a/src/main/java/bio/terra/cli/app/utils/DockerClientWrapper.java
+++ b/src/main/java/bio/terra/cli/app/utils/DockerClientWrapper.java
@@ -248,7 +248,8 @@ public class DockerClientWrapper {
     if (isDockerConnectionFailed) {
       return new UserActionableException(
           "Connecting to Docker daemon failed. Check that Docker is installed and running. "
-              + "To run apps without Docker, use the LOCAL_PROCESS app launch mode (terra config set app-launch LOCAL_PROCESS).", ex);
+              + "To run apps without Docker, use the LOCAL_PROCESS app launch mode (terra config set app-launch LOCAL_PROCESS).",
+          ex);
     } else {
       return ex;
     }
--- a/src/main/java/bio/terra/cli/command/config/set/AppLaunch.java
+++ b/src/main/java/bio/terra/cli/command/config/set/AppLaunch.java
@@ -1,6 +1,8 @@
 package bio.terra.cli.command.config.set;
 
+import bio.terra.cli.app.utils.DockerClientWrapper;
 import bio.terra.cli.businessobject.Config;
+import bio.terra.cli.businessobject.Config.CommandRunnerOption;
 import bio.terra.cli.businessobject.Context;
 import bio.terra.cli.command.shared.BaseCommand;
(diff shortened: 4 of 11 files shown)


tools/install.sh
```diff
@@ -4,6 +4,13 @@ set -e
 ## Dependencies: docker, gcloud
```
jaycarlton, line 4: This file is WIP

src/main/java/bio/terra/cli/app/utils/DockerClientWrapper.java
```diff
@@ -243,11 +243,12 @@ private RuntimeException wrapExceptionIfDockerConnectionFailed(RuntimeException
...
             && ex.getCause() instanceof IOException
             && ex.getCause().getMessage() != null
-            && (ex.getCause().getMessage().contains("Connection refused")
+            && (ex.getCause().getMessage().toLowerCase().contains("connection refused")
                 || ex.getCause().getMessage().contains("native connect() failed"));
     if (isDockerConnectionFailed) {
       return new UserActionableException(
-          "Connecting to Docker daemon failed. Check that Docker is installed and running.", ex);
+          "Connecting to Docker daemon failed. Check that Docker is installed and running. "
+              + "To run apps without Docker, use the LOCAL_PROCESS app launch mode (terra config set app-launch LOCAL_PROCESS).", ex);
```
jaycarlton, line 251: The original plan was to put a check in the config set code for app-launch that would block the user from picking docker when docker wasn't available. I like this change better because it's less intrusive.

marikomedlock (commented):
- Regardless of whether the user sets the new env var or not, they can always toggle the app-launch config property post-install. Currently, when someone updates the docker image id config property, we check whether the image exists on the machine. Perhaps you can add this same check to the app-launch config property toggle. And make both output something to the user (e.g. "image not found. do docker pull ___ or use the local-process option"). I think now it just logs whether the image exists or not, there's no output to the user.

tools/install.sh
```diff
@@ -4,6 +4,13 @@ set -e
...
 ## Usage: ./install.sh
 
+echo "--  Determining installation mode"
+if [[ -z "$TERRA_CLI_INSTALLATION_MODE" ]]; then
+  terraCliInstallationMode="SUPPORT_DOCKER"
+else
+  terraCliInstallationMode="LOCAL_ONLY"
+fi
```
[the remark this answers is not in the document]
jaycarlton: Yeah. Technically there might be people who have docker but prefer not to use it, and could still use the not available mode to save time during the installation, but it's probably an over-optimization to handle that case.

tools/install.sh
```diff
@@ -1,9 +1,23 @@
...
 set -e
 ## This script installs the Terra CLI from an unarchived release directory.
-## Dependencies: docker, gcloud
+## TERRA_CLI_INSTALLATION_MODE environment variable controls docker support. Set to
+##     SUPPORT_DOCKER (default) or LOCAL_ONLY to skip docker steps.
+# # Dependencies: docker, gcloud
 ## Usage: ./install.sh
 
+echo "--  Determining Docker availability mode"
+if [ -z "$TERRA_CLI_DOCKER_MODE" ] || [ "$TERRA_CLI_DOCKER_MODE" == "DOCKER_AVAILABLE" ]; then
```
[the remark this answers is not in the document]
marikomedlock: The main motivation for packaging a Docker image with the CLI was to reduce the installation burden on users who wanted to run apps (specifically Nextflow) in the context of a workspace. So you could just install the Terra CLI, and run `terra nextflow ...` without having to install a "good" version of Nextflow yourself.

It is possible to run `terra nextflow ...` with a local installation of Nextflow. That was the purpose of the `LOCAL_PROCESS` mode of launching apps, which was originally added as a back-door/troubleshooting option for when:
- users already had something installed locally and wanted to use their version, not our packaged one (e.g. an edge release of nextflow)
- the docker container environment presented problems (e.g. nextflow tries to run a container, and docker in docker doesn't really work)

So Iong way of saying, users can still run `terra nextflow ...` without Docker, but the installation burden is larger. Previously, I think it made more sense to reduce the installation steps (i.e. by not requiring separate install of `nextflow` and other tools). Now that there may be more users without docker installed, maybe the default should change, but that's more of a product question -- engineering-wise, either is fine.

If we do change the default, then a couple suggestions:
- consider adding some version checking for apps. e.g. for Nextflow, they will need a relatively recent release in order to run workflows in an enterprise terra workspace.
- update the docs & example usage to include links to the required third-party apps installation steps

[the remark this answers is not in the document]
jaycarlton: After discussion, decision for now is to make non-docker the default installation experience, and to take in an env var to have the installer pull the docker image.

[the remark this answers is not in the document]
jaycarlton: Filed https://broadworkbench.atlassian.net/browse/PF-1316 for third party app version checks and https://broadworkbench.atlassian.net/browse/PF-1317 for documenting their installation steps.

tools/install.sh
```diff
@@ -13,37 +26,42 @@ fi
...
+mkdir -p "$applicationDir"
 if [ ! -d "${applicationDir}" ]; then
     echo "Error creating application directory: ${applicationDir}"
     exit 1
 fi
 
 echo "--  Moving JARs to application directory"
 if [ -f "${applicationDir}/lib" ] || [ -d "${applicationDir}/lib" ]; then
-  rm -R "${applicationDir}/lib"
+  rm -R "${applicationDir:?}/lib"
```
marikomedlock, line 37: I think this will print an error if `applicationDir` isn't set, but I see that variable is defined above. Did you see some error with this line?

tools/install.sh
```diff
@@ -1,9 +1,22 @@
 #!/bin/bash
 set -e
 ## This script installs the Terra CLI from an unarchived release directory.
-## Dependencies: docker, gcloud
+## TERRA_CLI_DOCKER_MODE environment variable controls docker support. Set to
+##     SUPPORT_DOCKER (default) or LOCAL_ONLY to skip docker steps.
```
marikomedlock, line 5: Comment doesn't match the `DOCKER_NOT_AVAILABLE`, `DOCKER_AVAILABLE` literals below, or the default value.

tools/local-dev.sh
```diff
@@ -6,27 +6,43 @@ set -e
 ## Usage: source tools/local-dev.sh
```
marikomedlock, line 6: Please include a comment here about the env var, similar to the one in `install.sh`

tools/local-dev.sh
```diff
@@ -6,27 +6,43 @@ set -e
...
-defaultDockerImage=$(terra config get image)
-docker pull $defaultDockerImage
+alias terra="$(pwd)"/build/install/terra-cli/bin/terra
+
+if [ "$terraCliInstallationMode" == "DOCKER_NOT_AVAILABLE" ]; then
+  echo "Installing without docker image because TERRA_CLI_DOCKER_MODE is DOCKER_NOT_AVAILABLE."
+  terra config set app-launch LOCAL_PROCESS
+else
+  echo "Setting the Docker image id to the default"
+  terra config set image --default
```
marikomedlock, line 36: Set the `app-launch` mode here to `DOCKER_CONTAINER`.

tools/install.sh
```diff
@@ -13,37 +26,42 @@ fi
 
 echo "--  Checking for application directory"
 applicationDir="${HOME}/.terra"
-mkdir -p $applicationDir
+mkdir -p "$applicationDir"
```
cbookg, line 29: in general it would nicer to make all these formatting changes separately, just to make it easier to review (and revert if needed) the logic change.

cbookg (approved):
LGTM, though i suspect we'll want to follow up with:
- Does the CLI properly display a good error if the local tools is not installed.
- Fix the CLI to not mangle the user's gcloud config, and instead pass terra config as an argument (or something less intrusive).

marikomedlock (commented):
I think this may cause integration test failures if merged. Currently we run 3 types of tests:
- Unit tests, which always pull the docker image. These should continue to pass.
- Integration tests against an install from source (`./gradlew install`). These should also continue to pass because we pull the default docker image [afterwards](https://github.com/DataBiosphere/terra-cli/blob/2546f880b8599bab32e58ee4b4eab9cc411c898a/tools/install-for-testing.sh#L27).
- Integration tests against the latest release downloaded from GH. These I think will fail once your changes to the default `app-launch` property are part of the latest release.

You can test this hypothesis by:
- [publishing](https://github.com/DataBiosphere/terra-cli/blob/main/CONTRIBUTING.md#publish-a-release) a pre-release
- changing the [test install script](https://github.com/DataBiosphere/terra-cli/blob/2546f880b8599bab32e58ee4b4eab9cc411c898a/tools/install-for-testing.sh#L35) to install your pre-release instead of the latest release
- [kicking off](https://github.com/DataBiosphere/terra-cli/blob/main/CONTRIBUTING.md#run-tests) a run of the integration tests against an installation from your GH release (i.e. with the `-PtestInstallFromGitHub` flag)

If this does cause those tests to fail, then I think there are a couple options:
- Just change the test install script to set the `DOCKER_AVAILABLE` env var before testing. This is probably the easiest fix, but would mean we're not testing against a default installation.
- Add a new testing installation mode that uses the default `DOCKER_NOT_AVAILABLE` and either skipping the tests that require e.g. Nextflow, or installing Nextflow on the GHA runner. This is probably the more complete testing fix, so we're testing both methods of install.

tools/install.sh
jaycarlton: yeah, I can either make IntelliJ happy or my co-workers happy...

tools/install.sh
jaycarlton: The IDE suggested it, so we don't delete `/lib` if somehow it becomes `null` under maintenance.

jaycarlton:
> * Regardless of whether the user sets the new env var or not, they can always toggle the app-launch config property post-install. Currently, when someone updates the docker image id config property, we check whether the image exists on the machine. Perhaps you can add this same check to the app-launch config property toggle. And make both output something to the user (e.g. "image not found. do docker pull ___ or use the local-process option"). I think now it just logs whether the image exists or not, there's no output to the user.

I'm not seeing a check for the image existence. In fact, I can set the image name to "foo" and it successfully changes it.
```
$ ./terra config set image --image=foo
Docker image: foo (CHANGED FROM gcr.io/terra-cli-dev/terra-cli/jaycarlton_test4:stable)
```

tools/local-dev.sh
```diff
@@ -6,27 +6,42 @@ set -e
...
-echo "Pulling the default Docker image"
-defaultDockerImage=$(terra config get image)
-docker pull $defaultDockerImage
+alias terra="$(pwd)"/build/install/terra-cli/bin/terra
+
+if [ "$terraCliInstallationMode" == "LOCAL_ONLY" ]; then
+  echo "Installing without docker support"
+else
+  echo "Setting the Docker image id to the default"
+  terra config set image --default
```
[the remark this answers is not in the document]
jaycarlton: There's a related issue of whether we should allow setting the image before the image has been pulled (or if it otherwise isn't spelled right, etc). I had put in code to bail if it wasn't there, but this broke `install-for-testing.sh` and likely `install.sh` too.

For now I'm going to remove the check & throw I had put in `bio.terra.cli.businessobject.Config#setDockerImageId` and revisit under [PF-1332](https://broadworkbench.atlassian.net/browse/PF-1332).

README.md
```diff
@@ -387,16 +387,10 @@ For example, a bucket within the workspace Google project. You can create these
...
-- `gcs-object`
-- `bq-dataset`
-- `bq-table`
-- `git-repo`
+`add-ref` command.
 
 The `check-access` command lets you see whether you have access to a particular resource. This is useful when a
-different user created or added the resource and subsequently shared the workspace with you. `check-access` currently always 
-returns true for `git-repo` reference type because workspace doesn't support authentication to external git services yet.
+different user created or added the resource and subsequently shared the workspace with you.
```
marikomedlock, line 393: Is this maybe a bad merge? It looks like you removed some README additions from another PR.

src/main/java/bio/terra/cli/command/config/set/AppLaunch.java
```diff
@@ -18,6 +20,15 @@
   protected void execute() {
     Config config = Context.getConfig();
     Config.CommandRunnerOption prevAppLaunchOption = config.getCommandRunnerOption();
+    if (CommandRunnerOption.DOCKER_CONTAINER == mode) {
+      String imageId = Context.getConfig().getDockerImageId();
+      if (!new DockerClientWrapper().checkImageExists(Config.getDefaultImageId())) {
+        OUT.printf(
+            "WARNING: Either the Docker daemon isn't running or image %s was not found on the local machine. "
+                + "Do `docker pull %s` to obtain it.%n",
```
marikomedlock, line 28: If Docker isn't running or installed, then I think `checkImageExists` will throw an exception, not return false. You can check that by turning off Docker on your machine and trying to change the image id.

src/main/java/bio/terra/cli/command/config/set/Image.java
```diff
@@ -27,6 +28,12 @@ protected void execute() {
     Config config = Context.getConfig();
     String prevImageId = config.getDockerImageId();
     String newImageId = argGroup.useDefault ? Config.getDefaultImageId() : argGroup.imageId;
+    if (!new DockerClientWrapper().checkImageExists(newImageId)) {
+      OUT.printf(
+          "WARNING: Either the Docker daemon isn't running or image %s was not found on the local machine. "
+              + "Do `docker pull %s` to obtain it.%n",
```
marikomedlock, line 34: same comment about exception as in `config set app-launch` class.

tools/install-for-testing.sh
```diff
@@ -15,6 +15,7 @@ if [ $(basename $PWD) != 'terra-cli' ]; then
   exit 1
 fi
 
+# Tests currently assume Docker is available and running.
```
marikomedlock, line 18: Please add a note about this in `CONTRIBUTING.md` in the tests section.

tools/install.sh
```diff
@@ -1,9 +1,23 @@
 #!/bin/bash
 set -e
 ## This script installs the Terra CLI from an unarchived release directory.
-## Dependencies: docker, gcloud
+## TERRA_CLI_DOCKER_MODE environment variable controls docker support. Set to
+##     DOCKER_NOT_AVAILABLE (default) to skip pulling the Docker image
+#      or DOCKER_AVAILABLE to pull the image (requires Docker to be installed and running).
```
marikomedlock, line 6: Please add notes about this env var and the default to the `README`, in the Install & run and Requirements sections.

tools/install.sh
```diff
@@ -13,37 +27,42 @@ fi
...
+  echo "--   Setting the Docker image id to the default"
+  ./terra config set image --default
+  ./terra config set app-launch DOCKER_CONTAINER
+  echo "--   Pulling the default Docker image"
+  defaultDockerImage=$(./terra config get image)
+  docker pull "$defaultDockerImage"
+fi
 
-echo "-- Setting the server to its current value, to pull any changes"
+echo "--   Setting the server to its current value, to pull any changes"
```
marikomedlock, line 62: nit: just curious, why the spacing changes here and below?

tools/local-dev.sh
```diff
@@ -2,31 +2,50 @@
 
 set -e
 ## This script sets up the environment for local development.
-## Dependencies: docker, chmod
+## TERRA_CLI_DOCKER_MODE environment variable controls docker support. Set to
+##     DOCKER_NOT_AVAILABLE (default) to skip pulling the Docker image
+#      or DOCKER_AVAILABLE to pull the image (requires Docker to be installed and running).
```
marikomedlock, line 7: ditto above: Please add notes about this to the `CONTRIBUTING.md` in the Setup development environment section.

tools/local-dev.sh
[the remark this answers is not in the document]
marikomedlock: Right now, we allow setting the image id regardless of whether the image is there or not. This is particularly helpful during install, because then we can set the image id to the default and then ask the cli what the default is. So the install script doesn't need to know what the default is explicitly.

I think it would be good to set the image id to the default regardless of what the app-launch config property is, because the user may toggle that property later, and if so, we want them to pull the latest cli image.

marikomedlock (commented):
Looks good, thanks. Mostly small comments & requests for documentation. One follow-up comment about setting the image id when docker isn't installed.

README.md
jaycarlton: yes, good catch. I don't know how that happened.

jaycarlton:
> I think it would be good to set the image id to the default regardless of what the app-launch config property is, because the user may toggle that property later, and if so, we want them to pull the latest cli image.
This takes me back to the chicken and egg problem I had before. Setting the `app-launch` mode calls `bio.terra.cli.app.utils.DockerClientWrapper#checkImageExists`, which will throw if Docker isn't running.

One approach is to handle the exception when we can't connect to docker differently, and make `checkImageExists()` into a boolean method. Is that reasonable/worth it?

CONTRIBUTING.md
```diff
@@ -38,10 +38,13 @@
...
 ### Setup development environment
-From the top-level directory:
+The TERRA_CLI_DOCKER_MODE environment variable controls Docker support. Set it to
+   DOCKER_NOT_AVAILABLE (default) to skip pulling the Docker image
+   or DOCKER_AVAILABLE to pull the image (requires Docker to be installed and running).
+Then, from the top-level directory, run:
 ```
 source tools/local-dev.sh
-terra
+build/install/terra-cli/bin/terra
```
marikomedlock, line 47: I don't follow why you need this change, because `local-dev.sh` aliases `terra` for you.

CONTRIBUTING.md
```diff
@@ -158,6 +161,9 @@ CLI installation on the same machine.
 - Run a single test by specifying the `--tests` option:
 `./gradlew runTestsWithTag -PtestTag=unit --tests "unit.Workspace.createFailsWithoutSpendAccess" --info`
 
+#### Docker and Tests
```
marikomedlock, line 164: nit: please add a link to this section in the table of contents

README.md
```diff
@@ -48,6 +48,11 @@ curl -L https://github.com/DataBiosphere/terra-cli/releases/latest/download/down
 ./terra
 ```
 
+By default, the CLI will be installed without support for Docker (i.e. it won't pull the Docker image).
```
marikomedlock, line 51: Also update the example usage section, since those commands will no longer work with a default installation unless the user has setup their environment exactly.

marikomedlock (commented):
> One approach is to handle the exception when we can't connect to docker differently, and make checkImageExists() into a boolean method. Is that reasonable/worth it?

That makes sense to me.

> However, I just realized that local-dev.sh calls render-config.sh, which requires Docker anyway. So developers will always need Docker running to install. 😿 Should we just ignore the docker mode environment variable for local-dev.sh?

I think this is more a question for the team/other CLI developers. The CLI repo uses docker to run tests and pull creds from vault. As-is, this seems untenable to me without docker. Going forward, if developers are not going to have docker available, then I think we need to rework that. Perhaps you can break this off into a separate PR (allowing development without docker, which includes modifying `local-dev`, `render-config`, and separating the tests that require docker into a separate tag/suite.

CONTRIBUTING.md
jaycarlton: Good point. I was working with different install modes on this PR and another one, and it's easy to get confused.

jmczerk (approved):

jmczerk:
Approved despite conflict (as it appears trivial), conditioned on passing checks once conflict is resolved.

PR #192, opened 2022-01-27
[PF-1309] Removed use of pet SA to call WSM.

jaycarlton:
As of https://github.com/DataBiosphere/terra-workspace-manager/pull/443, we no longer need to use the pet SA on the client side to interact with WSM. It's still useful when making calls directly to GCP, though.

### DIFF
--- a/src/main/java/bio/terra/cli/businessobject/Resource.java
+++ b/src/main/java/bio/terra/cli/businessobject/Resource.java
@@ -176,7 +176,7 @@ public abstract class Resource {
   public abstract String resolve();
 
   /**
-   * Check whether a user can access a resource.
+   * Check whether a user's pet SA can access a resource.
    *
    * @return true if the user can access the referenced resource with the given credentials
    * @throws UserActionableException if the resource is CONTROLLED
--- a/src/main/java/bio/terra/cli/businessobject/Resource.java
+++ b/src/main/java/bio/terra/cli/businessobject/Resource.java
@@ -176,9 +176,9 @@ public abstract class Resource {
   public abstract String resolve();
 
   /**
-   * Check whether a user's pet SA can access a resource.
+   * Check whether a user can access a resource.
    *
-   * @return true if the user's pet SA can access the referenced resource with the given credentials
+   * @return true if the user can access the referenced resource with the given credentials
    * @throws UserActionableException if the resource is CONTROLLED
    */
   public boolean checkAccess() {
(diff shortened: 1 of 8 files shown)


zloery (approved):
This PR LGTM, but I think you'll need to change the WSM side before the clone workspace test here will pass

src/main/java/bio/terra/cli/businessobject/Resource.java
```diff
@@ -176,9 +176,9 @@ public void delete() {
   public abstract String resolve();
 
   /**
-   * Check whether a user's pet SA can access a resource.
+   * Check whether a user can access a resource.
```
marikomedlock, line 179: Aren't we still checking that the pet SA can access a resource? I thought we just moved the pet SA credential fetch from the CLI to WSM.

marikomedlock (commented):
Thanks for this change. One question about what the check-access command is doing now

src/main/java/bio/terra/cli/businessobject/Resource.java
jaycarlton: That's true. I'm wondering if at this point pets are really an implementation detail in WSM and clients shouldn't know about them.

marikomedlock: Clients (or at least the CLI) still need to know about pet SAs because they may talk directly to the cloud in some cases (e.g. to get the number of tables in a BQ dataset). It can't use the user's credentials for that, because of insufficient OAuth scopes, so it has to call SAM to get the pet SA credentials.

Relatedly, users also have to know about proxy groups, because they need to share external resources with their proxy group if they want to add a referenced resource.

jaycarlton: That's good to know. I'll restore the previous value of the comment.

src/main/java/bio/terra/cli/businessobject/Resource.java
```diff
@@ -178,7 +178,7 @@ public void delete() {
   /**
    * Check whether a user's pet SA can access a resource.
    *
-   * @return true if the user's pet SA can access the referenced resource with the given credentials
+   * @return true if the user can access the referenced resource with the given credentials
```
marikomedlock, line 181: nit: pet SA to match above comment?

marikomedlock (approved):
LGTM, thanks for pushing this through! One nit about a comment inline.

PR #194, opened 2022-02-04
[PF-1289] Create server flag for using janitor on external resources

melissachang:
CLI tests use janitor to clean up external resources. Verily deployment doesn't have janitor, so I removed `rendered/janitor-client.json`. (This is equivalent to not having the `JANITOR_CLIENT_SA_KEY` secret for GHA.) `./gradlew runTestsWithTag -PtestTag=unit --scan -Pserver=verily-devel --tests=BqDatasetReferenced.listDescribeReflectAdd` fails because it expects `janitor-client.json`.

CLI tests create 2 kinds of resources:

1. [Tests create external resources directly](https://github.com/DataBiosphere/terra-cli/blob/1e438a30212385c874f3ec57061a7e42573a263c/src/test/java/unit/FineGrainedAccessGcsObjectReference.java#L48)
2. WSM resources, eg test runs `terra workspace create`

Janitor is only used to clean up 1. (See PF-886.) So it's ok to enable the janitor flag for `broad-dev`. (As opposed to say WSM tests. WSM tests run against `wsmtest` instead of dev WSM, because `wsmtest` talks to `buffertest` which has `cleanupAfterHandout=true`.)

After this PR, if I remove `rendered/janitor-client.json`, then test passes.


src/main/resources/servers/broad-dev.json
```diff
@@ -5,5 +5,6 @@
   "dataRepoUri": "https://jade.datarepo-dev.broadinstitute.org",
   "samUri": "https://sam.dsde-dev.broadinstitute.org",
   "workspaceManagerUri": "https://workspace.dsde-dev.broadinstitute.org",
-  "wsmDefaultSpendProfile": "wm-default-spend-profile"
+  "wsmDefaultSpendProfile": "wm-default-spend-profile",
+  "useJanitorForExternalResourcesCreatedByTests": true
```
zloery, line 9: iiuc the current behavior is that this is on for all environments. should it also be enabled for other Broad servers to match the current state?

zloery (approved):

src/main/resources/servers/broad-dev.json
melissachang: Done

marikomedlock (changes requested):
When we run CLI tests that create external resources, they **always** create those resources in the `terra-cli-test` [project](https://github.com/DataBiosphere/terra-cli/blob/2546f880b8599bab32e58ee4b4eab9cc411c898a/src/test/java/harness/TestExternalResources.java#L17), which is [Terraformed](https://github.com/broadinstitute/terraform-ap-deployments/tree/21b7a54191627e35df6613bc45554a9c2a771472/cli-test) under the `BROADINSTITUTE.ORG` organization.

This is true regardless of which server the tests run against. So when we run tests against `verily-devel`, we're hitting Verily WSM + SAM deloyments and still creating external resources in the `terra-cli-test` project.

A Janitor deployed at the Broad has permission to delete resources in the `terra-cli-test` project and should cleanup any leaked resources from the CLI tests. We should continue this. We don't want leaked resources to stick around in that project indefinitely just because the tests happened to be hitting a different WSM.

So if we want to turn off Janitor for the Verily deployments, then I think we also need to modify the tests to create external resources somewhere other than the `terra-cli-test` project. Somewhere where we don't care about things getting cleaned up, or plan to do so without a Janitor.

Not sure, but one approach may be to think of this as a test config property/flag rather than a server config property/flag. e.g.
- Default (Broad) test config specifies that we create external resources in `terra-cli-test` and tell `tools` Janitor about it.
- Verily test config specifies that we create external resources in `terra-cli-test-verily` (made up example project name) and don't tell Janitor about it (or in the future, tells a different Janitor about it).
- Default (Broad) test config is defined in this repo. Any tests run from this repo (against any server) will use this config.
- Verily test config is defined in the verily-src repo. Any tests run from that repo (against any server) will override the default and use that config.

melissachang:
Thanks Mariko. I created test config in `src/main/resouces/tests`. This repo has `broad.json`. `verily.json` gets populated when you run `render-verily-config.sh` from `verily-src/terra-tool-cli` (will open PR for `render-verily-config.sh` change).

Tested:
```
# Run against Broad dev
./tools/render-config.sh
./gradlew runTestsWithTag -PtestTag=unit --scan --tests=BqDatasetReferenced.listDescribeReflectAdd

# Run against Verily devel
[From terra-tool-cli repo] ./tools/render-verily-config.sh
In TestUsers.java, temporarily change emails to `cli_testuser_0X@test.verily-bvdp.com` (Will move to test config in next PR)
./gradlew runTestsWithTag -PtestTag=unit -Pserver=verily-devel --scan --tests=BqDatasetReferenced.listDescribeReflectAdd
```

FYI @cbookg

build.gradle
```diff
@@ -182,6 +182,10 @@ task runTestsWithTag(type: Test) {
         environment "TERRA_DOCKER_IMAGE", terraDockerImage
 
         if (testTag == "unit") {
+            // set which file in src/test/resources/testconfigs to use
+            String terraTestConfigFileName = terraServer.contains("broad") ? "broad.json" : "verily.json"
```
marikomedlock, line 186: Instead of keying off the server file name, I think it would be better to expose this as a separate test parameter (similar to `terraServer` above), with a default of `broad.json` (since `verily.json` doesn't exist in this repo). This would fully decouple server + test config. That way we preserve the ability run tests against any deployment from any repo. In the short term, I think this would break the nightly tests run from this repo against `verily-devel.json`.

build.gradle
```diff
@@ -182,6 +182,10 @@ task runTestsWithTag(type: Test) {
...
+            environment "TERRA_TEST_CONFIG_FILE_NAME", terraTestConfigFileName
```
marikomedlock, line 187: nit: Can you use a `systemProperty` for this instead of an env var? I think you'll only need to access this inside the test Java code.

src/main/java/bio/terra/cli/utils/JacksonMapper.java
```diff
@@ -25,7 +26,7 @@
 
   /** Getter for the singleton instance of the default Jackson {@link ObjectMapper} instance. */
   public static ObjectMapper getMapper() {
-    return objectMapper;
+    return objectMapper.configure(JsonParser.Feature.ALLOW_COMMENTS, true);
```
marikomedlock, line 29: I didn't know about this config, would have been really helpful. Thanks for the pointer!

src/test/java/harness/CRLJanitor.java
```diff
@@ -25,21 +25,20 @@
...
+  public static final ClientConfig getClientConfig() {
+    ClientConfig.Builder builder = ClientConfig.Builder.newBuilder().setClient(DEFAULT_CLIENT_NAME);
+    if (TestConfig.get().getUseJanitorForExternalResourcesCreatedByTests()) {
+      builder.setCleanupConfig(
+          CleanupConfig.builder()
+              .setTimeToLive(Duration.ofHours(2))
+              .setCleanupId("cli-test-" + System.getProperty("TEST_RUN_ID"))
+              .setCredentials(CRLJanitor.getSACredentials())
+              .setJanitorTopicName("crljanitor-tools-pubsub-topic")
+              .setJanitorProjectId("terra-kernel-k8s")
```
marikomedlock, line 37: Doesn't need to be in this PR (maybe just a todo?), but I think the Janitor pubsub topic name and project id should probably also live in the test config. e.g. if we have a Janitor in Verily side at some point, these will be different.

tools/render-config.sh
```diff
@@ -64,3 +64,8 @@ readFromVault "$JANITOR_CLIENT_SA_VAULT_PATH" "janitor-client.json" "base64"
 # used for granting break-glass access to a workspace in the verilycli deployment
 echo "Reading the WSM app service account key file for the verilycli deployment from Vault"
 readFromVault "$VERILYCLI_WSM_SA_VAULT_PATH" "verilycli-wsm-sa.json" "base64"
+
+# If it exists, remove verily test config. If someone tries to run tests with -Pserver=verily-devel and
+# they haven't run render-verily-config.sh, the test will fail because verily.json doesn't exist. This forces them to
+# run render-verily-config.sh.
+rm -f terra-cli/src/main/resources/tests/verily.json
```
marikomedlock, line 71: Should this live in the `verily-src` repo? I think it's confusing to reference these files from here when they only exist in another repo.

src/test/java/harness/TestConfig.java
```diff
@@ -0,0 +1,80 @@
...
+  private static final Logger logger = LoggerFactory.getLogger(TestConfig.class);
+
+  // Some CLI tests directly create external resources, eg FineGrainedAccessGcsObjectReference.java
+  private String projectForExternalResources;
+
+  // Whether to use janitor to clean up external resources. For more about terra-cli and Janitor,
+  // see PF-886.
+  private boolean useJanitorForExternalResourcesCreatedByTests;
+
+  public static TestConfig get() {
```
marikomedlock, line 24: Perhaps you can create a singleton object here, so we don't load from disk & validate every time a tests calls `get()`.

marikomedlock (commented):
This looks great, thanks. Some comments inline, and can you please also add a section to `CONTRIBUTING.md` about test configs.

src/test/java/harness/TestConfig.java
melissachang: Done, thanks.

src/test/java/harness/TestConfig.java
```diff
@@ -0,0 +1,80 @@
...
+    TestConfig testConfig;
+    try {
+      String filePath = TESTCONFIGS_RESOURCE_DIRECTORY + "/" + fileName;
+      logger.debug("Reading test config from {}", filePath);
+      InputStream inputStream = FileUtils.getResourceFileHandle(filePath);
+      testConfig = JacksonMapper.getMapper().readValue(inputStream, TestConfig.class);
+    } catch (FileNotFoundException fnfEx) {
+      throw new SystemException(
+          String.format(
+              "Test config file %s not found in src/main/resources/tests. You may need to run render-verily-config.sh.",
```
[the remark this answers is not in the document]
melissachang: Same - not ideal but most practical. Otherwise people will see "Test config file %s not found in src/main/resources/tests" and have to figure out what to do.

zloery (commented):
a couple questions about references to verily-src artifacts but otherwise LGTM, thanks for this!

marikomedlock (approved):
Thanks for all the work on this. Included a suggestion for the future work referenced in a TODO.

build.gradle
```diff
@@ -182,6 +182,12 @@ task runTestsWithTag(type: Test) {
         environment "TERRA_DOCKER_IMAGE", terraDockerImage
 
         if (testTag == "unit") {
+            // set which file in src/test/resources/testconfigs to use
+            // TODO(PF-1289): After tests running against Verily deployment are using Verily testconfig and not Broad,
+            // delete testConfig property and base on terraServer instead.
```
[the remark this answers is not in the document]
melissachang: Updated TODO to what Mariko and I discussed over chat.

melissachang (commented):
Fixed spotless

PR #197, opened 2022-02-10
[PF-1333] use ADC when available for log into terra CLI

yuhuyoyo:
terra auth login currently requires user authenticating through browser oauth flow. 

I added a flag --use-app-default-credentials that allows logging into terra with application default credentials when it is available. If the flag is on but the app default credential is not available, we will still open the browser as a fallback. 

This is to log into terra as the Pet service account when creating a notebook instance to simplify end user's setup.

### DIFF
--- a/src/main/java/bio/terra/cli/app/utils/AppDefaultCredentialUtils.java
+++ b/src/main/java/bio/terra/cli/app/utils/AppDefaultCredentialUtils.java
@@ -60,7 +60,7 @@ public class AppDefaultCredentialUtils {
    * current user or their pet SA
    */
   public static void throwIfADCDontMatchContext() {
-    if (!doADCMatchContext()) {
+    if (!doAdcMatchContext()) {
       throw new UserActionableException(
           "Application default credentials do not match the user or pet SA emails.");
     }
@@ -71,7 +71,7 @@ public class AppDefaultCredentialUtils {
    * user+workspace, or if they are end-user credentials, which we can't validate directly. Throw an
    * exception if they are not defined.
    */
-  private static boolean doADCMatchContext() {
+  private static boolean doAdcMatchContext() {
     GoogleCredentials appDefaultCreds = getApplicationDefaultCredentials();
 
     if (appDefaultCreds instanceof UserCredentials) {
--- a/src/main/java/bio/terra/cli/app/utils/AppDefaultCredentialUtils.java
+++ b/src/main/java/bio/terra/cli/app/utils/AppDefaultCredentialUtils.java
@@ -72,7 +72,7 @@ public class AppDefaultCredentialUtils {
    * exception if they are not defined.
    */
   private static boolean doADCMatchContext() {
-    GoogleCredentials appDefaultCreds = getADC();
+    GoogleCredentials appDefaultCreds = getApplicationDefaultCredentials();
 
     if (appDefaultCreds instanceof UserCredentials) {
       logger.info("ADC are end-user credentials. Skipping account/email validation.");
@@ -91,7 +91,7 @@ public class AppDefaultCredentialUtils {
   }
 
   /** Get the application default credentials. Throw an exception if they are not defined. */
-  public static GoogleCredentials getADC() {
+  public static GoogleCredentials getApplicationDefaultCredentials() {
     try {
       return GoogleCredentials.getApplicationDefault();
     } catch (IOException ioEx) {
--- a/src/main/java/bio/terra/cli/app/utils/AppDefaultCredentialUtils.java
+++ b/src/main/java/bio/terra/cli/app/utils/AppDefaultCredentialUtils.java
@@ -91,7 +91,7 @@ public class AppDefaultCredentialUtils {
   }
 
   /** Get the application default credentials. Throw an exception if they are not defined. */
-  private static GoogleCredentials getADC() {
+  public static GoogleCredentials getADC() {
     try {
       return GoogleCredentials.getApplicationDefault();
     } catch (IOException ioEx) {
--- a/src/main/java/bio/terra/cli/businessobject/User.java
+++ b/src/main/java/bio/terra/cli/businessobject/User.java
@@ -55,8 +55,11 @@ public class User {
   // pet SA email that Terra associates with this user. the CLI queries SAM to populate this field.
   private String petSAEmail;
 
-  // Google credentials for the user
+  // Google credentials for the user. This can be either the end-user google credentials or pet
+  // service account google credentials pulled from app-default-credentials.
   private GoogleCredentials googleCredentials;
+
+  private boolean useAppDefaultCredentials;
 
   // these are the same scopes requested by Terra service swagger pages
   @VisibleForTesting
@@ -121,21 +124,37 @@ public class User {
   }
 
   /**
-   * Load any existing credentials for this user. Prompt for login if they are expired or do not
-   * exist.
+   * Load any existing credneitals for the user. Prompt for login if they are expired or do not
+   * exist. Do not use application default credentials for logging in.
    */
   public static void login() {
+    login(false);
+  }
+
+  /**
+   * Load any existing credentials for this user. Prompt for login if they are expired or do not
+   * exist.
+   *
+   * @param useAdc whether to use application default credentials to log in.
+   */
+  public static void login(boolean useAdc) {
     Optional<User> currentUser = Context.getUser();
     currentUser.ifPresent(User::loadExistingCredentials);
 
     // populate the current user object or build a new one
     User user = currentUser.orElseGet(() -> new User());
 
-    try {
-      user.checkForAppDefaultCredentials();
-    } catch (UserActionableException e) {
-      logger.debug("Failed to get app default credentials, do oauth login flow instead");
-      // do the login flow if the current user is undefined or has expired credentials
+    if (useAdc) {
+      try {
+        user.checkForAppDefaultCredentials();
+      } catch (UserActionableException e) {
+        logger.debug("Failed to get app default credentials, do oauth login flow instead");
+        // do the login flow if the current user is undefined or has expired credentials
+        if (currentUser.isEmpty() || currentUser.get().requiresReauthentication()) {
+          user.doOauthLoginFlow();
+        }
+      }
+    } else {
       if (currentUser.isEmpty() || currentUser.get().requiresReauthentication()) {
         user.doOauthLoginFlow();
       }
@@ -162,6 +181,7 @@ public class User {
 
   private void checkForAppDefaultCredentials() {
     if (googleCredentials == null) {
+      useAppDefaultCredentials = true;
       googleCredentials = AppDefaultCredentialUtils.getADC().createScoped(PET_SA_SCOPES);
     }
   }
--- a/src/main/java/bio/terra/cli/businessobject/User.java
+++ b/src/main/java/bio/terra/cli/businessobject/User.java
@@ -1,6 +1,7 @@
 package bio.terra.cli.businessobject;
 
 import bio.terra.cli.app.utils.AppDefaultCredentialUtils;
+import bio.terra.cli.command.auth.Login.LogInMode;
 import bio.terra.cli.exception.SystemException;
 import bio.terra.cli.serialization.persisted.PDUser;
 import bio.terra.cli.service.GoogleOauth;
@@ -60,11 +61,11 @@ public class User {
   private GoogleCredentials googleCredentials;
 
   /**
-   * User specified to use app-default-credentials when logging in. In the follow-on command, we
-   * check for ADC instead of user credentials that are stored on disk in the Terra CLI credential
-   * store ($home/.terra).
+   * User specified what mode to log-in. When log-in mode is {@code APP_DEFAULT_CREDENTIALS}, check
+   * for ADC instead of user credentials that are stored on disk in the Terra CLI credential store
+   * ($home/.terra).
    */
-  private boolean useApplicationDefaultCredentials;
+  private LogInMode logInMode;
 
   // these are the same scopes requested by Terra service swagger pages
   @VisibleForTesting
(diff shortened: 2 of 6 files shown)


src/main/java/bio/terra/cli/businessobject/User.java
```diff
@@ -130,9 +131,14 @@ public static void login() {
     // populate the current user object or build a new one
     User user = currentUser.orElseGet(() -> new User());
 
-    // do the login flow if the current user is undefined or has expired credentials
-    if (currentUser.isEmpty() || currentUser.get().requiresReauthentication()) {
-      user.doOauthLoginFlow();
+    try {
```
marikomedlock, line 134: I think we should make this an explicit option on the login command instead of doing this by default (e.g. `terra login --use-application-default`). I set app-default creds on my machine for other things, and it would be unexpected if the CLI pulled those. Also it would often not work, since they're usually not my pet SA, but an unrelated SA.

Then you can call login with this option in the notebook `post-startup.sh` script.

src/main/java/bio/terra/cli/businessobject/User.java
```diff
@@ -55,7 +56,7 @@
   private String petSAEmail;
 
   // Google credentials for the user
-  private UserCredentials userCredentials;
+  private GoogleCredentials googleCredentials;
```
marikomedlock, line 59: nit: add a comment here explaining that this can be either the end-user google creds or pet SA google creds pulled from app-default

src/main/java/bio/terra/cli/businessobject/User.java
```diff
@@ -362,7 +376,8 @@ public GoogleCredentials getPetSACredentials() {
 
   /** Return true if the user credentials are expired or do not exist on disk. */
   public boolean requiresReauthentication() {
-    if (userCredentials == null) {
+    checkForAppDefaultCredentials();
```
marikomedlock, line 379: nit: I think this method should just check if existing credentials are expired, and not pull them, and we can keep the logic about which creds to use in the login method.

src/main/java/bio/terra/cli/businessobject/User.java
```diff
@@ -376,7 +391,7 @@ public boolean requiresReauthentication() {
 
   /** Get the access token for the user credentials. */
   public AccessToken getUserAccessToken() {
-    return GoogleOauth.getAccessToken(userCredentials);
+    return GoogleOauth.getAccessToken(googleCredentials);
```
marikomedlock, line 394: I think this method will [try to refresh](https://github.com/DataBiosphere/terra-cli/blob/f27ee455a91fa2f6b1eb60cfc4682d6447af3729/src/main/java/bio/terra/cli/service/GoogleOauth.java#L257) the credentials, but only log (not throw) an error if the refresh fails. Does this refresh work for app-default creds? I'm not sure if you can just call `googleCredentials.refreshIfExpired()` on the credential like for end-user credentials, or whether you need to call `GoogleCredentials.getApplicationDefault()` again.

marikomedlock (commented):
It's really great to see this working! Thanks for all the work here.

src/main/java/bio/terra/cli/businessobject/User.java
yuhuyoyo: loadExistingCredentials is called by `BaseCommand` when every command is run. But `loadExistingCredentials` will set `googleCredentials` to null when there is no user credentials. I was not able to get ADC in that method for some reason (not sure if it has something to do with it is called by the `BaseCommand`.

src/main/java/bio/terra/cli/command/auth/Login.java
```diff
@@ -11,6 +12,11 @@
     description = "Authorize the CLI to access Terra APIs and data with user credentials.")
 public class Login extends BaseCommand {
 
+  @CommandLine.Option(
+      names = "--use-app-default-credentials",
+      description = "Use application default credentials to log into terra CLI instead of browser.")
+  private boolean useAppDefaultCreds;
```
zloery, line 18: If this flag isn't specified, does the value default to false?

src/main/java/bio/terra/cli/businessobject/User.java
```diff
@@ -119,20 +121,41 @@ public void loadExistingCredentials() {
     }
   }
 
+  /**
+   * Load any existing credneitals for the user. Prompt for login if they are expired or do not
```
zloery, line 125: nit: credentials

src/main/java/bio/terra/cli/businessobject/User.java
```diff
@@ -119,20 +121,41 @@ public void loadExistingCredentials() {
...
-    if (currentUser.isEmpty() || currentUser.get().requiresReauthentication()) {
-      user.doOauthLoginFlow();
+    if (useAdc) {
+      try {
+        user.checkForAppDefaultCredentials();
+      } catch (UserActionableException e) {
+        logger.debug("Failed to get app default credentials, do oauth login flow instead");
+        // do the login flow if the current user is undefined or has expired credentials
+        if (currentUser.isEmpty() || currentUser.get().requiresReauthentication()) {
+          user.doOauthLoginFlow();
```
zloery, line 152: imo I think this should throw the UserActionableException instead of falling back to the login flow. 
In this case the user has explicitly specified the `--use-app-default-credentials` flag, so I think they would expect the CLI to either use ADC or throw, rather than falling back to ignoring the flag.

src/main/java/bio/terra/cli/businessobject/User.java
```diff
@@ -154,11 +177,20 @@ public static void login() {
     }
   }
 
+  private void checkForAppDefaultCredentials() {
```
zloery, line 180: nit: this is loading the ADC into the user context, not just checking for them. Maybe something like `readAppDefaultCredentials()` or `loadAppDefaultCredentials()`?

zloery (approved):

src/main/java/bio/terra/cli/command/auth/Login.java
yuhuyoyo: yes

yuhuyoyo:
@marikomedlock what do you mean by setting a flag to the class. i tried setting a boolean flag but it get reset eveytime i run a new command

src/main/java/bio/terra/cli/businessobject/User.java
yuhuyoyo: done

src/main/java/bio/terra/cli/businessobject/User.java
```diff
@@ -54,8 +54,16 @@
...
-  // Google credentials for the user
-  private UserCredentials userCredentials;
+  // Google credentials for the user. This can be either 1) end-user google credentials from an
+  // oauth browser flow, or 2) end-user or pet sa google credentials pulled from the application
+  // default credentials.
+  private GoogleCredentials googleCredentials;
+
+  /**
+   * User specified to use app-default-credentials when logging in. In the follow-on command, we
+   * check for ADC instead of user credentials that are stored on disk.
```
marikomedlock, line 64: nit: on disk in the Terra CLI credential store (`$home/.terra`).

src/main/java/bio/terra/cli/businessobject/User.java
```diff
@@ -154,11 +180,25 @@ public static void login() {
     }
   }
 
+  /**
+   * Fetches a Google Application Credentials with {@code PET_SA_SCOPES} and set to {@code
+   * googleCredentials}.
+   */
+  private void loadAppDefaultCredentials() {
+    if (googleCredentials == null) {
+      useApplicationDefaultCredentials = true;
```
marikomedlock, line 189: Can you login with ADC, and then login with the regular OAuth flow? I'm not sure, but it looks like you set this to true and there's no way to get it set back to false.

If that's the case, perhaps you can set this property in the `login` command class. e.g. If they specify the use ADC flag, then call `User.setUseADC(true)` before calling `login`. And if they don't specify that flag, then call `User.setUseADC(false)` before calling `login`.

yuhuyoyo: thanks. that is a very good point. setUseAdc needs to be a static method to be called in `login` command class. So instead I reuse my current structure and set `useAdc` field in `login`. When `login(false)`, useAdc field is set to fales and vice versa. 

i tried terra auth login, and then terra auth login --use-app-default-credential and then terra auth login, I'm able to switch between the two

yuhuyoyo:
Hey Jay could you review too because Mariko is OOO now?

src/main/java/bio/terra/cli/app/utils/AppDefaultCredentialUtils.java
```diff
@@ -72,7 +72,7 @@ public static void throwIfADCDontMatchContext() {
    * exception if they are not defined.
    */
   private static boolean doADCMatchContext() {
```
jaycarlton, line 74: nit: should probably be `doAdcMatchContext()` per our conventions.

src/main/java/bio/terra/cli/businessobject/User.java
```diff
@@ -98,9 +108,13 @@ public void loadExistingCredentials() {
     // load existing user credentials from disk
     try (InputStream inputStream =
         User.class.getClassLoader().getResourceAsStream(CLIENT_SECRET_FILENAME)) {
-      userCredentials =
-          GoogleOauth.getExistingUserCredential(
-              USER_SCOPES, inputStream, Context.getContextDir().toFile());
+      if (useApplicationDefaultCredentials) {
```
jaycarlton, line 111: I think you should move this check outside of the try-with-resources, since we don't need to find & open the file for this option.

src/main/java/bio/terra/cli/businessobject/User.java
```diff
@@ -154,11 +183,24 @@ public static void login() {
     }
   }
 
+  /**
+   * Fetches a Google Application Credentials with {@code PET_SA_SCOPES} and set to {@code
+   * googleCredentials}.
```
jaycarlton, line 188: document that this only happens when googleCredentials is null. Maybe name it `loadAppDefaultCredentialsAsFallback()` or something.

src/main/java/bio/terra/cli/businessobject/User.java
```diff
@@ -352,17 +394,21 @@ public String getPetSaEmail() {
     return petSAEmail;
   }
 
-  public UserCredentials getUserCredentials() {
-    return userCredentials;
+  public GoogleCredentials getGoogleCredentials() {
```
jaycarlton, line 397: Maybe make this `Optional<GoogleCredentials>` since the field is nullable? Or is it only null during initialization?

src/main/java/bio/terra/cli/service/GoogleOauth.java
```diff
@@ -275,16 +275,18 @@ public static AccessToken getAccessToken(GoogleCredentials credential) {
    * @param credential credentials object
    */
   public static void revokeToken(GoogleCredentials credential) {
-    String endpoint = "https://oauth2.googleapis.com/revoke";
-    Map<String, String> headers =
-        ImmutableMap.of("Content-type", "application/x-www-form-urlencoded");
-    Map<String, String> params =
-        ImmutableMap.of("token", credential.getAccessToken().getTokenValue());
+    if (credential.getAccessToken() != null) {
```
jaycarlton, line 278: this method would also be nice as an `Optional`

src/main/java/bio/terra/cli/businessobject/User.java
```diff
@@ -119,20 +133,35 @@ public void loadExistingCredentials() {
...
+
   /**
    * Load any existing credentials for this user. Prompt for login if they are expired or do not
    * exist.
+   *
+   * @param useApplicationDefaultCredentials when true, use the application default credentials to
+   *     log-in instead of oauth flow in the browser.
    */
-  public static void login() {
+  public static void login(boolean useApplicationDefaultCredentials) {
```
jaycarlton, line 151: Maybe make this parameter an enum like `LoginMode` so it's more obvious at the call site and we can support more modes later.

src/main/java/bio/terra/cli/businessobject/User.java
```diff
@@ -54,8 +54,17 @@
   // pet SA email that Terra associates with this user. the CLI queries SAM to populate this field.
   private String petSAEmail;
 
-  // Google credentials for the user
-  private UserCredentials userCredentials;
+  // Google credentials for the user. This can be either 1) end-user google credentials from an
+  // oauth browser flow, or 2) end-user or pet sa google credentials pulled from the application
+  // default credentials.
```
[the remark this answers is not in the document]
yuhuyoyo: no. it's not only available in VM. for example, you can set GOOGLE_APPLICATION_CREDENTIALS to the secrive account file on your disk.
Google documentation on how ADC works:
```
If the environment variable GOOGLE_APPLICATION_CREDENTIALS is set, ADC uses the service account key or configuration file that the variable points to.

If the environment variable GOOGLE_APPLICATION_CREDENTIALS isn't set, ADC uses the service account that is attached to the resource that is running your code.

This service account might be a [default service account](https://cloud.google.com/iam/docs/service-accounts#default) provided by Compute Engine, Google Kubernetes Engine, App Engine, Cloud Run, or Cloud Functions. It might also be a [user-managed service account](https://cloud.google.com/iam/docs/service-accounts#user-managed) that you created.

If ADC can't use any of the above credentials, an error occurs.
```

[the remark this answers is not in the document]
jaycarlton: Thanks

src/main/java/bio/terra/cli/businessobject/User.java
jaycarlton: Right, but you can wrap it with `Optional.ofNullable()` in the getter for safety.

PR #198, opened 2022-02-10
[PF-1267] Move test users to test config

melissachang:
For diff, recommend viewing just `c8be9d6` commit.

Tested: Ran a bunch of tests using both verily and broad test config.

Will merge after #194.

@cbookg


gradle/config/spotbugs/exclude.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- http://spotbugs.readthedocs.io/en/latest/filter.html -->
+<FindBugsFilter>
+
+    <Match>
+        <!-- spotbugs thinks email/spendEnabled are "Unwritten public or
+         protected field". They are written when JacksonMapper reads from JSON
+         file. -->
```
marikomedlock, line 8: We've typically used annotations directly in the class/method to suppress false positives for static analysis ([example use](https://github.com/DataBiosphere/terra-cli/blob/6fc7eb78408fbabedd8eac4d8e63c13cdd3c2765/src/main/java/bio/terra/cli/serialization/userfacing/input/GcsBucketLifecycle.java#L20)). Perhaps something to check with the team about whether there's a norm/convention to follow here.

src/test/java/unit/SpendProfileUser.java
```diff
@@ -119,26 +119,26 @@ void onlyAdminCanEnableDisable() throws IOException {
 
   @Test
   @DisplayName("spend profile includes cli-testers group and the appropriate test users")
-  void testUsersSpendProfileMembership() throws IOException {
+  void TestUserSpendProfileMembership() throws IOException {
```
marikomedlock, line 122: nit: unintended rename?

tools/setup-test-users.sh
```diff
@@ -2,7 +2,7 @@
 
 set -e
 ## This script sets up users for running CLI tests. It only needs to be run once per SAM instance.
-## Keep this script in sync with the harness.TestUsers class in the src/test/java directory.
+## Keep this script in sync with test users in your testconfig file (eg `testconfig/broad.json`).
```
marikomedlock, line 5: Could this script take in the test config as input? If so, it's fine to do in a follow-on PR, but here can you please: add a todo to point to a ticket and rename this script e.g. `setup-broad-test-uesrs.sh`?

build.gradle
```diff
@@ -177,17 +177,18 @@ task runTestsWithTag(type: Test) {
         String terraServer = project.hasProperty("server") ? project.findProperty("server") : "broad-dev"
         environment "TERRA_SERVER", terraServer
 
+        // set which file in src/test/resources/testconfigs to use
+        // TODO(PR-1289): After tests running against Verily deployment are using Verily testconfig and not Broad,
```
zloery, line 181: nit: PF-1289, I think you fixed this in the base branch so this change might be out of date

zloery (approved):

tools/setup-test-users.sh
melissachang: Added TODO, thanks.

build.gradle
melissachang: Done

marikomedlock (approved):

PR #199, opened 2022-02-10
[PF-1289] Make deployment sub-dirs under `rendered`

melissachang:
Moves files like `test-user-account.json` from `rendered` to `rendered/broad`. This makes it easier to run tests against different deployments. Before this PR:

```
render-config.sh
run test against Broad (`-Pserver=broad-dev -Ptestconfig=broad`)
render-verily-config.sh   (from a different repo)
run test against Verily (`-Pserver=verily-devel -Ptestconfig=verily`)
```

I had to run render-config.sh to switch files like `test-user-account.json` from Broad to Verily.

After this PR:

```
run test against Broad (`-Pserver=broad-dev -Ptestconfig=broad`)
run test against Verily (`-Pserver=verily-devel -Ptestconfig=verily`)
```

Tested: Ran a test using Broad and Verily testconfigs, without running `render-config.sh`.


src/test/java/harness/TestUser.java
```diff
@@ -108,7 +99,7 @@ public GoogleCredentials getCredentialsWithCloudPlatformScope() throws IOExcepti
   /** Get domain-wide delegated Google credentials for this user. */
   private GoogleCredentials getCredentials(List<String> scopes) throws IOException {
     // get a credential for the test-user SA
-    Path jsonKey = Path.of("rendered", "test-user-account.json");
+    Path jsonKey = Path.of("rendered", TestConfig.getTestConfigName(), "test-user-account.json");
```
marikomedlock, line 102: Instead of hard-coding the SA key file names, I think it would be cleaner to make these properties of the test config. The filenames we currently expect `render-config.sh` to generate are not really documented anywhere, and this would obviate the need for that.
e.g. in json `"domainWideDelegationSA": "rendered/broad/test-user-account.json")`
       in java `TestConfig.getDomainWideDelegationSAFile()`

marikomedlock (commented):
If I'm reading this right, I think you've got 3 chained PRs. I think you can also change the base branch of a PR to show just the unique changes in each one.

src/test/java/harness/TestUser.java
melissachang: Discussed offline. Personally I prefer having fewer knobs.

This discussion brings up a good point: I need to document "Here's what to do for a new deployment". I will do that as part of PF-1289. The documentation will already say "Set up a `render-config.sh` for your deployment`". (And it's fairly certain someone will be reading this documentation as part of setting up a new deployment :)) So the documentation can just say "Create files name such-and-such".

melissachang:
Friendly ping for Zach

zloery (approved):
LGTM

PR #200, opened 2022-02-11
[PF-1340] clarify how to specify multiple metadata when creating notebook

yuhuyoyo:
it is unclear if user is not familiar with picoli command options what the input format is for map Command.Options.

### DIFF
--- a/src/main/java/bio/terra/cli/command/resource/create/GcpNotebook.java
+++ b/src/main/java/bio/terra/cli/command/resource/create/GcpNotebook.java
@@ -65,9 +65,11 @@ public class GcpNotebook extends BaseCommand {
   private String postStartupScript;
 
   @CommandLine.Option(
-      names = "--metadata",
+      names = "-M, --metadata",
       description =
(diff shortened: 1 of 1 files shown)


src/main/java/bio/terra/cli/command/resource/create/GcpNotebook.java
```diff
@@ -65,9 +65,11 @@
   private String postStartupScript;
 
   @CommandLine.Option(
-      names = "--metadata",
+      names = "-M, --metadata",
       description =
-          "Custom metadata to apply to this instance.\nBy default sets some jupyterlab extensions "
+          "Custom metadata to apply to this instance.\n"
+              + "specify multiple metadata in the format of --metadata=key1=value1 --Mkey2=value2.\n"
```
jaycarlton, line 71: should that be `-M` instead of `--M`?

yuhuyoyo: i'm following this example https://picocli.info/#_key_only_map_parameters and they use one dash for the short version and two dash for the full name.

zloery (approved):

PR #201, opened 2022-02-11
Fix GHA artifact upload

melissachang:
Now see:

image


zloery (approved):
Good catch! Related, could you also update the comments referring to `global-context.json` in `README.md` and `Context.java`?

melissachang:
> Good catch! Related, could you also update the comments referring to `global-context.json` in `README.md` and `Context.java`?

Done

marikomedlock (approved):
Thanks for finding this!

PR #202, opened 2022-02-11
[PF-1341] When setting metadata to the notebook, still append the default metadata unless specified to be override

yuhuyoyo:
the default metadata are workspace id and server id. Unless specifically override by user input, we should still append those when some arbitray metadata is appended.

### DIFF
--- a/src/main/java/bio/terra/cli/command/resource/create/GcpNotebook.java
+++ b/src/main/java/bio/terra/cli/command/resource/create/GcpNotebook.java
@@ -228,9 +228,9 @@ public class GcpNotebook extends BaseCommand {
             .populateMetadataFields()
             .stewardshipType(StewardshipType.CONTROLLED)
             .accessScope(AccessScope.PRIVATE_ACCESS);
-    Map<String, String> metadatas = defaultMetadata(Context.requireWorkspace().getId());
+    Map<String, String> allMetadata = defaultMetadata(Context.requireWorkspace().getId());
     if (metadata != null) {
-      metadatas.putAll(metadata);
+      allMetadata.putAll(metadata);
     }
     CreateGcpNotebookParams.Builder createParams =
         new CreateGcpNotebookParams.Builder()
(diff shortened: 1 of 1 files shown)


yuhuyoyo:
@deflaux FYI

Currently when metadata field is specified during notebook creation, we override the default metadata fields that include workspace id and terra server id. I changed it to append to the default metadata field and override when the user specified.

src/main/java/bio/terra/cli/command/resource/create/GcpNotebook.java
```diff
@@ -227,6 +227,9 @@ protected void execute() {
             .populateMetadataFields()
             .stewardshipType(StewardshipType.CONTROLLED)
             .accessScope(AccessScope.PRIVATE_ACCESS);
+    if (metadata != null) {
+      metadata.putAll(defaultMetadata(Context.requireWorkspace().getId()));
```
zloery, line 231: Checking my understanding: this will always set the metadata keys `terra-workspace-id` and `terra-cli-server` to appropriate values, including overwriting user-provided values for those keys if they're provided?

I think that's reasonable behavior, but it might be worth documenting or warning users that these are reserved metadata keys (if we don't do that already)

zloery (approved):

src/main/java/bio/terra/cli/command/resource/create/GcpNotebook.java
```diff
@@ -229,15 +228,18 @@ protected void execute() {
             .populateMetadataFields()
             .stewardshipType(StewardshipType.CONTROLLED)
             .accessScope(AccessScope.PRIVATE_ACCESS);
+    Map<String, String> metadatas = defaultMetadata(Context.requireWorkspace().getId());
```
jaycarlton, line 231: maybe name this `allMetadata` since "metadatas" isn't a word.

jaycarlton (approved):

PR #204, opened 2022-02-14
[PF-1310] Build with GraalVM Native Image

jaycarlton:
This PR is a proof of concept for building the CLI application as a native executable image on MacOS.

## The Good
```
file build/native-image/terra
build/native-image/terra: Mach-O 64-bit executable x86_64
```
Currently, the image is 80 MB, though it would likely need to grow in order to include some more classes.

The Gradle plugin I settled on did a good job of wrapping the build process, once I figured out the right command line options. The native executable is generated in about two minutes, and runs quickly (minimal startup time). The memory footprint should be better as well, though the runtime is so quick it's hard to get a feel for it.

The compiler output is informative, with reasonable error messages most of the time. 

There are some excellent videos on youtube and decent documentation for the toolchain.

The GraalVM JVM itself works well when running the CLI as a regular Java application. I didn't see any issues with the runtime.

## The Bad
Building the Native Image requires accounting for every class that is loaded dynamically during execution. There are thousands of such classes. This task is alleviated greatly by the native agent that ships with GraalVM, which generates config files by instrumenting the code at runtime and finding reflective calls. In the `resources/META-INF/native-image` directory, it's apparent that these configuration files can't be easily maintained by hand. Any missing class (i.e. any workflow or branch we didn't instrument) causes a runtime failure that our users won't be able to do anything about.

## The Ugly 
There were some bugs in Native Image that I never found a solution for. One of them involves an NPE in the reflection building code that seems to crop up once the reflect-config.json file gets to a certain size. It will take some experimentation and probably a bug with the maintainer to get progress there.

Another is that at  runtime, the Logback configuration code was crashing with an error about the integer component in the log file format, which makes no sense, as the same code works in the Graal JVM. I worked around this by removing the call to set the logging configuration dynamically and instead using the legacy `logback.xml` facility.

## Conclusions
1. It is possible to generate a working (or partially working) executable using this tool.
2. Maintenance and testing will both be expensive, as we need exhaustive coverage for the agent (ideally in a script) to activate all the classes we'll need in the image. Any changes to code, library versions, or native image itself could mean we're no longer including all the classes we need.
3. The NPE crash in the compiler is disconcerting, but possibly something we can get past. I've had to revert changes to the reflection config a couple of times when I've encountered it, but I could probably isolate the change(s) that are responsible.
4. I propose shelving this effort for a year or so and seeing how the tools improve in the meantime. If we continue the effort to build an executable that can accomplish all the CLI workflows and get past any remaining compiler bugs, we'd still be likely to have users run into a missing class at runtime, which is an error they can't fix without a new build of the CLI with additional reflection classes or resources listed.


src/main/java/bio/terra/cli/command/Main.java
```diff
@@ -187,12 +188,19 @@ public int handleExecutionException(Exception ex, CommandLine cmd, ParseResult p
...
-            "An unexpected error occurred in "
+            "An unexpected error occurred: "
                 + ex.getClass().getCanonicalName()
                 + ": "
-                + ex.getMessage();
+                + ex.getMessage()
+                + "\n";
         formattedErrorMessage =
-            systemAndUnexpectedErrorStyle.errorText("[ERROR] ").concat(errorMessage);
+            systemAndUnexpectedErrorStyle
```
jaycarlton, line 197: For a while, I was winding up with unexpected errors, but no stack trace printed, so I added this. I _think_ it's not strictly necessary, as I was getting duplicate stack traces after fixing the logging.

src/main/resources/META-INF/native-image/reflect-config.json
```diff
@@ -0,0 +1,2608 @@
+[
+{
```
jaycarlton, line 2: Believe it or not, I started out trying to build up this file by hand, adding a new class whenever I encountered a runtime error. That proved unworkable, and fortunately the agent was easier to use than I had anticipated.

src/main/resources/client_secret.json
```diff
@@ -1 +1,14 @@
-{"installed":{"client_id":"436986732241-nvjfkarirkhmcq7m3r1uemdau518uvdo.apps.googleusercontent.com","project_id":"terra-cli-dev","auth_uri":"https://accounts.google.com/o/oauth2/auth","token_uri":"https://oauth2.googleapis.com/token","auth_provider_x509_cert_url":"https://www.googleapis.com/oauth2/v1/certs","client_secret":"w7TLNFm2fog61d2cFNC_fO5L","redirect_uris":["urn:ietf:wg:oauth:2.0:oob","http://localhost"]}}
\ No newline at end of file
+{
+  "installed": {
```
jaycarlton, line 2: There was a problem initially with the JSON parsing inside the Google OAuth library, and I pretty-printed this file in an effort to sanity check it.

PR #205, opened 2022-02-15
[PF-1332] friendly error on missing executable

jaycarlton:
Factor out common logic to a `ToolCommand` intermediate base class. Insert a check for the `version` of the applicable tool. If that process fails to launch or returns a nonzero exit status, throw a `UserActionableException`.

### DIFF
--- a/src/main/java/bio/terra/cli/command/app/passthrough/ToolCommand.java
+++ b/src/main/java/bio/terra/cli/command/app/passthrough/ToolCommand.java
@@ -25,6 +25,8 @@ public abstract class ToolCommand extends BaseCommand {
     return "version";
   }
 
+  abstract public String getInstallationUrl();
+
   @CommandLine.Mixin protected WorkspaceOverride workspaceOption;
 
   @CommandLine.Unmatched protected final List<String> command = new ArrayList<>();
@@ -36,7 +38,11 @@ public abstract class ToolCommand extends BaseCommand {
         && !toolIsInstalled()) {
       throw new UserActionableException(
           String.format(
-              "Unable to launch %s. Please verify it is installed and included in the PATH.",
+              "Unable to launch %s. Please verify it is installed and included in the PATH. " +
+              "For installation instructions, see %s. Alternatively, run `terra config set " +
+                  "app-launch DOCKER_CONTAINER` to run %s via Docker.",
+              getExecutableName(),
+              getInstallationUrl(),
               getExecutableName()));
     }
     executeImpl();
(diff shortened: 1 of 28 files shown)


yuhuyoyo (approved):

jaycarlton:
> is this related to the executable error?
It's going in with https://github.com/DataBiosphere/terra-cli/pull/207, which is finishing up integration tests now. I'll merge it back out before submitting this one.

src/main/java/bio/terra/cli/command/app/passthrough/ToolCommand.java
```diff
@@ -0,0 +1,78 @@
...
+  @CommandLine.Unmatched protected final List<String> command = new ArrayList<>();
+
+  /** Pass the command through to the CLI Docker image. */
+  @Override
+  protected void execute() {
+    if (CommandRunnerOption.LOCAL_PROCESS == Context.getConfig().getCommandRunnerOption()
+        && !toolIsInstalled()) {
+      throw new UserActionableException(
+          String.format(
+              "Unable to launch %s. Please verify it is installed and included in the PATH.",
```
melissachang, line 39: I think it would be nice to print a url, so the message is more actionable. This could be helpful if user isn't technical.

At first I was thinking you could link to the README installation section. Then I was thinking it could be nice to link to installation instructions directly. Could `ToolCommand` have a method that each tools sets like:

```
String getInstallationUrl()
```

Also, I wonder if we should mention docker, in case the user might prefer docker over installing stuff manually.

```
Unable to launch %s. To install %s, see %s. Alternatively, run `terra config set app-launch DOCKER` to run %s inside a docker container.
```

src/main/java/bio/terra/cli/command/app/passthrough/ToolCommand.java
```diff
@@ -0,0 +1,78 @@
...
+    // container starts
+    command.add(0, getExecutableName());
+    Context.getConfig().getCommandRunnerOption().getRunner().runToolCommand(command);
+  }
+
+  /**
+   * Launch a no-op process to determine if the tool is installed. Don't use the full CommandRunner,
+   * because we don't want to set environment variables or clean up.
+   */
+  protected boolean toolIsInstalled() {
```
melissachang, line 61: Could there be integration test, to check that this works? Test would check error that is printed.

melissachang (approved):
Nice

src/main/java/bio/terra/cli/command/app/passthrough/ToolCommand.java
jaycarlton: It's hard to explain the docker option in an error message, but adding the URLs is easy enough.

PR #210, opened 2022-02-18
[PF-1321] increase limit when fetching workspace list

jaycarlton:
Stopgap to work around an issue caused by many leaked workspaces in `dev`. When fetching the workspace list, if there were too many workspaces, the one we care about might not be returned in the first page, as the test expects. Solution for now is to use a much larger `limit` parameter than the default.


jmczerk (approved):
I'm good with this as a workaround to limit exposure, but let's keep the bug open even when this goes in.  Thanks!

zloery (approved):

PR #211, opened 2022-02-23
PF-1352: Add git as a passthrough app

yuhuyoyo:
https://docs.google.com/document/d/14sn6ValS3Mtq9HDlYmUioJ3RzpDRLwsp5Yfby63JVKE/edit?usp=sharing

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -8,7 +8,7 @@ WORKDIR /usr
 RUN apt-get update \
     && apt-get install -y curl \
     && apt-get install -y python \
-    && apt-get install git
+    && apt-get install -y git
 
 # install Google Cloud SDK (gcloud, gsutil, bq)
 # these commands come from https://cloud.google.com/sdk/docs/downloads-versioned-archives
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -6,9 +6,10 @@ WORKDIR /usr
 
 # install curl (to download install packages), python (for gcloud sdk) and git
 RUN apt-get update \
-    && apt-get install -y curl \
-    && apt-get install -y python \
-    && apt-get install -y git
+    && apt-get install -y \
+    curl \
+    python \
+    git
 
 # install Google Cloud SDK (gcloud, gsutil, bq)
 # these commands come from https://cloud.google.com/sdk/docs/downloads-versioned-archives
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -4,10 +4,11 @@ FROM openjdk:11.0-jre-slim
 # working directory when the docker is run
 WORKDIR /usr
 
-# install curl (to download install packages) and python (for gcloud sdk)
+# install curl (to download install packages), python (for gcloud sdk) and git
 RUN apt-get update \
     && apt-get install -y curl \
-    && apt-get install -y python
+    && apt-get install -y python \
+    && apt-get install git
 
 # install Google Cloud SDK (gcloud, gsutil, bq)
 # these commands come from https://cloud.google.com/sdk/docs/downloads-versioned-archives
--- a/src/main/java/bio/terra/cli/command/app/passthrough/Git.java
+++ b/src/main/java/bio/terra/cli/command/app/passthrough/Git.java
@@ -21,7 +21,7 @@ public class Git extends ToolCommand {
   public String[] names;
 
   @CommandLine.Option(
-      names = "-a",
+      names = "--all",
       description = "clone all the git repo resources in the current workspace")
   public boolean cloneAll;
 
@@ -62,7 +62,7 @@ public class Git extends ToolCommand {
   }
 
   private void validateCloneCommand() {
-    if (command.size() != 2 || !command.get(0).equals("git") || !command.get(1).equals("clone")) {
+    if (command.size() != 2 || !command.get(1).equals("clone")) {
       throw new UserActionableException(
           "Did you mean to clone git repo resources in the workspace? If so, please use terra git clone");
     }
@@ -70,7 +70,6 @@ public class Git extends ToolCommand {
 
   private void cloneGitRepoResource(Resource resource) {
     List<String> cloneCommands = Stream.of("git", "clone").collect(Collectors.toList());
-    ;
     String gitRepoUrl = resource.resolve();
     cloneCommands.add(gitRepoUrl);
     try {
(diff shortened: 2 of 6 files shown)


docker/Dockerfile
```diff
@@ -4,10 +4,11 @@ FROM openjdk:11.0-jre-slim
 # working directory when the docker is run
 WORKDIR /usr
 
-# install curl (to download install packages) and python (for gcloud sdk)
+# install curl (to download install packages), python (for gcloud sdk) and git
 RUN apt-get update \
     && apt-get install -y curl \
```
jaycarlton, line 9: can these all be on one line like `apt-get install curl python git`?

src/main/java/bio/terra/cli/command/app/passthrough/Git.java
```diff
@@ -0,0 +1,87 @@
...
+
+  @CommandLine.Option(
+      names = "--resource",
+      description =
+          "name of the git-repo resources in the current workspace to be cloned. \n"
+              + "Example usage: git clone --resource=repo1 --resource=repo2")
+  public String[] names;
+
+  @CommandLine.Option(
+      names = "-a",
```
jaycarlton, line 24: I'd give this a longer name like `--clone-all`.

src/main/java/bio/terra/cli/command/app/passthrough/Git.java
```diff
@@ -0,0 +1,87 @@
...
+      for (String name : names) {
+        cloneGitRepoResource(Context.requireWorkspace().getResource(name));
+      }
+      return;
+    }
+    Context.getConfig().getCommandRunnerOption().getRunner().runToolCommand(command);
+  }
+
+  private void validateCloneCommand() {
+    if (command.size() != 2 || !command.get(0).equals("git") || !command.get(1).equals("clone")) {
```
jaycarlton, line 65: I'm not sure you can get here at all if the first token isn't `git`.

src/main/java/bio/terra/cli/command/app/passthrough/Git.java
```diff
@@ -0,0 +1,87 @@
...
+      throw new UserActionableException(
+          "Did you mean to clone git repo resources in the workspace? If so, please use terra git clone");
+    }
+  }
+
+  private void cloneGitRepoResource(Resource resource) {
+    List<String> cloneCommands = Stream.of("git", "clone").collect(Collectors.toList());
+    ;
```
jaycarlton, line 73: remove this line

src/main/java/bio/terra/cli/command/app/passthrough/Git.java
```diff
@@ -0,0 +1,87 @@
...
+    cloneCommands.add(gitRepoUrl);
+    try {
+      Context.getConfig().getCommandRunnerOption().getRunner().runToolCommand(cloneCommands);
+    } catch (PassthroughException e) {
+      OUT.println("Git clone for " + gitRepoUrl + " failed");
+    }
+  }
+
+  @Override
+  protected boolean requiresLogin() {
```
jaycarlton, line 84: Without logging in, can you even see your resources? Does this leak information about the workspace?

src/test/java/unit/PassthroughApps.java
```diff
@@ -191,6 +195,46 @@ void nextflowVersion() throws IOException {
...
+        "resource",
+        "add-ref",
+        "git-repo",
+        "--name=repo1",
+        "--repo-url=https://github.com/DataBiosphere/terra-example-notebooks.git");
+
+    // `terra git clone -a`
+    TestCommand.runCommandExpectSuccess("git", "clone", "-a");
+
+    assertTrue(Files.exists(Paths.get(System.getProperty("user.dir"), "terra-example-notebooks")));
```
jaycarlton, line 214: I'd check for a `.git` directory directly under that.

src/main/java/bio/terra/cli/command/app/passthrough/Git.java
```diff
@@ -0,0 +1,87 @@
...
+    if (cloneAll && names != null && names.length > 0) {
+      throw new UserActionableException(
+          "Conflicted input argument. Only specify one depending on "
+              + "whether you'd like to clone all or some of the git repo resources in this workspace");
+    }
+    if (cloneAll) {
+      validateCloneCommand();
+      Context.requireWorkspace().getResources().stream()
+          .filter(resource -> Resource.Type.GIT_REPO == resource.getResourceType())
+          .forEach(resource -> cloneGitRepoResource(resource));
```
jaycarlton, line 51: You need to check for duplicates in the repo names. Probably want to check the file system for a directory of the same name (e.g. if a user tries to clone the same  repo twice). Might also want an argument to specify the directory name of the clone (in case they have a legitimate desire to clone something more than once).

jaycarlton:
Also add instructions & examples to the README file.

src/main/java/bio/terra/cli/command/app/passthrough/Git.java
yuhuyoyo: i'd argue for all because the command would be like terra git clone --all and --clone-all would be slightly more repetitive.

docker/Dockerfile
yuhuyoyo: good point

src/main/java/bio/terra/cli/command/app/passthrough/Git.java
yuhuyoyo: ah i added this when i was trying to not add any workspace context to the env var for none cloning related commands but that ends up to be too hairy to disentangle. removing this.

src/main/java/bio/terra/cli/command/app/passthrough/Git.java
yuhuyoyo: it should be handled by git itself. if the file already exists, git will throw error and we will just go on to the next one.

cbookg (approved):

src/main/java/bio/terra/cli/command/app/passthrough/Git.java
jaycarlton: So you won't throw a `SystemException` if one of the invocations of `cloneGitRepoResource()` gives a non-zero exit code?

jaycarlton (approved):

src/main/java/bio/terra/cli/command/app/passthrough/Git.java
yuhuyoyo: yeah. error output from git will still be printed, but we should proceed to the next git resources if there's any.

PR #214, opened 2022-02-25
[PF-1333]: Fix bug when the log-in mode get lost in the follow-on command

yuhuyoyo:
User should stays in the same log-in mode in the follow-on command after they log in.


cbookg (approved):

jaycarlton (approved):
Interesting. I didn't realize we had follow-on commands.

yuhuyoyo:
> Interesting. I didn't realize we had follow-on commands.

by saying follow-on command, i just mean something like
terra auth login --mode=APP_DEFAULT_CREDENTIALS
terra workspace create

PR #215, opened 2022-02-25
[PF-1368] remove numResources from workspace JSON output

jaycarlton:
When we build a UFWorkspace object during `terra workspace list`, the backing API object doesn't have a resources list. This causes the  `numResources` field to be zero, which is incorrect. In the text output for workspace list, we don't show that field, so it's no issue, but the JSON was unreliable. Rather than try to make that field nullable and intelligently track whether the resources array in the `Workspace` class was actually populated versus just initialized with an empty list, I decided to just drop the field from the JSON.

This is a breaking fix for CLI users unfortunately.

### DIFF
--- a/src/main/java/bio/terra/cli/businessobject/Context.java
+++ b/src/main/java/bio/terra/cli/businessobject/Context.java
@@ -234,11 +234,7 @@ public class Context {
   }
 
   public static Workspace requireWorkspace() {
-    return getWorkspace()
-        .orElseThrow(
-            () -> {
-              throw new UserActionableException("No workspace set.");
-            });
+    return getWorkspace().orElseThrow(() -> new UserActionableException("No workspace set."));
   }
 
   public static void setWorkspace(Workspace workspace) {
--- a/src/main/java/bio/terra/cli/serialization/userfacing/UFWorkspace.java
+++ b/src/main/java/bio/terra/cli/serialization/userfacing/UFWorkspace.java
@@ -2,6 +2,7 @@ package bio.terra.cli.serialization.userfacing;
 
 import bio.terra.cli.businessobject.Workspace;
 import bio.terra.cli.utils.UserIO;
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
 import java.io.PrintStream;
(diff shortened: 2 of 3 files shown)


zloery (commented):
Trying to make sure I understand: what does it mean that this is a breaking change for the CLI? It seems like users won't be able to deserialize old UFWorkspace objects from JSON, so they'll need to set their workspace again after upgrading the CLI (and may get an obtuse Jackson error if they don't). Is that right?

jaycarlton:
I meant it's breaking if users had scripted something that depends on the `numResources` field in JSON. Deserializing shouldn't care about whether that field is persisted or not when reading it, if I understand `@JsonIgnore` correctly.

zloery (approved):

jmczerk (approved):

src/main/java/bio/terra/cli/businessobject/Context.java
```diff
@@ -234,11 +234,7 @@ public static void setUser(User user) {
   }
 
   public static Workspace requireWorkspace() {
-    return getWorkspace()
-        .orElseThrow(
-            () -> {
-              throw new UserActionableException("No workspace set.");
-            });
+    return getWorkspace().orElseThrow(() -> new UserActionableException("No workspace set."));
```
cbookg, line 237: remove unrelated change from this PR?

src/main/java/bio/terra/cli/serialization/userfacing/UFWorkspace.java
```diff
@@ -56,6 +61,12 @@ public void print() {
     OUT.println(
         "Cloud console: https://console.cloud.google.com/home/dashboard?project="
             + googleProjectId);
+    // Unfortunately, this UFWorkspace may have been built from a Workspace object built from a
+    // WorkspaceDescription, which does not provide the resources or their count. In that
```
cbookg, line 65: nit: I found this sentence hard to follow.

Maybe this comment could just be shortened to say that this field is in the text output, but not the json output because it won't be consistently present.

But I was surprised to see it removed from the 'describe' json.  As a developer if the text output had extra fields not present in the JSON that would be annoying if I wanted those.  I was expecting this PR to only remove the field from the "list" JSON.

It seems like having the field missing from "describe" is perhaps worse than having it be "0" in the list?

jaycarlton: In order to have a different output for describe vs list, I need to introduce a new backing type, like `UFWorkspaceLight` that has everything except the `numResources`. I'd have no real way of ensuring the real `UFWorkspace` is always built with the resources populated (at least under maintenance), and this adds some complexity to the structure, but it's doable. I don't know if I can safely have `UFWorkspace extends UFWorkspaceLight` because of the Jackson builder stuff.

PR #217, opened 2022-02-25
PF-1369: delete the notebook startup script in CLI repo

yuhuyoyo:
the script is copied to the wsm repo now. Also the script in WSM is updated so this one is outdated.


src/main/java/bio/terra/cli/command/resource/create/GcpNotebook.java
```diff
@@ -57,8 +57,6 @@
 
   @CommandLine.Option(
       names = "--post-startup-script",
-      defaultValue =
```
jaycarlton, line 60: Should we instead default to the URL of the one hosted in `terra-workspace-manager`?

jaycarlton (approved):

src/main/java/bio/terra/cli/command/resource/create/GcpNotebook.java
yuhuyoyo: i did set the default in wsm when creating an instance. Are you worried that the CLI might be talking to an old wsm?

jaycarlton: No, I was worried there wouldn't be a script. This argument overrides the other one I suppose.

Maybe put somewhere in the README where the canonical location is so folks know what's there and have an idea how to write their own.

zloery (approved):
Just double-checked and all Broad + Verily envs have the WSM change deployed, LGTM

PR #219, opened 2022-03-07
[PF-1289] Stop using pet SA key in unit tests

melissachang:
In Verily deployment, usage of pet SA keys is prohibited for security reasons.

If you look at [auth table in `CONTRIBUTING.md`](https://github.com/DataBiosphere/terra-cli/blob/mc/cli-preprod-tests/CONTRIBUTING.md#auth-overview), this PR involves `Unit test` row, `Tool command` column. Next PR will have integration tests stop using pet SA key, as well as remove all code involving pet SA keys.

We're lucky that a month ago, [`gcloud` added support](https://cloud.google.com/sdk/docs/release-notes#cloud_sdk_2) for specifying access token (`CLOUDSDK_AUTH_ACCESS_TOKEN`) instead of SA key. Before then, this PR wouldn't be possible. I bumped `CLOUD_SDK_VERSION` in `Dockerfile` to pick up this `gcloud` feature. Pure `gcloud` commands only need access token. However, `bq` and `gsutil` are not pure `gcloud` commands.

`bq`: Instead of pet SA key, we now use access token + refresh token. So I added refresh tokens to vault/GHA secrets. Ideally we would just need access token and not refresh token. (Storing refresh token is as bad as storing pet SA key.) However, when support for `CLOUDSDK_AUTH_ACCESS_TOKEN` was added, it did not completely include `bq`. `bq` recognizes `CLOUDSDK_AUTH_ACCESS_TOKEN`, but it also requires `adc.json` which has refresh token.

`gsutil`: Instead of pet SA key, we now use refresh token. See [auth table](https://github.com/DataBiosphere/terra-cli/blob/mc/cli-preprod-tests/CONTRIBUTING.md#auth-overview) for details.

Manually tested: Tested `terra bq, gcloud, gsutil, nextflow` in both `DOCKER_CONTAINER` and `LOCAL_PROCESS`.


CONTRIBUTING.md
```diff
@@ -474,3 +483,127 @@ that should be reported, so there's no need to make these messages readable to t
...
+
+There are 3 versions of `gsutil`:
+
+1. Standalone `gsutil`. This is the oldest, and predates `gcloud`.
+2. `gsutil` as part of gcloud
+3. `gcloud alpha storage`. This is the newest. [Faster than 2.](https://stackoverflow.com/collectives/google-cloud/articles/68475140/faster-cloud-storage-transfers-using-the-gcloud-command-line)
+
+This repo uses 2. To get auth to work for 2, we write the `~/.config/gcloud/legacy_credentials/default/.boto` that  `google-cloud-sdk/bin/bootstrapping/gsutil.py` expects. This file has refresh token.
+
+For 3, auth is simpler. You just set `CLOUDSDK_AUTH_ACCESS_TOKEN`  (which we are already setting for other gcloud pass-through commands).
```
zloery, line 584: Do we use `gcloud alpha storage` in this repo or is this noting a future improvement?

src/test/java/harness/TestUser.java
```diff
@@ -62,27 +78,97 @@
...
   public void login() throws IOException {
+    System.out.println("Logging in test user: " + email);
+
     // get domain-wide delegated credentials for this user. use the same scopes that are requested
     // of CLI users when they login.
-    System.out.println("Logging in test user: " + email);
-    GoogleCredentials delegatedUserCredential = getCredentials(User.USER_SCOPES);
+    GoogleCredentials googleCredentials = getCredentials(User.USER_SCOPES);
+
+    // We're not using domain-wide delegation SA key file (for security reasons), so auth is more
```
zloery, line 87: This comment is a little confusing to me since `getCredentials` is still using the domain-wide delegation SA key file to set `.terra/StoredCredential`. It's only `writeBqCredentialFile` and `writeGsUtilCredentialFile` which don't use this.

melissachang: You're right, thanks. Fixed comment and moved.

zloery (approved):

CONTRIBUTING.md
```diff
@@ -474,3 +482,126 @@ that should be reported, so there's no need to make these messages readable to t
...
+
+At the beginning of each test, before running terra CLI, [test user is logged in](https://github.com/DataBiosphere/terra-cli/blob/8adf7cdaaa1f74f9407c10cccc8f7c0c4623eb6b/src/test/java/harness/baseclasses/SingleWorkspaceUnit.java#L30).
+
+`TestUser.login()` [writes `.terra/StoredCredential`](https://github.com/DataBiosphere/terra-cli/blob/8adf7cdaaa1f74f9407c10cccc8f7c0c4623eb6b/src/test/java/harness/TestUser.java#L76), using domain-wide delegation to avoid browser flow.
+</td>
+<td>
+
+**Test user**
+
+Auth is complicated because we are not using Pet SA key (which is a security risk).
```
cbookg, line 555: You already describe above about not using Pet SA.

Maybe the caption here should be something along the lines of "Test authenticate using an end-user test account"?

src/main/java/bio/terra/cli/app/CommandRunner.java
```diff
@@ -21,10 +20,8 @@
 public abstract class CommandRunner {
   private static final Logger logger = LoggerFactory.getLogger(CommandRunner.class);
 
-  // the Java system property that allows tests to specify a SA key file for the ADC and gcloud
-  // credentials
-  @VisibleForTesting
-  public static final String CREDENTIALS_OVERRIDE_SYSTEM_PROPERTY = "TERRA_GOOGLE_CREDENTIALS";
+  // Only unit tests set this.
+  @VisibleForTesting public static final String TEST_USER_ACCESS_TOKEN = "TERRA_GOOGLE_CREDENTIALS";
```
cbookg, line 24: This feels like a generic env name for something specific to tests.  Maybe you just don't want to rename it to keep compatibilty?

src/main/java/bio/terra/cli/app/DockerCommandRunner.java
```diff
@@ -80,19 +80,11 @@ protected int runToolCommandImpl(String command, Map<String, String> envVars)
...
-      bindMounts.put(adcFileOnContainer, credentialsFileForTest.get());
-
-      // set the env var and gcloud auth credentials using the pet SA key file
-      envVars.put("GOOGLE_APPLICATION_CREDENTIALS", adcFileOnContainer.toString());
-      command =
-          "echo \"Setting the gcloud credentials to match the application default credentials\"; "
-              + "gcloud auth activate-service-account --key-file=${GOOGLE_APPLICATION_CREDENTIALS}; "
-              + command;
+    // For unit tests, set CLOUDSDK_AUTH_ACCESS_TOKEN. This is how to programmatically authenticate
+    // as test user, without SA key file
```
cbookg, line 84: drop the "without SA key file" part?

src/test/java/harness/TestUser.java
```diff
@@ -39,6 +52,9 @@
   public static final String CLOUD_PLATFORM_SCOPE =
       "https://www.googleapis.com/auth/cloud-platform";
 
+  private static final String GCLOUD_CLIENT_ID = "32555940559.apps.googleusercontent.com";
```
cbookg, line 55: Where does this come from?  Maybe put that in a comment?

src/test/java/harness/TestUser.java
```diff
@@ -62,27 +78,92 @@
...
+    System.out.println("Logging in test user: " + email);
+
     // get domain-wide delegated credentials for this user. use the same scopes that are requested
     // of CLI users when they login.
-    System.out.println("Logging in test user: " + email);
-    GoogleCredentials delegatedUserCredential = getCredentials(User.USER_SCOPES);
+    GoogleCredentials googleCredentials = getCredentials(User.USER_SCOPES);
+    writeTerraOAuthCredentialFile(googleCredentials);
+
+    // We're not using pet SA key file (for security reasons), so auth is more complicated.
```
cbookg, line 88: // We're authenticating as an end user, not a pet SA?

In general I'm making these comments because I feel like we should describe what we're authenticating as, where as saying "not SA" seems only relevant if you knew how it previously worked?

src/test/java/harness/TestUser.java
```diff
@@ -62,27 +78,92 @@
...
+    Files.createDirectories(path.getParent());
+    Files.write(path, json.toString().getBytes(), StandardOpenOption.CREATE);
+  }
 
-    // do the login flow to populate the global context with the current user
-    User.login();
+  /**
+   * Writes ~/.config/gcloud/legacy_credentials/default/.boto that
+   * google-cloud-sdk/bin/bootstrapping/gsutil.py expects.
+   *
```
cbookg, line 146: I don't understand this comment.  It looks like you're saying .boto is needed both above and below?

cbookg (approved):
LGTM, added some (optional) comment suggestions.

CONTRIBUTING.md
melissachang: I took out `(which is a security risk)`. I left in `Auth is complicated` part. The reason we have to do bespoke, complicated things for each tool is because we're not using SA key.

If there's a new tool, or if any these break for existing tools, it's non-trivial to figure out how to authenticate. It's not documented anywhere. I had to use pdb debugger to figure out what to do.

src/main/java/bio/terra/cli/app/CommandRunner.java
melissachang: I don't understand this comment. I changed to:

```
TEST_USER_ACCESS_TOKEN_PROPERTY = "TEST_USER_ACCESS_TOKEN";
```

Hopefully that's better.

src/main/java/bio/terra/cli/app/DockerCommandRunner.java
melissachang: I think it's important. In general, GCP only documents how to programmatically authenticate with SA key. CLOUDSDK_AUTH_ACCESS_TOKEN isn't documented anywhere.

src/test/java/harness/TestUser.java
melissachang: This isn't documented in GCP documentation, so linked to a Medium article

src/test/java/harness/TestUser.java
melissachang: I think these comments are useful if there's a new tool, or if what's we're doing for bq/gsutil breaks. GCP only documents SA key way / tools only cleanly support SA key way. It let's person know figuring out auth for a new tool might be hard.

src/test/java/harness/TestUser.java
melissachang: Deleted below

melissachang (commented):
Thanks

src/main/java/bio/terra/cli/app/CommandRunner.java
cbookg: I just meant the "TERRA_GOOGLE_CREDENTIALS" part.

PR #227, opened 2022-03-21
Fix failing notebook tests

melissachang:
Example failure:

https://scans.gradle.com/s/o5r6kcpudawae/tests/:runTestsWithTag/unit.WorkspaceOverride/notebooks()?top-execution=1

### DIFF
--- a/src/test/java/unit/GcpNotebookControlled.java
+++ b/src/test/java/unit/GcpNotebookControlled.java
@@ -283,7 +283,9 @@ public class GcpNotebookControlled extends SingleWorkspaceUnit {
                 "--name=" + resourceName,
                 "--workspace=" + workspaceId);
     assertEquals(notebookState, describeNotebook.state, "notebook state matches");
-    assertNotNull(describeNotebook.proxyUri, "proxy url is populated");
+    if (!notebookState.equals("PROVISIONING")) {
+      assertNotNull(describeNotebook.proxyUri, "proxy url is populated");
+    }
   }
 
   /**
--- a/src/test/java/unit/GcpNotebookControlled.java
+++ b/src/test/java/unit/GcpNotebookControlled.java
@@ -271,12 +271,8 @@ public class GcpNotebookControlled extends SingleWorkspaceUnit {
    * Helper method to call `terra resource describe` and assert that the notebook state matches that
    * given. Filters on the specified workspace id; Uses the current workspace if null.
    */
-  static void assertNotebookState(String resourceName, String notebookState, UUID workspaceId)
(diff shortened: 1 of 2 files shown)


src/test/java/unit/GcpNotebookControlled.java
```diff
@@ -273,6 +273,10 @@ private static void assertNotebookState(String resourceName, String notebookStat
    */
   static void assertNotebookState(String resourceName, String notebookState, UUID workspaceId)
       throws JsonProcessingException {
+    Preconditions.checkState(
+        !notebookState.equals("PROVISIONING"),
+        "Don't assert that state is PROVISIONING. This leads to flaky tests, because by the time you assert PROVISIONING, state might already be ACTIVE.");
```
jaycarlton, line 278: I'm confused. Wouldn't it also be flaky to assert that it's not provisioning anymore?

jaycarlton (commented):
Thanks. I had this on my flake list.

src/test/java/unit/GcpNotebookControlled.java
melissachang: That's a good point. I deleted this and instead made `assertNotebookState` private. So tests should only call `pollDescribeForNotebookState`, which isn't flaky.

melissachang (commented):
> Thanks. I had this on my flake list.

Yeah, I saw you reran last week and it passed. For both Verily and Broad, it failed the last 3 days. I guess GCP got faster. :)

src/test/java/unit/GcpNotebookControlled.java
```diff
@@ -284,9 +283,7 @@ static void assertNotebookState(String resourceName, String notebookState, UUID
                 "--name=" + resourceName,
                 "--workspace=" + workspaceId);
     assertEquals(notebookState, describeNotebook.state, "notebook state matches");
-    if (!notebookState.equals("PROVISIONING")) {
```
yuhuyoyo, line 287: can you keep this? since proxy url shouldn't exist for PROVISIONING state.

melissachang: Oops, done

yuhuyoyo (approved):

jaycarlton (approved):
Another approach would be to have the assertion take in a set of states, so you could assert that it's either PROVISIONING or ACTIVE.

melissachang:
> Another approach would be to have the assertion take in a set of states, so you could assert that it's either PROVISIONING or ACTIVE.

Thanks. I think it's a little cleaner to say "Here's the final state I care about". It doesn't matter what the intermediate states are; someone could add a new intermediate state in the future.

PR #228, opened 2022-03-23
Add `terra server set` to installation instructions

melissachang:
CLI defaults to a Broad test environment, which is confusing to real users.

I didn't add the part about "Here's how to talk to Broad production" because there is no Broad prod environment right now, only Verily.

I tried to make it so:

1. Real users won't skip this step
2. CLI developers know to skip this step

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,15 +42,12 @@
 To install the latest version:
 ```
 curl -L https://github.com/DataBiosphere/terra-cli/releases/latest/download/download-install.sh | bash && export SUPPRESS_GCLOUD_CREDS_WARNING=true
-./terra
+sudo mv terra /usr/local/bin
+# Talk to Verily production environment (as opposed to a test environment):
+terra server set --name=verily --quiet
 ```
 
-To install a specific version:
-```
-export TERRA_CLI_VERSION=0.106.0
-curl -L https://github.com/DataBiosphere/terra-cli/releases/latest/download/download-install.sh | bash
-./terra
-```
+To install a specific version: `export TERRA_CLI_VERSION=0.106.0`
 
 By default, the CLI will be installed without support for Docker (i.e. it won't pull the Docker image).
 The TERRA_CLI_DOCKER_MODE environment variable controls Docker support. Set it to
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -42,15 +42,12 @@
 To install the latest version:
 ```
 curl -L https://github.com/DataBiosphere/terra-cli/releases/latest/download/download-install.sh | bash && export SUPPRESS_GCLOUD_CREDS_WARNING=true
-./terra
+sudo mv terra /usr/local/bin
```
wnojopra, line 45: There's a line later on in the README that says "This will install the Terra CLI in the current directory. Afterwards, you may want to add it to your $PATH directly or move it to a place that is already on your $PATH (e.g. /usr/local/bin)". I'd prefer it that way so that the use of `sudo` is an understood decision by the user.

README.md
```diff
@@ -42,15 +42,12 @@
...
+terra server set --name=verily --quiet
 ```
 
-To install a specific version:
-```
-export TERRA_CLI_VERSION=0.106.0
-curl -L https://github.com/DataBiosphere/terra-cli/releases/latest/download/download-install.sh | bash
-./terra
-```
+To install a specific version: `export TERRA_CLI_VERSION=0.106.0`
```
wnojopra, line 50: Don't you still need to run the curl command after setting the TERRA_CLI_VERSION variable?

wnojopra (approved):
Thanks for adding this! Two comments inline.

README.md
melissachang: I added comment `# Optional: Move to /usr/local/bin` and deleted that paragraph.

My motivation: There's a wall of text that people might read for their first installation, but they can skip for subsequent installations:

image

For the things that need to be done **every** installation, I put them in one section for convenience:

image

WDYT? (I assume everyone runs `sudo mv terra /usr/local/bin`, is that not the case?)

README.md
melissachang: Changed to `To install a specific version, run before above commands:`

README.md
wnojopra: Nice! I like having a "straight to the point" section for subsequent installs. 

I think installing binaries to a location that then get added to your `PATH` is more common than sudo moving to /usr/local/bin. This is what `gcloud` suggests, for example: https://cloud.google.com/sdk/docs/install

cbookg (approved):

jaycarlton (approved):

PR #229, opened 2022-03-29
[PF-1428] Fix/remove tests

jaycarlton:
Now that access checks on reference creation and update have been removed, these tests started failing because of expected failures that turned into successes. Deleted two of them in the interest of saving time, as we already have coverage of updating references.

### DIFF
--- a/src/test/java/unit/FineGrainedAccessGcsObjectReference.java
+++ b/src/test/java/unit/FineGrainedAccessGcsObjectReference.java
@@ -2,7 +2,6 @@ package unit;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNull;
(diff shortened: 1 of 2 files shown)


jmczerk (approved):

src/test/java/unit/FineGrainedAccessGcsObjectReference.java
```diff
@@ -185,7 +183,7 @@ void updateObjectReferenceWithBucketLevelAccess() throws IOException {
 
   @Test
   @DisplayName("user tries to update reference of a bucket object that they don't have access to")
-  void attemptToUpdatePrivateBucketObject() throws IOException {
+  void updatePrivateBucketObject() throws IOException {
```
zloery, line 186: Should this one still be `attempt...`? It looks like it's expected to fail since workspace READERS can't modify referenced resources

zloery (approved):

PR #230, opened 2022-04-06
[PF-1441, PF-1440] Add command for getting and regenerating ssh key

yuhuyoyo:
ecm is not yet deployed beyond devel in the verily environment yet so the server files are not yet updated for those environment.


src/main/java/bio/terra/cli/command/user/sshkey/Get.java
```diff
@@ -0,0 +1,27 @@
...
+
+import bio.terra.cli.command.shared.BaseCommand;
+import bio.terra.cli.command.shared.options.Format;
+import bio.terra.cli.serialization.userfacing.UFSshKey;
+import bio.terra.cli.service.ExternalCredentialsManagerService;
+import bio.terra.externalcreds.model.SshKeyPairType;
+import picocli.CommandLine;
+import picocli.CommandLine.Command;
+
+@Command(name = "get", description = "Get a terra generated and managed ssh key.")
```
zloery, line 11: This is actually "get, or create if it doesn't exist" which I wouldn't expect from the command `terra user sshkey get`. 
Should `get and `create` be different commands, or is there a way to make this command's function clearer?

src/main/java/bio/terra/cli/service/ExternalCredentialsManagerService.java
```diff
@@ -0,0 +1,162 @@
...
+    return new ExternalCredentialsManagerService(
+        Context.requireUser().getUserAccessToken(), Context.getServer());
+  }
+
+  /**
+   * Gets Ssh key pair from ECM. If one does not exist yet, generate one.
+   *
+   * @param keyPairType git server tied to the key (e.g. GitHub, GitLab, Azure).
+   */
+  public SshKeyPair getSshKeyPair(SshKeyPairType keyPairType) {
```
zloery, line 54: nit: maybe name something like `getOrCreateSshKeyPair`?

src/main/java/bio/terra/cli/service/ExternalCredentialsManagerService.java
```diff
@@ -0,0 +1,162 @@
...
+   * @param errorMsg error message for the the {@link SystemException} that wraps any exceptions
+   *     thrown by the ECM client or the retries
+   */
+  private <T> T handleClientExceptions(
+      HttpUtils.SupplierWithCheckedException<T, HttpStatusCodeException> makeRequest,
+      String errorMsg) {
+    try {
+      return makeRequest.makeRequest();
+    } catch (HttpStatusCodeException | InterruptedException ex) {
+      // if this is an ECN client exception, check for a message in the response body
```
zloery, line 115: ECM*

src/main/java/bio/terra/cli/service/ExternalCredentialsManagerService.java
```diff
@@ -0,0 +1,162 @@
...
+    logger.error(
+        "ECM exception status code: {}, response body: {}, message: {}",
+        ex.getStatusCode(),
+        ex.getResponseBodyAsString(),
+        ex.getMessage());
+
+    // try to deserialize the response body into an ErrorReport
+    var responseBody = ex.getResponseBodyAsString();
+
+    // if we found a SAM error message, then return it
```
zloery, line 156: ECM*

src/main/java/bio/terra/cli/service/ExternalCredentialsManagerService.java
```diff
@@ -0,0 +1,162 @@
...
+      return HttpUtils.callWithRetries(
+          () -> sshKeyPairApi.getSshKeyPair(keyPairType),
+          ExternalCredentialsManagerService::isRetryable);
+    } catch (HttpStatusCodeException | InterruptedException e) {
+      if (e instanceof HttpStatusCodeException) {
+        var httpStatusException = (HttpStatusCodeException) e;
+        logErrorMessage(httpStatusException);
+        if (httpStatusException.getStatusCode() == HttpStatus.NOT_FOUND) {
+          return callWithRetries(
+              () -> sshKeyPairApi.generateSshKeyPair(Context.requireUser().getEmail(), keyPairType),
```
cbookg, line 66: so i don't think we should auto-generate here.  for the cli i think we do want some kind of 'setup' action that can include this type of thing.  but for the commands i would to be a bit closer to the api.

src/main/java/bio/terra/cli/service/ExternalCredentialsManagerService.java
```diff
@@ -0,0 +1,162 @@
...
+
+  /**
+   * Execute a function that includes hitting ECM endpoints. If an exception is thrown by the ECM
+   * client or the retries, make sure the HTTP status code and error message are logged.
+   *
+   * @param makeRequest function with a return value
+   * @param errorMsg error message for the the {@link SystemException} that wraps any exceptions
+   *     thrown by the ECM client or the retries
+   */
+  private <T> T handleClientExceptions(
```
cbookg, line 109: nit: this seems like the kind of function that should be common, and not copied for each service.  but doesn't need to be solved here.

src/main/resources/servers/verily-devel.json
```diff
@@ -5,5 +5,6 @@
   "samUri": "https://terra-devel-sam.api.verily.com",
   "samInviteRequiresAdmin": true,
   "workspaceManagerUri": "https://terra-devel-wsm.api.verily.com",
-  "wsmDefaultSpendProfile": "wm-default-spend-profile"
+  "wsmDefaultSpendProfile": "wm-default-spend-profile",
+  "externalCredsUri": "https://terra-devel-ecm.api.verily.com"
```
cbookg, line 9: i feel like you should just add the other envs.  we know what the URLs will be.

src/main/java/bio/terra/cli/service/ExternalCredentialsManagerService.java
yuhuyoyo: Jenni suggested to not have a create because the UI doesn't have a create

src/main/java/bio/terra/cli/command/SshKey.java
```diff
@@ -0,0 +1,11 @@
+package bio.terra.cli.command;
+
+import bio.terra.cli.command.user.sshkey.Get;
+import bio.terra.cli.command.user.sshkey.Regenerate;
+import picocli.CommandLine.Command;
+
+@Command(
+    name = "sshkey",
```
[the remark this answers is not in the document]
jaycarlton: yes, see `bq-dataset`

src/main/java/bio/terra/cli/command/user/sshkey/Get.java
jaycarlton: nit: "terra-generated and -managed"

src/main/java/bio/terra/cli/command/user/sshkey/Get.java
```diff
@@ -0,0 +1,27 @@
...
+public class Get extends BaseCommand {
+
+  @CommandLine.Mixin Format formatOption;
+
+  @Override
+  protected void execute() {
+    var ecmService = ExternalCredentialsManagerService.returnFromContext();
+    var sshKeyPair = ecmService.getSshKeyPair(SshKeyPairType.GITHUB);
+    formatOption.printReturnValue(
+        UFSshKey.createUFSshKey(
```
jaycarlton, line 21: why doesn't this factory method simply take in an sshKeyPair?

src/main/java/bio/terra/cli/command/user/sshkey/Get.java
```diff
@@ -0,0 +1,27 @@
...
+
+@Command(name = "get", description = "Get a terra generated and managed ssh key.")
+public class Get extends BaseCommand {
+
+  @CommandLine.Mixin Format formatOption;
+
+  @Override
+  protected void execute() {
+    var ecmService = ExternalCredentialsManagerService.returnFromContext();
+    var sshKeyPair = ecmService.getSshKeyPair(SshKeyPairType.GITHUB);
```
jaycarlton, line 19: nit: I only use `var` when the type is explicitly listed on the right hand side.

src/main/java/bio/terra/cli/serialization/userfacing/UFSshKey.java
```diff
@@ -0,0 +1,67 @@
+package bio.terra.cli.serialization.userfacing;
+
+import bio.terra.cli.utils.UserIO;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+import java.io.PrintStream;
+
+@JsonDeserialize(builder = UFSshKey.Builder.class)
+public class UFSshKey {
```
jaycarlton, line 9: Maybe name this `UFSshKeyPair` since it has two keys in it. Same for business object, PD, etc.

src/main/java/bio/terra/cli/service/ExternalCredentialsManagerService.java
```diff
@@ -0,0 +1,162 @@
...
+              "failed to regenerate an ssh key");
+        }
+        throw httpStatusException;
+      }
+      throw new SystemException("Failed to get ssh key", e);
+    }
+  }
+
+  /** Regenerate an SshKey for the user. */
+  public SshKeyPair regenerateSshKeyPair(SshKeyPairType keyPairType) {
```
jaycarlton, line 76: If there's no difference between generate and regenerate, then I'd drop the "re".

src/test/java/unit/SshKey.java
```diff
@@ -0,0 +1,51 @@
...
+import harness.TestUser;
+import harness.baseclasses.ClearContextUnit;
+import java.io.IOException;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+@Tag("unit")
+public class SshKey extends ClearContextUnit {
+
+  TestUser testUser = TestUser.chooseTestUser();
```
jaycarlton, line 18: You can put the initialization of `testUser` and the login in a `@BeforeEach` method if you like.

jaycarlton (approved):
Since the values are sensitive, users might not want them appearing in cleartext in their terminal logs. Maybe have the option of writing the keys to files instead of on the command line? Just a thought.

src/main/java/bio/terra/cli/command/user/sshkey/Get.java
yuhuyoyo: I'm not opposed to having a create but because the UI auto create one, this is suggested by Jenni.

src/main/java/bio/terra/cli/serialization/userfacing/UFSshKey.java
```diff
@@ -0,0 +1,67 @@
...
+        .userEmail(userEmail)
+        .build();
+  }
+
+  /** Print out this object in text format. */
+  public void print() {
+    PrintStream OUT = UserIO.getOut();
+    OUT.println("User email: " + userEmail);
+    OUT.println("Private key: ");
+    OUT.println(privateSshKey);
```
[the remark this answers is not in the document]
yuhuyoyo: Yeah I debated on that since the JSON format will need to show both still so that i can access it in a script. I will remove the private ssh key from the output print and add an option to save the key in the ./ssh path.

jaycarlton (approved):

zloery (approved):

src/main/java/bio/terra/cli/command/user/sshkey/Generate.java
```diff
@@ -0,0 +1,46 @@
...
+      description = "Save the terra ssh key pair as file, skip printing out the key")
+  boolean saveToFile;
+
+  @Override
+  protected void execute() {
+    var ecmService = ExternalCredentialsManagerService.fromContext();
+    var sshKeyPair = ecmService.generateSshKeyPair(SshKeyPairType.GITHUB);
+    if (saveToFile) {
+      try {
+        BufferedWriter writer = new BufferedWriter(new FileWriter("terra_id_rsa"));
```
jaycarlton, line 31: nit: name this `privateWriter` and the other one `publicWriter`.

src/test/java/unit/SshKeyPair.java
```diff
@@ -0,0 +1,63 @@
...
+  @Override
+  protected void setupOnce() throws Exception {
+    super.setupOnce();
+    testUser = TestUser.chooseTestUser();
+    testUser.login();
+  }
+
+  @Test
+  void getSshKey() throws IOException {
+    testUser.login();
```
jaycarlton, line 31: `testUser` is already logged in. Similarly below.

jaycarlton (approved):

src/test/java/unit/SshKeyPair.java
yuhuyoyo: realized that i actually need to keep this in the test. loggin in in the setup doesn't work

PR #232, opened 2022-04-12
[PF-1441] Add an interactive prompt when the user wants to generate an ssh key.

yuhuyoyo:


src/main/java/bio/terra/cli/command/user/sshkey/Generate.java
```diff
@@ -22,8 +23,15 @@ public class Generate extends BaseCommand {
       description = "Save the terra ssh key pair as file, skip printing out the key")
   boolean saveToFile;
 
+  @CommandLine.Mixin ConfirmationPrompt confirmationPromptOption;
+
   @Override
   protected void execute() {
+    confirmationPromptOption.confirmOrThrow(
+        "Generating a new Terra Ssh key will replace the old Terra Ssh key if existed. "
```
jaycarlton, line 31: nit: "SSH", "if it exists."

src/main/java/bio/terra/cli/command/user/sshkey/Generate.java
```diff
@@ -22,8 +23,15 @@ public class Generate extends BaseCommand {
...
+            + "You must un-associate the old SSH public key from your GitHub account through here https://github.com/settings/keys. "
```
jaycarlton, line 32: "disassociate"

replace "through here" with "using"

jaycarlton (approved):

src/main/java/bio/terra/cli/command/user/sshkey/Generate.java
```diff
@@ -22,8 +23,15 @@ public class Generate extends BaseCommand {
       description = "Save the terra ssh key pair as file, skip printing out the key")
   boolean saveToFile;
 
+  @CommandLine.Mixin ConfirmationPrompt confirmationPromptOption;
+
   @Override
   protected void execute() {
+    confirmationPromptOption.confirmOrThrow(
+        "Generating a new Terra Ssh key will replace the old Terra SSH key if it exists. "
```
cbookg, line 31: Ssh -> SSH

src/main/java/bio/terra/cli/command/user/sshkey/Generate.java
```diff
@@ -22,8 +23,15 @@ public class Generate extends BaseCommand {
       description = "Save the terra ssh key pair as file, skip printing out the key")
   boolean saveToFile;
 
+  @CommandLine.Mixin ConfirmationPrompt confirmationPromptOption;
```
cbookg, line 26: maybe just "confirmationPrompt"?

src/main/java/bio/terra/cli/command/user/sshkey/Generate.java
```diff
@@ -22,8 +23,15 @@ public class Generate extends BaseCommand {
...
+
   @Override
   protected void execute() {
+    confirmationPromptOption.confirmOrThrow(
+        "Generating a new Terra Ssh key will replace the old Terra SSH key if it exists. "
+            + "You must disassociate the old SSH public key from your GitHub account using https://github.com/settings/keys. "
```
cbookg, line 32: I feel like this is misleading because you don't really have to do this. Maybe the message should be more proactive, like "You must add the new public key to your GitHub account..."

src/main/java/bio/terra/cli/command/user/sshkey/Generate.java
```diff
@@ -22,8 +23,15 @@ public class Generate extends BaseCommand {
       description = "Save the terra ssh key pair as file, skip printing out the key")
   boolean saveToFile;
 
+  @CommandLine.Mixin ConfirmationPrompt confirmationPromptOption;
+
   @Override
   protected void execute() {
+    confirmationPromptOption.confirmOrThrow(
```
cbookg, line 30: Would be nice if we only showed this when there was a key...

cbookg: Maybe not worth the extra complexity though...

yuhuyoyo: agree. i don't think it's worth another round trip

cbookg (approved):

PR #233, opened 2022-04-12
Fix failing test: convert email to lower cases before comparing

yuhuyoyo:


zloery (approved):
LGTM, but agreed I think you can just use `toLowerCase()`

PR #234, opened 2022-04-14
Add test troubleshooting tip

melissachang:
My Linux desktop was just reimaged. I had to run `sudo chmod 666 /var/run/docker.sock` to get tests to work.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,7 +16,7 @@
     * [Override context directory](#override-context-directory)
     * [Setup test users](#setup-test-users)
     * [Automated tests](#automated-tests)
-    * [Troubleshooting](#troubleshooting)
+    * [Troubleshooting](#troubleshooting-1)
     * [Debugging tips](#debugging-tips)
 4. [Docker](#docker)
     * [Pull an existing image](#pull-an-existing-image)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -269,7 +269,7 @@ GCP org; Verily deployment uses a project in Verily GCP org.
 
 #### Troubleshooting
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -16,6 +16,7 @@
     * [Override context directory](#override-context-directory)
     * [Setup test users](#setup-test-users)
     * [Automated tests](#automated-tests)
+    * [Troubleshooting](#troubleshooting-1)
```
jaycarlton, line 19: what's the -1 for?

CONTRIBUTING.md
```diff
@@ -266,6 +267,11 @@ By default, tests run against Broad deployment. To run against a different deplo
 For example, consider the project that external resources are created in. The Broad deployment uses a project in Broad
 GCP org; Verily deployment uses a project in Verily GCP org.
 
+#### Troubleshooting
+
+If you see `Connecting to Docker daemon failed. Check that Docker is installed and running.`
```
jaycarlton, line 272: If it's just for linux, then it would be good to specify that.

jaycarlton (approved):

CONTRIBUTING.md
melissachang: There's 2 `Troubleshooting` sections. Without the `-1`, it takes you to first one:

https://github.com/DataBiosphere/terra-cli/blob/main/CONTRIBUTING.md#troubleshooting

CONTRIBUTING.md
melissachang: Done.

I *think* this may only affect Googlers. On linux, we install docker differently from others. But I figure this is the best place, rather than some private Google place.

cbookg (approved):

PR #236, opened 2022-04-15
Only test the environment with available ecm server

yuhuyoyo:
https://github.com/verily-src/terra-tool-cli/runs/6035813739?check_suite_focus=true

### DIFF
--- a/src/test/java/unit/SshKeyPair.java
+++ b/src/test/java/unit/SshKeyPair.java
@@ -5,8 +5,6 @@ import static org.junit.jupiter.api.Assertions.assertNotEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 
 import bio.terra.cli.serialization.userfacing.UFSshKeyPair;
-import bio.terra.cli.serialization.userfacing.UFStatus;
-import com.fasterxml.jackson.core.JsonProcessingException;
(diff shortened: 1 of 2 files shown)


src/test/java/unit/SshKeyPair.java
```diff
@@ -18,54 +22,70 @@ public class SshKeyPair extends SingleWorkspaceUnit {
   private TestUser testUser = TestUser.chooseTestUser();
   private TestUser testUser2 = TestUser.chooseTestUserWhoIsNot(testUser);
 
+  // TODO (PF-1497): Remove this once ECM is deployed to all environment in verily.
+  private static final List<String> AVAILABLE_VERILY_ECM_SERVER = ImmutableList.of("verily-devel");
+
```
jaycarlton, line 27: Did you look at a `@DisableIf` and `@DisabledIfEnvironmentVariable`? If we had an environment variable for the server (and I bet we do somewhere), we could use that at the test level.

jaycarlton (approved):

zloery (approved):

src/test/java/unit/SshKeyPair.java
yuhuyoyo: i didn't know. changed to DisabledIfEnvironmentVariable instead.

PR #240, opened 2022-05-09
[PF-1620] Consider credentials expired 60 seconds early.

cbookg:
This ensures we don't skip the refresh and then have the credentials
expire mid-test or mid-commands.

### DIFF
--- a/src/main/java/bio/terra/cli/businessobject/User.java
+++ b/src/main/java/bio/terra/cli/businessobject/User.java
@@ -328,9 +328,9 @@ public class User {
 
     // check if the token is expired
     logger.debug("Access token expiration date: {}", accessToken.getExpirationTime());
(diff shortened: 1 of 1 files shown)


src/main/java/bio/terra/cli/businessobject/User.java
```diff
@@ -325,7 +328,9 @@ public boolean requiresReauthentication() {
 
     // check if the token is expired
     logger.debug("Access token expiration date: {}", accessToken.getExpirationTime());
-    return accessToken.getExpirationTime().compareTo(new Date()) <= 0;
+    Date toCheck = new Date();
```
yuhuyoyo, line 331: what about rename to cutOffDate?

yuhuyoyo (approved):

PR #245, opened 2022-05-10
googleGuest wifi will cause could not login

ginaygoogle:
image
The log file shows the [error] network can’t reachable. Because used WIFI from googleGuest. 
Change to Google-A or others will be solved. 

Just add one line in CONTRIBUTING.md after talked with Jay.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -59,6 +59,7 @@ listed in the command usage.
 #### Troubleshooting
 - Wipe the global context directory. `rm -R $HOME/.terra`.
 - Re-run the setup script. `source tools/local-dev.sh`.
+- Do not use GoogleGuest wifi in the google office.
 
 
 ### Publish a release


CONTRIBUTING.md
```diff
@@ -59,6 +59,7 @@ listed in the command usage.
 #### Troubleshooting
 - Wipe the global context directory. `rm -R $HOME/.terra`.
 - Re-run the setup script. `source tools/local-dev.sh`.
+- Do not use GoogleGuest wifi in the google office.
```
cbookg, line 62: But why?  Do we know the reason?

melissachang (commented):
Nice catch!

Could we put this in Terra onboarding doc instead? This repo is public, `Do not use GoogleGuest wifi in the google office.` is specific to Google.

CONTRIBUTING.md
ginaygoogle: I do not know... Maybe the googleGuest wifi has some special restrictions.

ginaygoogle:
Close this PR and change this to Terra Onboarding doc. Thanks Melissa.

PR #246, opened 2022-05-10
[PF-886] Clean up test workspaces with Janitor

zloery:
This change modifies all unit tests which create real workspaces to attempt to register them with Janitor to ensure they're always cleaned up.  This is currently only configured for `broad-dev` which we use for PR and nightly tests - if anyone is using other personal/developer environments I can add those to the Janitor configuration later.

I also moved pubsub topic/projectID to test config, which we'll need later for PF-963.


src/test/java/harness/utils/WorkspaceUtils.java
```diff
@@ -0,0 +1,48 @@
...
+/** Utilities for working with workspaces in CLI tests. */
+public class WorkspaceUtils {
+
+  /**
+   * Create a new workspace and register it with Janitor if this test is running in an environment
+   * where Janitor is enabled.
+   *
+   * @param workspaceCreator The user who owns the workspace and should be used to clean it up.
+   */
+  public static UFWorkspace createWorkspaceWithCleanup(TestUser workspaceCreator)
```
melissachang, line 18: Drive-by review. I'm also creating a WorkspaceUtils in my branch. :)

Can this be `createWorkspace` instead of `createWorkspaceWithCleanup`?

zloery: honestly I'm a little surprised it didn't exist already. and sg, done

src/test/java/harness/CRLJanitor.java
```diff
@@ -26,6 +45,18 @@ public class CRLJanitor {
 
   private static final String DEFAULT_CLIENT_NAME = "cli-test";
 
+  // How long Janitor should wait before cleaning test workspaces. It might be useful
+  // to keep these workspaces around long enough to debug, this should be lowered if not.
+  private static final Duration WORKSPACE_TIME_TO_LIVE = Duration.ofHours(48);
+
+  // Map from CLI server to Janitor's identifier for WSM instance.
+  public static final Map<String, String> serverToWsmInstanceIdentifier =
+      ImmutableMap.of("broad-dev", "dev");
```
melissachang, line 54: Opt - Hmm, in your Janitor config map, should we switch from `dev` to `broad-dev`, so we don't have to maintain this map?

Or, instead of maintaining 2 maps:
- In Janitor config, instead of map from WSM nickname to WSM url, just have a list of WSM urls
- In test config, put WSM url

src/test/java/harness/TestConfig.java
```diff
@@ -24,6 +24,12 @@ public final class TestConfig {
   // see PF-886.
   private boolean useJanitorForExternalResourcesCreatedByTests;
```
melissachang, line 25: Now that Janitor cleans up workspaces, can you rename this to `useJanitor` ?

src/test/java/harness/TestConfig.java
```diff
@@ -24,6 +24,12 @@ public final class TestConfig {
...
 
+  // The PubSub topic to use for publishing Janitor cleanup requests
+  private String janitorPubSubTopic;
```
melissachang, line 28: Optional - This is the same for all envs right? Can this be a constant instead of test config?

src/test/java/harness/utils/WorkspaceUtils.java
```diff
@@ -0,0 +1,48 @@
...
+        TestCommand.runAndParseCommandExpectSuccess(UFWorkspace.class, "workspace", "create");
+    CRLJanitor.registerWorkspaceForCleanupIfEnabled(workspace, workspaceCreator);
+    return workspace;
+  }
+
+  /**
+   * Create a new workspace and register it with Janitor if this test is running in an environment
+   * where Janitor is enabled.
+   *
+   * @param workspaceCreator The user who owns the workspace and should be used to clean it up.
```
melissachang, line 31: opt My personal preference is to omit these `@param` javadocs. go/java-style says `Javadoc is optional for "simple, obvious" members`. Something like

```
* @param name The name of the workspace
```

isn't adding anything.

src/test/java/harness/CRLJanitor.java
```diff
@@ -51,4 +81,80 @@ private static GoogleCredentials getSACredentials() {
...
+      return;
+    }
+    registerWorkspaceForCleanup(workspace, testUser);
+  }
+
+  /**
+   * Register a given workspace to be cleaned up. Wrapper around {@link
+   * #registerWorkspaceForCleanup(UUID, String, String)}
+   */
+  public static void registerWorkspaceForCleanup(UFWorkspace workspace, TestUser testUser) {
```
melissachang, line 114: You have 3 methods:

1. registerWorkspaceForCleanupIfEnabled
2. registerWorkspaceForCleanup
3. registerWorkspaceForCleanup takes in instanceId

Callers should only call 1, right? Can these three combined into method?

```
  public static void registerWorkspaceForCleanup(
      UFWorkspace workspace, TestUser testUser) {
    if (!TestConfig.get().getUseJanitorForExternalResourcesCreatedByTests()) {
      return;
    }
    ...
```

3 throws an exception if `UseJanitorForExternalResourcesCreatedByTests` isn't true. I don't think we should throw an exception if `UseJanitorForExternalResourcesCreatedByTests` isn't true. We should be able to turn off `UseJanitorForExternalResourcesCreatedByTests` indepedent of code.

src/test/java/harness/utils/WorkspaceUtils.java
```diff
@@ -0,0 +1,48 @@
...
+   */
+  public static UFWorkspace createWorkspace(TestUser workspaceCreator)
+      throws JsonProcessingException {
+    // `terra workspace create --format=json`
```
melissachang, line 20: // `terra workspace create`

src/test/java/harness/utils/WorkspaceUtils.java
```diff
@@ -0,0 +1,48 @@
...
+   * Create a new workspace and register it with Janitor if this test is running in an environment
+   * where Janitor is enabled.
+   *
+   * @param workspaceCreator The user who owns the workspace and should be used to clean it up.
+   * @param name The name of the workspace
+   * @param description Description of the workspace
+   */
+  public static UFWorkspace createWorkspace(
+      TestUser workspaceCreator, String name, String description) throws JsonProcessingException {
+    // `terra workspace create --format=json --name=$name --description=$description`
```
melissachang, line 37: Take out `--format=json`

src/test/java/harness/CRLJanitor.java
zloery: Hm, good point. It's kind of clunky keeping a bunch of hardcoded strings in sync across multiple repos, keeping a url allowlist in Janitor config would be cleaner but would still prevent calls to arbitrary URLs.
opened https://github.com/DataBiosphere/terra-resource-janitor/pull/140 for this, I'll update this PR after that one's merged

src/test/java/harness/TestConfig.java
zloery: It's the same for all Broad environments but would be different for a Verily deployment, so I think this needs to live in TestConfig

src/test/java/harness/utils/WorkspaceUtils.java
zloery: Good point. I think it's worth documenting that `workspaceCreator` will be impersonated in a WSM call here, but the other params aren't adding anything.

src/test/java/harness/CRLJanitor.java
zloery: Good point. I wrote all three thinking they would be useful, but after changing the tests I think this only needs the one method.

src/test/java/harness/utils/WorkspaceUtils.java
zloery: `runAndParseCommandExpectSuccess` always adds the "--format=json" argument, this is actually running // `terra workspace create --format=json`

src/test/java/harness/CRLJanitor.java
```diff
@@ -51,4 +80,59 @@ private static GoogleCredentials getSACredentials() {
       throw new RuntimeException("Error reading SA credentials for Janitor client.", ioEx);
     }
   }
+
+  private static Publisher initializeJanitorPubsubPublisher() {
```
melissachang, line 84: PubSub

src/test/java/harness/TestConfig.java
```diff
@@ -83,6 +89,12 @@ private static void validateTestConfig(TestConfig testConfig, String testConfigF
             && !testConfig.getProjectForExternalResources().isEmpty(),
         "In %s, projectForExternalResources must be set",
         testConfigFileName);
+    if (testConfig.useJanitor) {
+      Preconditions.checkState(
+          testConfig.janitorPubSubProjectId != null && testConfig.janitorPubSubTopic != null,
+          "In %s, if useJanitorForExternalResourcesCreatedByTests is set, janitorPubSubProjectId and janitorPubSubTopic must also be set.",
```
melissachang, line 95: useJanitor

src/test/java/harness/TestConfig.java
```diff
@@ -22,7 +22,13 @@ public final class TestConfig {
 
   // Whether to use janitor to clean up external resources. For more about terra-cli and Janitor,
```
melissachang, line 23: Update comment

melissachang (approved):
Thanks! Upcoming changes are straightforward, so approving now

src/test/java/harness/CRLJanitor.java
```diff
@@ -51,4 +80,59 @@ private static GoogleCredentials getSACredentials() {
...
+    try {
+      data = ByteString.copyFromUtf8(JacksonMapper.getMapper().writeValueAsString(janitorRequest));
+    } catch (IOException e) {
+      throw new JanitorException(
+          String.format("Failed to serialize CreateResourceRequestBody: [%s]", janitorRequest), e);
+    }
+    PubsubMessage janitorMessage = PubsubMessage.newBuilder().setData(data).build();
+    ApiFuture<String> messageIdFuture = publisher.publish(janitorMessage);
+    try {
+      // Wait for the Future to complete
```
jaycarlton, line 131: so `get()` waits for completion?

src/test/java/harness/baseclasses/SingleWorkspaceUnit.java
```diff
@@ -29,9 +30,7 @@ protected void setupOnce() throws Exception {
 
     workspaceCreator.login();
 
-    // `terra workspace create --format=json`
-    UFWorkspace createWorkspace =
-        TestCommand.runAndParseCommandExpectSuccess(UFWorkspace.class, "workspace", "create");
+    UFWorkspace createWorkspace = WorkspaceUtils.createWorkspace(workspaceCreator);
```
jaycarlton, line 33: nit: `createdWorkspace`

src/test/java/harness/utils/WorkspaceUtils.java
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  /**
+   * Create a new workspace and register it with Janitor if this test is running in an environment
+   * where Janitor is enabled.
+   *
+   * @param workspaceCreator The user who owns the workspace. This user will be impersonated to in
+   *     the WSM workspaceDelete request.
+   */
+  public static UFWorkspace createWorkspace(
```
jaycarlton, line 35: maybe name these `createTrackedWorkspace` so folks know they're special. I wouldn't have associated these calls with Janitor since the class is just `WorkspaceUtils`.

src/test/java/unit/CloneWorkspace.java
```diff
@@ -108,9 +109,7 @@ public void cloneWorkspace(TestInfo testInfo) throws Exception {
     workspaceCreator.login();
 
     // create a workspace
-    // `terra workspace create --format=json`
-    sourceWorkspace =
-        TestCommand.runAndParseCommandExpectSuccess(UFWorkspace.class, "workspace", "create");
+    sourceWorkspace = WorkspaceUtils.createWorkspace(workspaceCreator);
```
jaycarlton, line 112: for clone, the destination workspace won't be tracked automatically. We'll have to make sure the tests clean that up manually.

src/test/resources/testconfigs/broad.json
```diff
@@ -26,5 +26,7 @@
     }
   ],
   "projectForExternalResources": "terra-cli-test",
-  "useJanitorForExternalResourcesCreatedByTests": true
+  "useJanitor": true,
+  "janitorPubSubTopic": "crljanitor-tools-pubsub-topic",
```
jaycarlton, line 30: it's kinda redundant to have `-pubsub-topic` in a topic name, but I guess this exists already.

Also, is this config for all Broad envs?

jaycarlton (approved):

src/test/java/harness/utils/WorkspaceUtils.java
```diff
@@ -0,0 +1,48 @@
...
+      TestUser workspaceCreator, String name, String description) throws JsonProcessingException {
+    // `terra workspace create --format=json --name=$name --description=$description`
+    UFWorkspace workspace =
+        TestCommand.runAndParseCommandExpectSuccess(
+            UFWorkspace.class,
+            "workspace",
+            "create",
+            "--name=" + name,
+            "--description=" + description);
+    CRLJanitor.registerWorkspaceForCleanup(workspace, workspaceCreator);
```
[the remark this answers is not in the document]
melissachang: (Given that `runAndParseCommandExpectSuccess` and this method return `UFWorkspace`, I think it's ok to use here)

src/test/java/harness/CRLJanitor.java
zloery: yep!

src/test/java/harness/utils/WorkspaceUtils.java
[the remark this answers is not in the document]
zloery: Most tests run commands the same way users do (as full commands) rather than calling internal methods, so I think UF object is right here.

(Whether we really should be doing that in unit tests is a whole separate discussion, too...)

src/test/java/harness/utils/WorkspaceUtils.java
zloery: Both names seem reasonable but I'd lean towards keeping this short - from a test reader perspective it's not important that the test objects are cleaned up later, it's the `createWorkspace` behavior I care about.

Added a comment about leaking workspaces if tests don't use this method

src/test/java/unit/CloneWorkspace.java
zloery: Yeah, definitely. Those tests can call `CRLJanitor.registerWorkspaceForCleanup`, but we'll have to be sure we're actually doing that

src/test/resources/testconfigs/broad.json
zloery: agreed, but the topic already exists so 🤷 
yes, we plan to use the same Janitor instance (and same pubsub topic) to clean up after all envs, though that's currently just dev.

PR #249, opened 2022-05-12
[PF-1637] Reload resource after update

jaycarlton:
The update code doesn't change all the properties on the current resource. My solution was to reload the resource from the workspace, which bypasses the cache (apparently). Name and description didn't need this treatment, as they went through `bio.terra.cli.businessobject.Resource#updatePropertiesAndSync()`, which directly updates the properties on the resource business object.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,7 +23,8 @@
     * [Build a new image](#build-a-new-image)
     * [Publish a new image](#publish-a-new-image)
     * [Update the default image](#update-the-default-image)
-5. [Code structure](#code-structure)
+5. [Code](#code)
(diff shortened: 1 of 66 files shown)


melissachang (approved):
Could you add unit tests for BqTable, GcsObject, GitRepo? I feel like this part of the code is brittle.

jaycarlton:
> Could you add unit tests for BqTable, GcsObject, GitRepo? I feel like this part of the code is brittle.
The only place I've seen the issue is with cloning instructions, and those don't support cloning instructions update in the CLI yet. I'll just double check the options that they do support.

yuhuyoyo (approved):

PR #250, opened 2022-05-12
[PF-1335] Gradle Enterprise Trial Navigator Step 1

melissachang:
This sends build scans to https://gradle-enterprise-prod-helm-cluster.api.verily.com/ instead of scans.gradle.com. Here are scans from both local run and GHA:

image

@ghale

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -1,1 +1,18 @@
+plugins {
+    id "com.gradle.enterprise" version "3.10"
+    id 'com.gradle.common-custom-user-data-gradle-plugin' version '1.6.5'
+}
+
 rootProject.name = 'terra-cli'
+
+// Gradle Enterprise Trial Navigator Step 1
+gradleEnterprise {
+    server = "https://gradle-enterprise-prod-helm-cluster.api.verily.com"
+    buildScan {
+        publishAlways()
+        capture {
+            taskInputFiles = true
+        }
+        uploadInBackground = System.getenv("CI") == null
+    }
+}


jaycarlton (approved):

zloery (approved):

settings.gradle
```diff
@@ -1 +1,18 @@
...
+
+// Gradle Enterprise Trial Navigator Step 1
+gradleEnterprise {
+    server = "https://gradle-enterprise-prod-helm-cluster.api.verily.com"
+    buildScan {
+        publishAlways()
+        capture {
+            taskInputFiles = true
+        }
+        uploadInBackground = System.getenv("CI") == null
```
ghale, line 16: One question to answer is whether `CI` is a valid environment variable on your CI agents.  We use this as an example of detecting the environment in our docs, but it's just an example.  In any event, this needs to be _some_ environment variable that can be expected to be present for a build running on a CI agent.  [These](https://github.com/gradle/common-custom-user-data-gradle-plugin/blob/main/src/main/java/com/gradle/CustomBuildScanEnhancements.java#L226-L264) are the environment variables we look at in our custom user data plugin to detect different common CI vendors.  On the other hand, if `CI` _is_ expected to be present, then that should work fine as well.

melissachang: We use GHA. CI works ([docs](https://docs.github.com/en/actions/learn-github-actions/environment-variables#:~:text=CI,to%20true.)). I assume [these Gradle docs](https://docs.gradle.com/enterprise/gradle-plugin/#disabling_programmatically) use `CI` because it works for GHA.

PR #252, opened 2022-05-16
update gcloud version

yuhuyoyo:


jaycarlton (approved):

melissachang (approved):
In general, it's nice to explain why you're doing something. Why are you upgrading gcloud?

yuhuyoyo:
> In general, it's nice to explain why you're doing something. Why are you upgrading gcloud?

tbh this was random.. i was just updating my `gcloud` on my local machine and wanted to check what version of `gcloud` cli docker is using. I saw it's a little outdated so I updated it.

PR #253, opened 2022-05-16
PF1630: terra group list shows in table view with the test on 2 groups

ginaygoogle:
Screen Shot 2022-05-12 at 10 22 56 AM

### DIFF
--- a/src/main/java/bio/terra/cli/command/group/List.java
+++ b/src/main/java/bio/terra/cli/command/group/List.java
@@ -38,7 +38,8 @@ public class List extends BaseCommand {
   private enum UFGroupColumns implements ColumnDefinition<UFGroup> {
     NAME("NAME", g -> g.name, 30, LEFT),
     EMAIL("EMAIL", g -> g.email, 45, LEFT),
-    MEMBERS("MEMBERS", g -> g.numMembers.toString(), 7, RIGHT),
+    // numMembers will be null if we do not have permission to list the members
+    MEMBERS("MEMBERS", g -> g.numMembers == null ? "unknown" : g.numMembers.toString(), 7, RIGHT),
     POLICIES("POLICIES", g -> g.currentUserPolicies.toString(), 15, LEFT);
 
     private final String columnLabel;
--- a/src/main/java/bio/terra/cli/command/group/List.java
+++ b/src/main/java/bio/terra/cli/command/group/List.java
@@ -1,6 +1,7 @@
 package bio.terra.cli.command.group;
 
 import static bio.terra.cli.app.utils.tables.ColumnDefinition.Alignment.LEFT;
+import static bio.terra.cli.app.utils.tables.ColumnDefinition.Alignment.RIGHT;
 
 import bio.terra.cli.app.utils.tables.ColumnDefinition;
 import bio.terra.cli.app.utils.tables.TablePrinter;
@@ -36,7 +37,7 @@ public class List extends BaseCommand {
   /** Column information for fields in `resource list` output */
   private enum UFGroupColumns implements ColumnDefinition<UFGroup> {
     EMAIL("EMAIL", g -> g.email, 42, LEFT),
-    MEMBERS("MEMBERS", g -> g.numMembers.toString(), 7, LEFT),
+    MEMBERS("MEMBERS", g -> g.numMembers.toString(), 7, RIGHT),
     POLICIES("POLICIES", g -> g.currentUserPolicies.toString(), 10, LEFT);
 
     private final String columnLabel;
--- a/src/main/java/bio/terra/cli/command/group/List.java
+++ b/src/main/java/bio/terra/cli/command/group/List.java
@@ -1,11 +1,16 @@
 package bio.terra.cli.command.group;
 
+import static bio.terra.cli.app.utils.tables.ColumnDefinition.Alignment.LEFT;
+
+import bio.terra.cli.app.utils.tables.ColumnDefinition;
+import bio.terra.cli.app.utils.tables.TablePrinter;
 import bio.terra.cli.businessobject.Group;
 import bio.terra.cli.command.shared.BaseCommand;
 import bio.terra.cli.command.shared.options.Format;
 import bio.terra.cli.serialization.userfacing.UFGroup;
 import bio.terra.cli.utils.UserIO;
 import java.util.Comparator;
+import java.util.function.Function;
 import picocli.CommandLine;
 import picocli.CommandLine.Command;
 
@@ -24,8 +29,50 @@ public class List extends BaseCommand {
 
   /** Print this command's output in text format. */
   private static void printText(java.util.List<UFGroup> returnValue) {
-    for (UFGroup group : returnValue) {
-      group.print();
+    TablePrinter<UFGroup> printer = bio.terra.cli.command.group.List.UFGroupColumns::values;
+    OUT.println(printer.print(returnValue));
+  }
+
+  /** Column information for fields in `resource list` output */
+  private enum UFGroupColumns implements ColumnDefinition<UFGroup> {
+    EMAIL("EMAIL", g -> g.email, 20, LEFT),
+    MEMBERS("MEMBERS", g -> g.numMembers.toString(), 20, LEFT),
+    POLICIES("POLICIES", g -> g.currentUserPolicies.toString(), 20, LEFT);
+
+    private final String columnLabel;
+    private final Function<UFGroup, String> valueExtractor;
+    private final int width;
+    private final Alignment alignment;
+
+    UFGroupColumns(
+        String columnLabel,
+        Function<UFGroup, String> valueExtractor,
+        int width,
+        Alignment alignment) {
+      this.columnLabel = columnLabel;
+      this.valueExtractor = valueExtractor;
+      this.width = width;
+      this.alignment = alignment;
+    }
+
+    @Override
+    public String getLabel() {
+      return columnLabel;
+    }
+
+    @Override
+    public Function<UFGroup, String> getValueExtractor() {
+      return valueExtractor;
+    }
+
+    @Override
+    public int getWidth() {
+      return width;
+    }
+
+    @Override
+    public Alignment getAlignment() {
+      return alignment;
     }
   }
 }
--- a/src/main/java/bio/terra/cli/command/group/List.java
+++ b/src/main/java/bio/terra/cli/command/group/List.java
@@ -36,9 +36,10 @@ public class List extends BaseCommand {
 
   /** Column information for fields in `resource list` output */
   private enum UFGroupColumns implements ColumnDefinition<UFGroup> {
-    EMAIL("EMAIL", g -> g.email, 42, LEFT),
+    NAME("NAME", g -> g.name, 30, LEFT),
+    EMAIL("EMAIL", g -> g.email, 45, LEFT),
     MEMBERS("MEMBERS", g -> g.numMembers.toString(), 7, RIGHT),
-    POLICIES("POLICIES", g -> g.currentUserPolicies.toString(), 10, LEFT);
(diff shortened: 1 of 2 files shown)


src/main/java/bio/terra/cli/command/group/List.java
```diff
@@ -24,8 +29,50 @@ protected void execute() {
...
   private static void printText(java.util.List<UFGroup> returnValue) {
-    for (UFGroup group : returnValue) {
-      group.print();
+    TablePrinter<UFGroup> printer = bio.terra.cli.command.group.List.UFGroupColumns::values;
+    OUT.println(printer.print(returnValue));
+  }
+
+  /** Column information for fields in `resource list` output */
+  private enum UFGroupColumns implements ColumnDefinition<UFGroup> {
+    EMAIL("EMAIL", g -> g.email, 20, LEFT),
```
melissachang, line 38: It would be nice if email weren't cut off. Also, members and policies don't need to be so wide. Can you try:

email: 40
members: 9
policies: 10

Can you try this out in verily-devel and verily-prod, to make sure emails aren't cut off?

src/test/java/unit/Group.java
```diff
@@ -61,6 +63,17 @@ void listDescribeUsersReflectCreateDelete() throws IOException {
     // check that the group creator is included in the list and is an admin
     expectListedMemberWithPolicies(name, testUser.email, GroupPolicy.ADMIN);
 
+    // create another group
```
melissachang, line 66: In general, when you add a code change, we usually add a corresponding test. Usually the new test fails before the code change, and passes after the code change. In this case, the new test code isn't really related to your code change. The new test code passes before your code change.

I would just delete lines 66-76. @jaycarlton  ?

If you keep this, change `name1` to `name2`. (For the second thing of something, we usually go with `2` instead of `1`.)

melissachang (approved):

src/test/java/unit/Group.java
ginaygoogle: sorry. Melissa. I did wrong test code and this is a draft review. 

I talked with Jay. He suggests me to use regular expression to test the table by each row. and verify the head and table content. It fails in my local... I am debugging now.

ginaygoogle:
Screen Shot 2022-05-16 at 5 13 44 PM
email:42
members:7
policies:10

jaycarlton:
The members column would look better right aligned.

src/test/java/unit/Group.java
```diff
@@ -106,6 +106,28 @@ void listDescribeUsersReflectCreateDelete() throws IOException {
...
+    TestCommand.runAndParseCommandExpectSuccess(
+        UFGroup.class, "group", "create", "--name=" + name1);
+
+    TestCommand.Result cmd = TestCommand.runCommand("group", "list");
+
+    // use regular expression testing the table format and content inside
+    assertEquals(0, cmd.exitCode, "group list returned successfully");
+    String[] rows = cmd.stdOut.split("\\r?\\n");
+    String[] rowHead = rows[0].split("\\s+");
+    assertEquals(rowHead[0].trim().replace("\r", ""), "EMAIL");
```
jaycarlton, line 121: these should be the other way around, for example `assertEquals("EMAIL", ...)`. The expected value goes before the actual value.

src/test/java/unit/Group.java
```diff
@@ -106,6 +106,28 @@ void listDescribeUsersReflectCreateDelete() throws IOException {
...
+    assertEquals(rowHead[1].trim().replace("\r", ""), "MEMBERS");
+    assertEquals(rowHead[2].trim().replace("\r", ""), "POLICIES");
+
+    for (int i = 1; i < rows.length; i = i + 1) {
+      String[] rowi = rows[i].split("\\s+");
+      assertTrue(rowi[0].matches("^[a-zA-Z0-9_-]+@[a-zA-Z0-9_-]+(\\.[a-zA-Z0-9_-]+)+$"));
+      assertTrue(rowi[1].matches("[0-9]+"));
+      assertTrue(Arrays.asList("[ADMIN]", "[MEMBER]").contains(rowi[2]));
```
jaycarlton, line 129: Have we found a way to display more than one policy? That would break this check.

src/test/java/unit/Group.java
```diff
@@ -106,6 +106,28 @@ void listDescribeUsersReflectCreateDelete() throws IOException {
...
+    // create another group
+    String name1 = SamGroups.randomGroupName();
+    TestCommand.runAndParseCommandExpectSuccess(
+        UFGroup.class, "group", "create", "--name=" + name1);
+
+    TestCommand.Result cmd = TestCommand.runCommand("group", "list");
+
+    // use regular expression testing the table format and content inside
+    assertEquals(0, cmd.exitCode, "group list returned successfully");
+    String[] rows = cmd.stdOut.split("\\r?\\n");
```
jaycarlton, line 119: Assert that `cmd.stdErr` is `null` or empty.

ginaygoogle:
Talked with Jay. There is a bug on `terra group list`. after delete one group, when list again, it will failed due to could not find the deleted group.
image

src/test/java/unit/Group.java
```diff
@@ -96,16 +97,39 @@ void listDescribeUsersReflectCreateDelete() throws IOException {
...
+    String[] rowHead = rows[0].split("\\s+");
+    assertEquals("EMAIL", rowHead[0].trim().replace("\r", ""));
+    assertEquals("MEMBERS", rowHead[1].trim().replace("\r", ""));
+    assertEquals("POLICIES", rowHead[2].trim().replace("\r", ""));
+
+    for (int i = 1; i < rows.length; i = i + 1) {
+      String[] rowi = rows[i].split("\\s+");
+      assertTrue(rowi[0].matches("^[a-zA-Z0-9_-]+@[a-zA-Z0-9_-]+(\\.[a-zA-Z0-9_-]+)+$"));
+      assertTrue(rowi[1].matches("[0-9]+"));
+      assertTrue(Arrays.asList("[ADMIN]", "[MEMBER]").contains(rowi[2]));
```
jaycarlton, line 131: If there is a member too, it will print like `[ADMIN, MEMBER]` because you're doing `List.toString()`. That's where the square brackets are coming from.

jaycarlton (changes requested):

src/main/java/bio/terra/cli/command/group/List.java
```diff
@@ -24,8 +30,50 @@ protected void execute() {
...
   /** Print this command's output in text format. */
   private static void printText(java.util.List<UFGroup> returnValue) {
-    for (UFGroup group : returnValue) {
-      group.print();
+    TablePrinter<UFGroup> printer = bio.terra.cli.command.group.List.UFGroupColumns::values;
+    OUT.println(printer.print(returnValue));
+  }
+
+  /** Column information for fields in `resource list` output */
+  private enum UFGroupColumns implements ColumnDefinition<UFGroup> {
```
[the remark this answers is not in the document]
ginaygoogle: done

src/test/java/unit/Group.java
```diff
@@ -96,16 +97,39 @@ void listDescribeUsersReflectCreateDelete() throws IOException {
...
+
+    // create another group
+    String name2 = SamGroups.randomGroupName();
+    TestCommand.runAndParseCommandExpectSuccess(
+        UFGroup.class, "group", "create", "--name=" + name2);
+
+    TestCommand.Result cmd = TestCommand.runCommand("group", "list");
+
+    // use regular expression testing the table format and content inside
+    assertNull(cmd.stdErr);
```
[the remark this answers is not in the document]
ginaygoogle: done

src/test/java/unit/Group.java
ginaygoogle: done

src/test/java/unit/Group.java
```diff
@@ -106,6 +106,28 @@ void listDescribeUsersReflectCreateDelete() throws IOException {
...
+    assertEquals(0, cmd.exitCode, "group list returned successfully");
+    String[] rows = cmd.stdOut.split("\\r?\\n");
+    String[] rowHead = rows[0].split("\\s+");
+    assertEquals(rowHead[0].trim().replace("\r", ""), "EMAIL");
+    assertEquals(rowHead[1].trim().replace("\r", ""), "MEMBERS");
+    assertEquals(rowHead[2].trim().replace("\r", ""), "POLICIES");
+
+    for (int i = 1; i < rows.length; i = i + 1) {
+      String[] rowi = rows[i].split("\\s+");
+      assertTrue(rowi[0].matches("^[a-zA-Z0-9_-]+@[a-zA-Z0-9_-]+(\\.[a-zA-Z0-9_-]+)+$"));
```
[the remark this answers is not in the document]
ginaygoogle: done

src/test/java/unit/Group.java
ginaygoogle: done

jaycarlton:
> This PR is blocked by the bug. https://broadworkbench.atlassian.net/browse/PF-1652
You should be able to continue this PR despite that bug, since it's limited to a particular group in that one environment.

ginaygoogle:
> > This PR is blocked by the bug. https://broadworkbench.atlassian.net/browse/PF-1652
> 
> You should be able to continue this PR despite that bug, since it's limited to a particular group in that one environment.

Hi @jaycarlton  I do not quite understand here. We require pass 3 checks and get 2 approve to merge. but if there is a bug, we could not pass the 3 checks and could not merge. So that is the meaning I said I was blocked there.

src/test/java/unit/Group.java
```diff
@@ -96,16 +98,45 @@ void listDescribeUsersReflectCreateDelete() throws IOException {
...
+        UFGroup.class, "group", "create", "--name=" + name2);
+
+    TestCommand.Result cmd = TestCommand.runCommand("group", "list");
+
+    // use regular expression testing the table format and content inside
+    assertTrue(cmd.stdErr == null || cmd.stdErr.isEmpty());
+    assertEquals(0, cmd.exitCode, "group list returned successfully");
+    String[] rows = cmd.stdOut.split("\\n");
+    String[] rowHead = rows[0].split("\\s+");
+    assertEquals("NAME", rowHead[0].trim().replace("\r", ""));
```
jaycarlton, line 124: You don't need the `.trim.replace()` for these. The split takes care of the trimming, and there are no carriage returns coming from the table printer (or any other CLI output that I'm aware of).

src/test/java/unit/Group.java
```diff
@@ -293,6 +296,39 @@ private static void expectListedMemberWithPolicies(
...
+    String[] rowHead = rows[0].split("\\s+");
+    assertEquals("NAME", rowHead[0]);
+    assertEquals("EMAIL", rowHead[1]);
+    assertEquals("MEMBERS", rowHead[2]);
+    assertEquals("POLICIES", rowHead[3]);
+
+    // assert name, email and policies are correct
+    boolean nameAndEmailExists = false;
+
+    System.out.println(name);
```
jaycarlton, line 315: Remove these three debug `println` statements.

src/test/java/unit/Group.java
```diff
@@ -293,6 +296,39 @@ private static void expectListedMemberWithPolicies(
...
+      System.out.println(rows[i]);
+      // Convert to lower-case because email in broad.json is mixed case
+      if (rows[i].contains(email) && rows[i].contains(name)) {
+        nameAndEmailExists = true;
+        for (var policy : policies) {
+          assertThat("Row has all policies.", rows[i].contains(policy.toString()));
+        }
+      }
+    }
+    // name and email address for group member should be included in table output.
```
jaycarlton, line 327: put this comment inside the `assertTrue()` statement so it shows up when the assertion isn't met.

jaycarlton (approved):

PR #254, opened 2022-05-17
PF-1631: terra status format more readable

ginaygoogle:
Screen Shot 2022-05-16 at 11 31 55 PM

### DIFF
--- a/src/main/java/bio/terra/cli/serialization/userfacing/UFWorkspaceLight.java
+++ b/src/main/java/bio/terra/cli/serialization/userfacing/UFWorkspaceLight.java
@@ -52,12 +52,12 @@ public class UFWorkspaceLight {
   /** Print out a workspace object in text format. */
   public void print() {
     PrintStream OUT = UserIO.getOut();
-    OUT.println("Terra workspace id: " + id);
-    OUT.println("Display name: " + name);
-    OUT.println("Description: " + description);
-    OUT.println("Google project: " + googleProjectId);
+    OUT.println("Terra workspace id:  " + id);
+    OUT.println("Display name:        " + name);
+    OUT.println("Description:         " + description);
+    OUT.println("Google project:      " + googleProjectId);
     OUT.println(
-        "Cloud console: https://console.cloud.google.com/home/dashboard?project="
+        "Cloud console:       https://console.cloud.google.com/home/dashboard?project="
             + googleProjectId);
   }
 
--- a/src/main/java/bio/terra/cli/serialization/userfacing/UFWorkspaceLight.java
+++ b/src/main/java/bio/terra/cli/serialization/userfacing/UFWorkspaceLight.java
@@ -52,10 +52,10 @@ public class UFWorkspaceLight {
   /** Print out a workspace object in text format. */
   public void print() {
     PrintStream OUT = UserIO.getOut();
-    OUT.println("Terra workspace id:  " + id);
-    OUT.println("Display name:        " + name);
-    OUT.println("Description:         " + description);
-    OUT.println("Google project:      " + googleProjectId);
+    OUT.println("Id:                " + id);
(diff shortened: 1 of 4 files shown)


src/main/java/bio/terra/cli/serialization/userfacing/UFWorkspaceLight.java
```diff
@@ -52,12 +52,12 @@ protected UFWorkspaceLight() {
   /** Print out a workspace object in text format. */
   public void print() {
     PrintStream OUT = UserIO.getOut();
-    OUT.println("Terra workspace id: " + id);
-    OUT.println("Display name: " + name);
-    OUT.println("Description: " + description);
-    OUT.println("Google project: " + googleProjectId);
+    OUT.println("Terra workspace id:  " + id);
```
jaycarlton, line 55: Are these spaces? I thought we were using tabs but I guess this works.

jaycarlton (approved):

ginaygoogle:
> Can you have everything line up?
> 
> * The `0` after `# Resources:` should be under `h` in `https`
> * `verily-devel` should be under `h` in `https`

Hi Melissa. @melissachang  
Thanks for your comment. I also notice that when I do the investigation. There are 2 reasons why I left `Resources` and `verily-devel`. Please correct me if there is any misunderstanding. 
First is from the https://broadworkbench.atlassian.net/browse/PF-1631 Ticket screenshot for the format. The `gcs-bucket, object...` are not aligned with the above items. 
The other reason is the code for the first 5 items from the same `UFWorkspaceLight.java` file. the `#Resource` and `verly-devel` both from different file. 

Not sure whether we should keep consistent with the `resource update`. I am willing to do either way.  just want to clear there is any other meaning/reason/knowledge which I missed for resource keep like that way.

melissachang:
> > Can you have everything line up?
> > 
> > * The `0` after `# Resources:` should be under `h` in `https`
> > * `verily-devel` should be under `h` in `https`
> 
> Hi Melissa. @melissachang Thanks for your comment. I also notice that when I do the investigation. There are 2 reasons why I left `Resources` and `verily-devel`. Please correct me if there is any misunderstanding. First is from the https://broadworkbench.atlassian.net/browse/PF-1631 Ticket screenshot for the format. The `gcs-bucket, object...` are not aligned with the above items.

I see, I don't think that (misalignment) was intentional.

> The other reason is the code for the first 5 items from the same `UFWorkspaceLight.java` file. the `#Resource` and `verly-devel` both from different file.

`Terra server:` is from `Status.java`. That can be used for lots of commands, so let's not touch that.

`# Resources:` is from `UFWorkspace.java`. That should only be used with `terra workspace describe`. So let's change that. Please add comments in both places, to keep in sync with the other place.  @jaycarlton let us know if you disagree with anything. :)

@jaycarlton can you remind me why we have `UFWorkspaceLight.java` again?
> 
> Not sure whether we should keep consistent with the `resource update`. I am willing to do either way. just want to clear there is any other meaning/reason/knowledge which I missed for resource keep like that way.

jaycarlton:
> @jaycarlton let us know if you disagree with anything. :)
> @jaycarlton can you remind me why we have UFWorkspaceLight.java again?
`UFWorkspaceLight` is there to handle absent resources. It's backed by a `Workspace`, that used the `WorkspaceDescription` constructor, and no resources are available on that object. We had a bug where we erroneously showed zero resources in the JSON for workspaces coming in `workspace list`, and this fixed it so we don't show this field in that case. Now that we're refreshing the resources after `workspace list` (for the post-clone bug), we might be able to move away from that. Happy to talk offline.

I want to keep the num resources field around, as it's helpful in identifying which workspaces I was working on. I guess the UFID will help with this too. It's also really nice to have in the "are you sure you want to delete this workspace" prompt, which is why I put it in in the first place.

PR #255, opened 2022-05-17
PF-1569: Add Option to use ID Tokens for AuthN in Terra API Calls

jmczerk:
This PR adds the ability to optionally authenticate to Terra using JWT ID Tokens instead of opaque Access Tokens, configured at the `Server` level.

This PR also configures Verily `Server` definitions to enable ID Token authentication when calling Terra there.

### DIFF
--- a/src/main/java/bio/terra/cli/app/utils/AppDefaultCredentialUtils.java
+++ b/src/main/java/bio/terra/cli/app/utils/AppDefaultCredentialUtils.java
@@ -101,7 +101,7 @@ public class AppDefaultCredentialUtils {
    * ADC must be properly scoped; passing improperly scoped credentials will result in a {@code
    * SystemException}. Any other failure to obtain the token will result in an {@code IoException}.
    */
-  private static IdToken getIdTokenFromADC(GoogleCredentials applicationDefaultCredentials)
+  private static IdToken getIdTokenFromAdc(GoogleCredentials applicationDefaultCredentials)
       throws IOException {
     if (!(applicationDefaultCredentials instanceof IdTokenProvider)) {
       throw new SystemException(
@@ -123,12 +123,12 @@ public class AppDefaultCredentialUtils {
     return idTokenCredentials.getIdToken();
   }
 
-  public static TerraCredentials getExistingADC(List<String> scopes) throws IOException {
+  public static TerraCredentials getExistingAdc(List<String> scopes) throws IOException {
     GoogleCredentials applicationDefaultCredentials =
         AppDefaultCredentialUtils.getApplicationDefaultCredentials().createScoped(scopes);
 
     return new TerraCredentials(
         applicationDefaultCredentials,
-        AppDefaultCredentialUtils.getIdTokenFromADC(applicationDefaultCredentials));
+        AppDefaultCredentialUtils.getIdTokenFromAdc(applicationDefaultCredentials));
   }
 }
--- a/src/main/java/bio/terra/cli/app/utils/AppDefaultCredentialUtils.java
+++ b/src/main/java/bio/terra/cli/app/utils/AppDefaultCredentialUtils.java
@@ -1,14 +1,20 @@
 package bio.terra.cli.app.utils;
 
 import bio.terra.cli.businessobject.Context;
+import bio.terra.cli.exception.SystemException;
 import bio.terra.cli.exception.UserActionableException;
+import com.google.api.client.googleapis.auth.oauth2.GoogleClientSecrets;
 import com.google.auth.oauth2.ComputeEngineCredentials;
 import com.google.auth.oauth2.GoogleCredentials;
+import com.google.auth.oauth2.IdToken;
+import com.google.auth.oauth2.IdTokenCredentials;
+import com.google.auth.oauth2.IdTokenProvider;
 import com.google.auth.oauth2.ImpersonatedCredentials;
 import com.google.auth.oauth2.ServiceAccountCredentials;
 import com.google.auth.oauth2.UserCredentials;
 import java.io.IOException;
 import java.nio.file.Path;
+import java.util.List;
 import java.util.Optional;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -87,5 +93,33 @@ public class AppDefaultCredentialUtils {
           "Application default credentials are not defined. Run `gcloud auth application-default login`",
           ioEx);
     }
+  }
+
+  /**
+   * Get an ID token from an Application Default Credential and Client Secrets. Note that the passed
+   * ADC must be properly scoped, passing improperly scoped credentials will result in a {@code
+   * SystemException}. Any other failure to obtain the token will result in an {@code IoException}.
+   */
+  public static IdToken getIdTokenFromApplicationDefaultCredentials(
+      GoogleCredentials applicationDefaultCredentials, GoogleClientSecrets clientSecrets)
+      throws IOException {
+    if (!(applicationDefaultCredentials instanceof IdTokenProvider)) {
+      throw new SystemException(
+          "Passed credential is not an IdTokenProvider, please ensure only scoped ADC are passed.");
+    }
+
+    // To obtain an ID token from ADC, the ADC must be properly scoped and the OAuth2 Client ID must
+    // be passed as the target audience
+
+    IdTokenCredentials idTokenCredentials =
+        IdTokenCredentials.newBuilder()
+            .setIdTokenProvider((IdTokenProvider) applicationDefaultCredentials)
+            .setTargetAudience(clientSecrets.getDetails().getClientId())
+            .setOptions(List.of(IdTokenProvider.Option.FORMAT_FULL))
+            .build();
+
+    // Must call refresh() to obtain the token.
+    idTokenCredentials.refresh();
+    return idTokenCredentials.getIdToken();
   }
 }
--- a/src/main/java/bio/terra/cli/app/utils/AppDefaultCredentialUtils.java
+++ b/src/main/java/bio/terra/cli/app/utils/AppDefaultCredentialUtils.java
@@ -3,7 +3,7 @@ package bio.terra.cli.app.utils;
 import bio.terra.cli.businessobject.Context;
 import bio.terra.cli.exception.SystemException;
 import bio.terra.cli.exception.UserActionableException;
-import com.google.api.client.googleapis.auth.oauth2.GoogleClientSecrets;
+import bio.terra.cli.service.GoogleOauth;
 import com.google.auth.oauth2.ComputeEngineCredentials;
 import com.google.auth.oauth2.GoogleCredentials;
 import com.google.auth.oauth2.IdToken;
@@ -97,11 +97,10 @@ public class AppDefaultCredentialUtils {
 
   /**
    * Get an ID token from an Application Default Credential and Client Secrets. Note that the passed
-   * ADC must be properly scoped, passing improperly scoped credentials will result in a {@code
+   * ADC must be properly scoped; passing improperly scoped credentials will result in a {@code
    * SystemException}. Any other failure to obtain the token will result in an {@code IoException}.
    */
-  public static IdToken getIdTokenFromApplicationDefaultCredentials(
-      GoogleCredentials applicationDefaultCredentials, GoogleClientSecrets clientSecrets)
+  public static IdToken getIdTokenFromADC(GoogleCredentials applicationDefaultCredentials)
       throws IOException {
     if (!(applicationDefaultCredentials instanceof IdTokenProvider)) {
       throw new SystemException(
@@ -114,7 +113,7 @@ public class AppDefaultCredentialUtils {
     IdTokenCredentials idTokenCredentials =
         IdTokenCredentials.newBuilder()
             .setIdTokenProvider((IdTokenProvider) applicationDefaultCredentials)
-            .setTargetAudience(clientSecrets.getDetails().getClientId())
+            .setTargetAudience(GoogleOauth.getClientSecrets().getDetails().getClientId())
             .setOptions(List.of(IdTokenProvider.Option.FORMAT_FULL))
             .build();
 
--- a/src/main/java/bio/terra/cli/app/utils/AppDefaultCredentialUtils.java
+++ b/src/main/java/bio/terra/cli/app/utils/AppDefaultCredentialUtils.java
@@ -99,7 +99,7 @@ public class AppDefaultCredentialUtils {
   /**
    * Get an ID token from an Application Default Credential and Client Secrets. Note that the passed
    * ADC must be properly scoped; passing improperly scoped credentials will result in a {@code
-   * SystemException}. Any other failure to obtain the token will result in an {@code IoException}.
+   * SystemException}. Any other failure to obtain the token will result in an {@code IOException}.
    */
   private static IdToken getIdTokenFromAdc(GoogleCredentials applicationDefaultCredentials)
       throws IOException {
--- a/src/main/java/bio/terra/cli/app/utils/AppDefaultCredentialUtils.java
+++ b/src/main/java/bio/terra/cli/app/utils/AppDefaultCredentialUtils.java
@@ -4,6 +4,7 @@ import bio.terra.cli.businessobject.Context;
 import bio.terra.cli.exception.SystemException;
 import bio.terra.cli.exception.UserActionableException;
 import bio.terra.cli.service.GoogleOauth;
+import bio.terra.cli.service.utils.TerraCredentials;
 import com.google.auth.oauth2.ComputeEngineCredentials;
 import com.google.auth.oauth2.GoogleCredentials;
 import com.google.auth.oauth2.IdToken;
@@ -85,7 +86,7 @@ public class AppDefaultCredentialUtils {
   }
 
   /** Get the application default credentials. Throw an exception if they are not defined. */
-  public static GoogleCredentials getApplicationDefaultCredentials() {
+  private static GoogleCredentials getApplicationDefaultCredentials() {
     try {
       return GoogleCredentials.getApplicationDefault();
     } catch (IOException ioEx) {
@@ -100,7 +101,7 @@ public class AppDefaultCredentialUtils {
    * ADC must be properly scoped; passing improperly scoped credentials will result in a {@code
    * SystemException}. Any other failure to obtain the token will result in an {@code IoException}.
    */
-  public static IdToken getIdTokenFromADC(GoogleCredentials applicationDefaultCredentials)
+  private static IdToken getIdTokenFromADC(GoogleCredentials applicationDefaultCredentials)
       throws IOException {
     if (!(applicationDefaultCredentials instanceof IdTokenProvider)) {
       throw new SystemException(
@@ -120,5 +121,14 @@ public class AppDefaultCredentialUtils {
     // Must call refresh() to obtain the token.
     idTokenCredentials.refresh();
     return idTokenCredentials.getIdToken();
+  }
+
+  public static TerraCredentials getExistingADC(List<String> scopes) throws IOException {
+    GoogleCredentials applicationDefaultCredentials =
+        AppDefaultCredentialUtils.getApplicationDefaultCredentials().createScoped(scopes);
+
+    return new TerraCredentials(
+        applicationDefaultCredentials,
+        AppDefaultCredentialUtils.getIdTokenFromADC(applicationDefaultCredentials));
   }
 }
--- a/src/main/java/bio/terra/cli/businessobject/Server.java
+++ b/src/main/java/bio/terra/cli/businessobject/Server.java
@@ -42,6 +42,7 @@ public class Server {
   private final String wsmDefaultSpendProfile;
   private final String dataRepoUri;
   private final String externalCredsUri;
+  private final boolean idTokenAuthentication;
 
   private static final String DEFAULT_SERVER_FILENAME = "broad-dev-cli-testing.json";
   @VisibleForTesting public static final String RESOURCE_DIRECTORY = "servers";
@@ -57,6 +58,7 @@ public class Server {
     this.wsmDefaultSpendProfile = configFromDisk.wsmDefaultSpendProfile;
     this.dataRepoUri = configFromDisk.dataRepoUri;
     this.externalCredsUri = configFromDisk.externalCredsUri;
+    this.idTokenAuthentication = configFromDisk.idTokenAuthentication;
   }
 
   /** Return an instance of this class with default values. */
@@ -200,5 +202,9 @@ public class Server {
 
   public String getExternalCredsUri() {
     return externalCredsUri;
+  }
+
+  public boolean getIdTokenAuthentication() {
+    return idTokenAuthentication;
   }
 }
--- a/src/main/java/bio/terra/cli/businessobject/Server.java
+++ b/src/main/java/bio/terra/cli/businessobject/Server.java
@@ -42,7 +42,9 @@ public class Server {
   private final String wsmDefaultSpendProfile;
   private final String dataRepoUri;
   private final String externalCredsUri;
-  private final boolean idTokenAuthentication;
+  
+  // Terra services in the service instance are configured to accept JWT ID tokens for authentication.
+  private final boolean supportsIdToken;
 
   private static final String DEFAULT_SERVER_FILENAME = "broad-dev-cli-testing.json";
   @VisibleForTesting public static final String RESOURCE_DIRECTORY = "servers";
@@ -58,7 +60,7 @@ public class Server {
     this.wsmDefaultSpendProfile = configFromDisk.wsmDefaultSpendProfile;
     this.dataRepoUri = configFromDisk.dataRepoUri;
     this.externalCredsUri = configFromDisk.externalCredsUri;
-    this.idTokenAuthentication = configFromDisk.idTokenAuthentication;
+    this.supportsIdToken = configFromDisk.supportsIdToken;
   }
 
   /** Return an instance of this class with default values. */
@@ -204,7 +206,7 @@ public class Server {
     return externalCredsUri;
   }
 
-  public boolean getIdTokenAuthentication() {
-    return idTokenAuthentication;
+  public boolean getSupportsIdToken() {
+    return supportsIdToken;
   }
 }
--- a/src/main/java/bio/terra/cli/businessobject/Server.java
+++ b/src/main/java/bio/terra/cli/businessobject/Server.java
@@ -42,8 +42,9 @@ public class Server {
   private final String wsmDefaultSpendProfile;
   private final String dataRepoUri;
   private final String externalCredsUri;
-  
-  // Terra services in the service instance are configured to accept JWT ID tokens for authentication.
+
+  // Terra services in the service instance are configured to accept JWT ID tokens for
+  // authentication.
   private final boolean supportsIdToken;
 
   private static final String DEFAULT_SERVER_FILENAME = "broad-dev-cli-testing.json";
--- a/src/main/java/bio/terra/cli/businessobject/User.java
+++ b/src/main/java/bio/terra/cli/businessobject/User.java
@@ -6,14 +6,20 @@ import bio.terra.cli.exception.SystemException;
 import bio.terra.cli.serialization.persisted.PDUser;
 import bio.terra.cli.service.GoogleOauth;
 import bio.terra.cli.service.SamService;
+import bio.terra.cli.service.utils.TerraCredentials;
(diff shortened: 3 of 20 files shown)


jmczerk:
CC: @cbookg for visibility

src/main/java/bio/terra/cli/app/utils/AppDefaultCredentialUtils.java
```diff
@@ -88,4 +94,32 @@ public static GoogleCredentials getApplicationDefaultCredentials() {
...
     }
   }
+
+  /**
+   * Get an ID token from an Application Default Credential and Client Secrets. Note that the passed
+   * ADC must be properly scoped, passing improperly scoped credentials will result in a {@code
+   * SystemException}. Any other failure to obtain the token will result in an {@code IoException}.
+   */
+  public static IdToken getIdTokenFromApplicationDefaultCredentials(
+      GoogleCredentials applicationDefaultCredentials, GoogleClientSecrets clientSecrets)
```
jaycarlton, line 104: why not simply take in `IdTokenProvider` instead of using `instanceof`?

src/main/java/bio/terra/cli/businessobject/Server.java
```diff
@@ -57,6 +58,7 @@ public Server(PDServer configFromDisk) {
     this.wsmDefaultSpendProfile = configFromDisk.wsmDefaultSpendProfile;
     this.dataRepoUri = configFromDisk.dataRepoUri;
     this.externalCredsUri = configFromDisk.externalCredsUri;
+    this.idTokenAuthentication = configFromDisk.idTokenAuthentication;
```
jaycarlton, line 61: nit: to sound like a boolean, maybe name it `useIdTokenAuthentication`.

src/main/java/bio/terra/cli/businessobject/User.java
```diff
@@ -330,9 +364,14 @@ public boolean requiresReauthentication() {
     return accessToken.getExpirationTime().compareTo(cutOffDate) <= 0;
   }
 
-  /** Get the access token for the user credentials. */
+  /** Get the access token from the user's credentials. */
   public AccessToken getUserAccessToken() {
-    return GoogleOauth.getAccessToken(googleCredentials);
+    return GoogleOauth.getAccessToken(terraCredentials);
```
jaycarlton, line 369: what happens if `terraCredentials` is `null`?

src/main/java/bio/terra/cli/service/DataRepoService.java
```diff
@@ -38,7 +39,9 @@ private DataRepoService(@Nullable User user, Server server) {
     if (user != null) {
       // fetch the user access token
       // this method call will attempt to refresh the token if it's already expired
-      this.apiClient.setAccessToken(user.getUserAccessToken().getTokenValue());
+      AccessToken token =
+          server.getIdTokenAuthentication() ? user.getUserIdToken() : user.getUserAccessToken();
```
jaycarlton, line 43: You have to do this ternary bit twice; maybe make a method like `getEffectiveToken()`.

src/main/java/bio/terra/cli/service/GoogleOauth.java
```diff
@@ -142,98 +143,246 @@ public void browse(String url) {
...
    *
    * @param scopes list of scopes requested of the user
-   * @param clientSecretFile stream to the client secret file
+   * @param clientSecrets client secrets object
    * @param dataStoreDir directory where the local credential store is persisted
    * @return credentials object for the user
    */
-  public static UserCredentials getExistingUserCredential(
-      List<String> scopes, InputStream clientSecretFile, File dataStoreDir)
+  public static TerraCredentials getExistingUserCredential(
```
jaycarlton, line 180: Mark this `@Nullable` or, better yet, return `Optional<TerraCredentials>`.

jaycarlton: Document up here the situations in which the caller needs to refresh the token. I wonder if that couldn't all be done in one call, like `getOrRefreshUserCredentials()`.

src/main/java/bio/terra/cli/service/GoogleOauth.java
```diff
@@ -142,98 +143,246 @@ public void browse(String url) {
...
+    }
+  }
+
+  /**
+   * Helper class that creates and binds together several objects required for the OAuth token flow
+   * across different authentication modalities: the {@code GoogleAuthorizationCodeFlow}, the {@code
+   * IdCredentialListener} instance required to obtain an ID Token from the flow, and the Credential
+   * store used for both storing and retrieving credentials.
+   */
+  private static class TerraAuthenticationHelper {
```
jaycarlton, line 302: I'd probably just name this `OAuthHelper`.

src/main/java/bio/terra/cli/service/GoogleOauth.java
```diff
@@ -142,98 +143,246 @@ public void browse(String url) {
...
+      // create an IdCredentialListener pointing at the credential datastore in order to store and
+      // retrieve ID tokens
+      idCredentialListener =
+          new IdCredentialListener(
+              fileDataStoreFactory, StoredCredential.DEFAULT_DATA_STORE_ID, ID_TOKEN_STORE_KEY);
+
+      // create the code flow object, pointing at the credential datastore in order to store and
+      // retrieve ID tokens, and registering our IdCredentialListener for callbacks on token
+      // creation/refresh
+      googleAuthorizationCodeFlow =
```
jaycarlton, line 327: Is this for both browser and non-browser authentication?

src/main/java/bio/terra/cli/service/GoogleOauth.java
```diff
@@ -142,98 +143,246 @@ public void browse(String url) {
...
+      return new TerraAuthenticationHelper(scopes, clientSecrets, dataStoreDir);
+    }
+
+    public GoogleAuthorizationCodeFlow getGoogleAuthorizationCodeFlow() {
+      return googleAuthorizationCodeFlow;
+    }
+
+    private static <T extends Serializable> void deleteFromDataStore(
+        DataStore<T> dataStore, String key) throws IOException {
+      if (!dataStore.containsKey(key)) {}
```
jaycarlton, line 361: there's no body here

src/main/java/bio/terra/cli/service/GoogleOauth.java
```diff
@@ -142,98 +143,246 @@ public void browse(String url) {
...
+
+      dataStore.delete(key);
+    }
+
+    /** Delete stored ID token from the credential store */
+    public void deleteStoredIdToken() throws IOException {
```
jaycarlton, line 367: Would it make sense to have these objects implement `AutoCloseable` so we could delete them in a try-with-resources block?

src/main/java/bio/terra/cli/service/GoogleOauth.java
```diff
@@ -252,19 +401,39 @@ public static ServiceAccountCredentials getServiceAccountCredential(
...
+  /**
+   * @param credentials Terra credentials object
+   * @return id token
+   */
+  public static IdToken getIdToken(TerraCredentials credentials) {
+    IdToken idToken = credentials.getIdToken();
+    long expiration = idToken.getExpirationTime().getTime();
+    long now = System.currentTimeMillis();
+    if (now > expiration) {
+      // If we get here, there is no action we can take since we can't refresh an ID token, and no
```
jaycarlton, line 428: Is there anything about an expired token that's useful to the caller? If not, why do we return it?

src/main/java/bio/terra/cli/service/GoogleOauth.java
```diff
@@ -275,13 +444,15 @@ public static AccessToken getAccessToken(GoogleCredentials credential) {
    *
    * @param credential credentials object
    */
-  public static void revokeToken(Optional<GoogleCredentials> credential) {
-    if (credential.isPresent() && credential.get().getAccessToken() != null) {
+  public static void revokeToken(Optional<TerraCredentials> credential) {
+    if (credential.isPresent()
+        && credential.get().getGoogleCredentials().getAccessToken() != null) {
       String endpoint = "https://oauth2.googleapis.com/revoke";
```
jaycarlton, line 450: These can be private static final strings.

src/main/java/bio/terra/cli/service/SamService.java
```diff
@@ -66,7 +66,10 @@ public static SamService fromContext() {
 
   /** Factory method for class that talks to SAM. Pulls the current server from the context. */
   public static SamService forUser(User user) {
-    return new SamService(user.getUserAccessToken(), Context.getServer());
+    Server server = Context.getServer();
+    AccessToken token =
```
jaycarlton, line 70: This is an odd access pattern, since you're asking one object which method to call on another object. It's almost like the server knows too much.

src/main/java/bio/terra/cli/service/utils/TerraCredentials.java
```diff
@@ -0,0 +1,30 @@
+package bio.terra.cli.service.utils;
+
+import com.google.auth.oauth2.GoogleCredentials;
+import com.google.auth.oauth2.IdToken;
+
+/**
+ * Wrapper class to hold a user's credential to use when calling Terra API. Consists of a
```
jaycarlton, line 7: Do you mean the Workspace Manager API?

src/test/java/harness/TestUser.java
```diff
@@ -173,9 +179,12 @@ private GoogleCredentials getCredentials(List<String> scopes) throws IOException
               + jsonKey.toAbsolutePath()
               + ")");
     }
+
+    List<String> scopesWithCloudPlatform = new ArrayList<>(User.USER_SCOPES);
```
jaycarlton, line 183: You can use `List.of()` here I think.

src/test/java/unit/AuthLoginLogout.java
```diff
@@ -38,14 +39,26 @@ void loginTestUser() throws IOException {
     testUser.login();
 
     // check that the credential exists in the store on disk
-    DataStore<StoredCredential> dataStore = TestUser.getCredentialStore();
-    assertEquals(1, dataStore.keySet().size(), "credential store only contains one entry");
+    DataStore<StoredCredential> credentialStore = TestUser.getCredentialStore();
+    assertEquals(2, credentialStore.keySet().size(), "credential store contains two entries");
```
jaycarlton, line 43: you can use `assertThat(credentialStore.keySet(), hasSize(2), ...)`.

src/test/java/unit/AuthLoginLogout.java
```diff
@@ -38,14 +39,26 @@ void loginTestUser() throws IOException {
...
+        credentialStore.containsKey(GoogleOauth.ID_TOKEN_STORE_KEY),
+        "credential store contains  id token");
+    StoredCredential storedCredential = credentialStore.get(GoogleOauth.CREDENTIAL_STORE_KEY);
     assertThat(storedCredential.getAccessToken(), CoreMatchers.not(emptyOrNullString()));
 
+    DataStore<IdToken> idTokenStore = TestUser.getCredentialStore();
+    assertEquals(
+        idTokenStore.keySet(),
+        credentialStore.keySet(),
+        "credentialStore and idTokenStore are different views of same underlying datastore");
```
jaycarlton, line 57: Are you asserting an implementation detail here? I.e. could you have just as easily implemented it with two different data stores?

jaycarlton (approved):

src/main/java/bio/terra/cli/service/GoogleOauth.java
```diff
@@ -142,98 +143,246 @@ public void browse(String url) {
...
+    IdToken idToken = helper.getStoredIdToken();
+
+    if (idToken == null) {
+      // We have existing credentials, but do not have an ID token.  Return null in order to force a
+      // credential refresh which will capture and store an ID token with the new credentials.
+      logger.info("Credentials present, but no ID token present.");
+      return null;
+    }
+
+    // Since we can't refresh ID tokens later, check expiration and only use the token if within a
```
melissachang, line 217: Why can't ID tokens be refreshed later? Can you elaborate in your comment?

src/main/java/bio/terra/cli/service/SamService.java
```diff
@@ -66,7 +66,10 @@ public static SamService fromContext() {
...
+        server.getIdTokenAuthentication() ? user.getUserIdToken() : user.getUserAccessToken();
+    return new SamService(token, Context.getServer());
```
melissachang, line 72: Change `Context.getServer()` to `server`

src/test/java/harness/TestUser.java
```diff
@@ -84,7 +86,8 @@ public void login() throws IOException {
     // get domain-wide delegated credentials for this user. use the same scopes that are requested
     // of CLI users when they login.
     GoogleCredentials googleCredentials = getCredentials(User.USER_SCOPES);
-    writeTerraOAuthCredentialFile(googleCredentials);
+    IdToken idToken = getIdToken(googleCredentials);
```
melissachang, line 89: opt Delete line 89 and just have `idToken = getIdToken(googleCredentials)` inside `writeTerraOAuthCredentialFile`

src/test/java/harness/TestUser.java
```diff
@@ -173,9 +179,12 @@ private GoogleCredentials getCredentials(List<String> scopes) throws IOException
               + jsonKey.toAbsolutePath()
               + ")");
     }
+
+    List<String> scopesWithCloudPlatform = new ArrayList<>(User.USER_SCOPES);
+    scopesWithCloudPlatform.add(CLOUD_PLATFORM_SCOPE);
     GoogleCredentials serviceAccountCredential =
         ServiceAccountCredentials.fromStream(new FileInputStream(jsonKey.toFile()))
-            .createScoped(scopes);
+            .createScoped(scopesWithCloudPlatform);
```
melissachang, line 187: Shouldn't you use the `scopes` that is passed into this method?

src/test/java/harness/TestUser.java
```diff
@@ -184,12 +193,35 @@ private GoogleCredentials getCredentials(List<String> scopes) throws IOException
...
+  public static <T> DataStore getCredentialStore() throws IOException {
     Path globalContextDir = Context.getContextDir();
     FileDataStoreFactory dataStoreFactory = new FileDataStoreFactory(globalContextDir.toFile());
     return dataStoreFactory.getDataStore(StoredCredential.DEFAULT_DATA_STORE_ID);
   }
 
+  /** Get an ID token for this user. */
+  private IdToken getIdToken(GoogleCredentials credentials) throws IOException {
+
+    // Note that the passed GoogleCredential will be a domain-wide delegated credential obtained
```
melissachang, line 205: Thanks for the comments in this PR! :)

src/test/java/unit/AuthLoginLogout.java
```diff
@@ -38,14 +39,26 @@ void loginTestUser() throws IOException {
...
+    assertTrue(
+        credentialStore.containsKey(GoogleOauth.CREDENTIAL_STORE_KEY),
+        "credential store contains hard-coded user key");
     assertTrue(
-        dataStore.containsKey(GoogleOauth.CREDENTIAL_STORE_KEY),
-        "credential store contains hard-coded single user key");
-    StoredCredential storedCredential = dataStore.get(GoogleOauth.CREDENTIAL_STORE_KEY);
+        credentialStore.containsKey(GoogleOauth.ID_TOKEN_STORE_KEY),
+        "credential store contains  id token");
+    StoredCredential storedCredential = credentialStore.get(GoogleOauth.CREDENTIAL_STORE_KEY);
```
melissachang, line 50: Can there be a line similar to 50-51, but for ID_TOKEN_STORE_KEY?

melissachang (approved):

src/main/java/bio/terra/cli/app/utils/AppDefaultCredentialUtils.java
```diff
@@ -88,4 +94,32 @@ public static GoogleCredentials getApplicationDefaultCredentials() {
           ioEx);
     }
   }
+
+  /**
+   * Get an ID token from an Application Default Credential and Client Secrets. Note that the passed
+   * ADC must be properly scoped, passing improperly scoped credentials will result in a {@code
+   * SystemException}. Any other failure to obtain the token will result in an {@code IoException}.
+   */
+  public static IdToken getIdTokenFromApplicationDefaultCredentials(
```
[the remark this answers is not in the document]
jaycarlton: `getIdTokenFromAdc()` is more in line with our patterns I think

src/main/java/bio/terra/cli/businessobject/User.java
```diff
@@ -108,11 +114,10 @@ public void loadExistingCredentials() {
       loadAppDefaultCredentials();
       return;
     }
-    try (InputStream inputStream =
-        User.class.getClassLoader().getResourceAsStream(CLIENT_SECRET_FILENAME)) {
-      googleCredentials =
+    try {
+      terraCredentials =
           GoogleOauth.getExistingUserCredential(
```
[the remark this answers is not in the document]
jaycarlton: I was thinking about ways to make this more OOPy, and maybe an object representing the user's auth "stuff" would make sense. It could change whenever we switch servers, but be outlived by the `User` object. It could also implement an interface that abstracts over whether it's using a JWT or an access token.

src/main/java/bio/terra/cli/businessobject/User.java
```diff
@@ -330,9 +364,14 @@ public boolean requiresReauthentication() {
     return accessToken.getExpirationTime().compareTo(cutOffDate) <= 0;
```
jaycarlton, line 364: It's a little more readable to use `Instant.now().isBefore(cutoffInstant)`.

src/main/java/bio/terra/cli/service/utils/TerraCredentials.java
jmczerk: No, any Terra API (WSM API, Sam API, ECM API...).  Is there a better way to word that?

jaycarlton: maybe "any Terra service API"

src/main/java/bio/terra/cli/app/utils/AppDefaultCredentialUtils.java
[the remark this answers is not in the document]
jmczerk: I moved the concept/object `clientSecrets` into a static final in `GoogleOauth` that gets initialized at static init time to guarantee that the JSON only ever gets read once.  This does therefore also guarantee that it gets read once even if `GoogleOauth.getClientSecrets()` is never called...

I was initially hoping to make it lazy-initialized on call to `getClientSecrets()`, but we call it from static functions and I didn't want to worry about race conditions on writing a static variable at runtime (even though I think this would only happen in a single-threaded context...).

I suspect that (1) `getClientSecrets()` is always called and (2) any further optimization is probably not worthwhile given this is reading out of the Jar and fairly lightweight.

src/main/java/bio/terra/cli/businessobject/Server.java
jmczerk: Per some other comments, given this is a property on the server, perhaps it makes sense for it to be `acceptsIdToken` or `supportsIdToken` (dropping the `...Authentication` for brevity)?

src/main/java/bio/terra/cli/businessobject/User.java
```diff
@@ -163,19 +168,32 @@ public static void login(LogInMode logInMode) {
   }
 
   /**
-   * Fetches a Google Application Credentials with {@code PET_SA_SCOPES} and set to {@code
-   * googleCredentials}.
+   * Fetches a Google Application Credentials with {@code PET_SA_SCOPES} and stores in {@code
+   * terraCredentials}.
    */
   private void loadAppDefaultCredentials() {
```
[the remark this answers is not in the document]
jmczerk: This sets member variable `terraCredentials`.

src/main/java/bio/terra/cli/businessobject/User.java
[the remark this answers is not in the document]
jmczerk: Moved the client secrets into GoogleOauth.  I think there's more refactoring we could do per Jay's comment, but I think I'd like to leave that for a separate effort.

src/test/java/unit/AuthLoginLogout.java
jmczerk: I'm not completely sure what the intent of this test is, but I tried to match it.  I did very intentionally put the ID token into the same "DataStore" file as the GoogleCredential, under different keys, rather than having them exist in separate files.  

Side note, the semantics of `DataStore<T>` are very odd to me, in that it uses the parameterized type to give a view of an entry in a file data store that may contain different keyed entries of different types.  Seems like it should be something like `DataStoreEntry<T>` 🤷

src/main/java/bio/terra/cli/service/DataRepoService.java
melissachang: I still like `user.getToken` (or `user.getEffectiveToken`). The fact that `user.getToken` uses Server is an implementation detail. (One that will hopefully go away in the future.) Conceptually, you're getting a token for the `user` that was passed in.

melissachang (approved):
Btw, github review UI is very unintuitive. I got a separate email for each of your comments. There's another way of sending review comments, where all comments are in one email. One advantage is that I'll know when you're done responding to comments. Click `Start a Review as described in [this video](https://www.youtube.com/watch?v=lSnbOtw4izI).

src/main/java/bio/terra/cli/businessobject/User.java
jaycarlton: Not necessarily. It's fine.

src/test/java/unit/AuthLoginLogout.java
jaycarlton: Yeah it was the first time I'd seen that pattern.

jmczerk:
I think I have addressed everything.  Please re-review when you have a chance.

src/main/java/bio/terra/cli/businessobject/Server.java
```diff
@@ -43,6 +43,10 @@ public class Server {
   private final String dataRepoUri;
   private final String externalCredsUri;
 
+  // Terra services in the service instance are configured to accept JWT ID tokens for
+  // authentication.
```
melissachang, line 47: // This is a flag because Broad services don't support ID tokens yet.

melissachang (approved):

src/main/java/bio/terra/cli/service/utils/TerraCredentials.java
```diff
@@ -0,0 +1,30 @@
+package bio.terra.cli.service.utils;
+
+import com.google.auth.oauth2.GoogleCredentials;
+import com.google.auth.oauth2.IdToken;
+
+/**
+ * Wrapper class to hold a user's credential to use when making Terra service API calls. Consists of
+ * a GoogleCredential object containing access and refresh tokens and an IdToken object containing a
+ * JWT ID token.
```
jaycarlton, line 9: If we have to maintain three different tokens now, how did our lives get any easier in the switch to JWT?

jaycarlton (approved):
I tried to read up on Oauth, OpenID, and JWT the other day. I'm still pretty ignorant of what's going on here.

src/main/java/bio/terra/cli/app/utils/AppDefaultCredentialUtils.java
```diff
@@ -88,4 +95,40 @@ public static GoogleCredentials getApplicationDefaultCredentials() {
           ioEx);
     }
   }
+
+  /**
+   * Get an ID token from an Application Default Credential and Client Secrets. Note that the passed
+   * ADC must be properly scoped; passing improperly scoped credentials will result in a {@code
+   * SystemException}. Any other failure to obtain the token will result in an {@code IoException}.
```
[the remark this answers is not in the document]
jmczerk: 👍

src/main/java/bio/terra/cli/businessobject/User.java
```diff
@@ -163,19 +158,24 @@ public static void login(LogInMode logInMode) {
   }
 
   /**
-   * Fetches a Google Application Credentials with {@code PET_SA_SCOPES} and set to {@code
-   * googleCredentials}.
+   * Fetches a Google Application Credentials with {@code PET_SA_SCOPES} and stores in {@code
```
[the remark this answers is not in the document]
jmczerk: Yeah, from the comment on `PET_SA_SCOPES` declaration, it adds cloud platform to the other scopes required to by Terra services ("openid", "email", and "profile"), as these are needed to make GCP calls.  I guess it assumes that if we're using ADC under normal operations we're acting as a Pet and thus want to be able to make GCP calls?

src/main/java/bio/terra/cli/service/GoogleOauth.java
```diff
@@ -39,44 +47,68 @@
...
+  private static final GoogleClientSecrets clientSecrets = readClientSecrets();
 
   private GoogleOauth() {}
 
   private static final JsonFactory JSON_FACTORY = JacksonFactory.getDefaultInstance();
 
+  /** Load the client secrets file to pass to oauth API's. */
+  private static GoogleClientSecrets readClientSecrets() {
+    try (InputStream inputStream =
+        User.class.getClassLoader().getResourceAsStream(CLIENT_SECRET_FILENAME)) {
```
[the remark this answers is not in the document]
jmczerk: Bad copy paste.... thanks.

src/main/java/bio/terra/cli/service/GoogleOauth.java
```diff
@@ -39,44 +47,68 @@
...
+
+      GoogleClientSecrets clientSecrets =
+          GoogleClientSecrets.load(
+              JacksonFactory.getDefaultInstance(),
+              new InputStreamReader(inputStream, StandardCharsets.UTF_8));
+
+      return clientSecrets;
+    } catch (IOException ioException) {
+      throw new SystemException(
+          String.format("Could not open client secret file '%s'.", CLIENT_SECRET_FILENAME),
```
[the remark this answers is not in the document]
jmczerk: Clarified error accordingly.

src/main/java/bio/terra/cli/service/GoogleOauth.java
```diff
@@ -146,94 +177,216 @@ public void browse(String url) {
...
    * Get the existing credential for the given user.
    *
    * @param scopes list of scopes requested of the user
-   * @param clientSecretFile stream to the client secret file
    * @param dataStoreDir directory where the local credential store is persisted
    * @return credentials object for the user
    */
-  public static UserCredentials getExistingUserCredential(
-      List<String> scopes, InputStream clientSecretFile, File dataStoreDir)
+  public static TerraCredentials getExistingUserCredential(List<String> scopes, File dataStoreDir)
```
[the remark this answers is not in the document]
jmczerk: 👍

src/main/java/bio/terra/cli/service/GoogleOauth.java
```diff
@@ -252,19 +405,40 @@ public static ServiceAccountCredentials getServiceAccountCredential(
...
    * @return access token
    */
-  public static AccessToken getAccessToken(GoogleCredentials credential) {
+  public static AccessToken getAccessToken(TerraCredentials credential) {
     try {
-      credential.refreshIfExpired();
+      credential.getGoogleCredentials().refreshIfExpired();
     } catch (IOException ioEx) {
       logger.warn("Error refreshing access token", ioEx);
       // don't throw an exception here because the token may not be expired, in which case it's fine
```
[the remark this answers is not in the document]
jmczerk: I don't fully get the intention either; speculatively, `refreshIfExpired()` is supposed to refresh "if expired or expiring imminently", so I guess maybe the plan is to try the call anyway in case it was "imminent" and refresh call hit a spurious error.  Then maybe the downstream call would succeed due to the token not yet being expired (and/or clock skew)?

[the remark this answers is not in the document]
jaycarlton: Seems like a dubious optimization, but it's plausible. Thanks.

src/test/java/unit/Workspace.java
```diff
@@ -132,7 +132,7 @@ void statusDescribeListReflectUpdate() throws IOException {
 
     // `terra workspace update --format=json --new-id=$newId --new-name=$newName
     // --new-description=$newDescription`
```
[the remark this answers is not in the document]
ginaygoogle: Yes. Correct. John. Thank you for the help. 

I am wondering we have the `workspace delete` lastly, where made the old name workspace exist here. just want to learn where the bug comes from.

PR #259, opened 2022-05-23
PF-1621 ufid in workspace update and clone

ginaygoogle:
workspace update and clone lake of ufid
image
image

workspace update and clone successfully with ufid
Screen Shot 2022-05-23 at 5 44 00 PM
Screen Shot 2022-05-23 at 5 45 43 PM

### DIFF
--- a/src/main/java/bio/terra/cli/businessobject/Workspace.java
+++ b/src/main/java/bio/terra/cli/businessobject/Workspace.java
@@ -145,16 +145,17 @@ public class Workspace {
   /**
    * Update the mutable properties of the current workspace.
    *
-   * @param ufid optional display id
+   * @param userFacingId optional user-facing ID
    * @param name optional display name
    * @param description optional description
    * @throws UserActionableException if there is no current workspace
    */
   public Workspace update(
-      @Nullable String ufid, @Nullable String name, @Nullable String description) {
+      @Nullable String userFacingId, @Nullable String name, @Nullable String description) {
     // call WSM to update the existing workspace object
     WorkspaceDescription updatedWorkspace =
-        WorkspaceManagerService.fromContext().updateWorkspace(uuid, ufid, name, description);
+        WorkspaceManagerService.fromContext()
+            .updateWorkspace(uuid, userFacingId, name, description);
     logger.info("Updated workspace: {}", updatedWorkspace);
 
     // convert the WSM object to a CLI object
@@ -252,15 +253,15 @@ public class Workspace {
   /**
    * Clone the current workspace into a new one
    *
-   * @param ufid - id of the new workspace
+   * @param userFacingId - user-facing ID of the new workspace
    * @param name - name of the new workspace
    * @param description - description of the new workspace
    * @return - ClonedWorkspace structure with details on each resource
    */
   public ClonedWorkspace clone(
-      @Nullable String ufid, @Nullable String name, @Nullable String description) {
+      @Nullable String userFacingId, @Nullable String name, @Nullable String description) {
     CloneWorkspaceResult result =
-        WorkspaceManagerService.fromContext().cloneWorkspace(uuid, ufid, name, description);
+        WorkspaceManagerService.fromContext().cloneWorkspace(uuid, userFacingId, name, description);
     return result.getWorkspace();
   }
 
--- a/src/main/java/bio/terra/cli/businessobject/Workspace.java
+++ b/src/main/java/bio/terra/cli/businessobject/Workspace.java
@@ -145,14 +145,16 @@ public class Workspace {
   /**
    * Update the mutable properties of the current workspace.
    *
+   * @param id optional display id
    * @param name optional display name
    * @param description optional description
    * @throws UserActionableException if there is no current workspace
    */
-  public Workspace update(@Nullable String name, @Nullable String description) {
+  public Workspace update(
+      @Nullable String id, @Nullable String name, @Nullable String description) {
     // call WSM to update the existing workspace object
     WorkspaceDescription updatedWorkspace =
-        WorkspaceManagerService.fromContext().updateWorkspace(uuid, name, description);
+        WorkspaceManagerService.fromContext().updateWorkspace(uuid, id, name, description);
     logger.info("Updated workspace: {}", updatedWorkspace);
 
     // convert the WSM object to a CLI object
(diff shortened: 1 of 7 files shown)


src/main/java/bio/terra/cli/businessobject/Workspace.java
```diff
@@ -145,14 +145,16 @@ private static Workspace convertWorkspaceDescriptionToWorkspace(
...
    * Update the mutable properties of the current workspace.
    *
+   * @param id optional display id
    * @param name optional display name
    * @param description optional description
    * @throws UserActionableException if there is no current workspace
    */
-  public Workspace update(@Nullable String name, @Nullable String description) {
+  public Workspace update(
+      @Nullable String id, @Nullable String name, @Nullable String description) {
```
jaycarlton, line 154: Somewhere we should check that one or more of these is non-null.

src/main/java/bio/terra/cli/businessobject/Workspace.java
```diff
@@ -250,13 +252,15 @@ public List<Resource> listResourcesAndSync() {
...
    * Clone the current workspace into a new one
    *
+   * @param id - id of the new workspace
    * @param name - name of the new workspace
    * @param description - description of the new workspace
    * @return - ClonedWorkspace structure with details on each resource
    */
-  public ClonedWorkspace clone(@Nullable String name, @Nullable String description) {
+  public ClonedWorkspace clone(
+      @Nullable String id, @Nullable String name, @Nullable String description) {
```
jaycarlton, line 261: I'd call this `ufid` for extra clarity.

src/main/java/bio/terra/cli/command/workspace/Clone.java
```diff
@@ -21,6 +21,9 @@
 /** This corresponds to the third-level "terra workspace clone" command. */
 @Command(name = "clone", description = "Clone an existing workspace.")
 public class Clone extends BaseCommand {
+  @CommandLine.Option(names = "--id", required = true, description = "Workspace ID")
```
jaycarlton, line 24: Indicate if this is for the source or destination workspace.

src/main/java/bio/terra/cli/command/workspace/Update.java
```diff
@@ -20,6 +20,9 @@ public class Update extends BaseCommand {
   Update.UpdateArgGroup argGroup;
 
   static class UpdateArgGroup {
+    @CommandLine.Option(names = "--new-id", required = false, description = "Workspace ID.")
+    private String displayId;
```
jaycarlton, line 24: Call this userFacingId for consistency.

jaycarlton (approved):

src/main/java/bio/terra/cli/businessobject/Workspace.java
ginaygoogle: done

src/main/java/bio/terra/cli/command/workspace/Update.java
ginaygoogle: done

src/main/java/bio/terra/cli/command/workspace/Clone.java
ginaygoogle: done

src/main/java/bio/terra/cli/businessobject/Workspace.java
ginaygoogle: Do you think i can write an assertion here(L155) and L262?

This is not a test, I think if I write an assertnotnull will be wired? Or just a if statement is fine?

jaycarlton: I'd do an `if` statement and throw a `UserActionableException`. There should be examples where we do this.

melissachang (commented):
Can you please rerun the cli commands after making these changes, and paste the output? You might find it easier to just copy the output and paste between ```, rather than paste an image.

src/main/java/bio/terra/cli/command/workspace/Clone.java
```diff
@@ -21,7 +21,7 @@
 /** This corresponds to the third-level "terra workspace clone" command. */
 @Command(name = "clone", description = "Clone an existing workspace.")
 public class Clone extends BaseCommand {
-  @CommandLine.Option(names = "--id", required = true, description = "Source workspace ID")
+  @CommandLine.Option(names = "--new-id", required = true, description = "ID for new workspace")
```
jaycarlton, line 24: Double check this is what happens. I.e. that it applies to the destination and not the source. (We have commands that take in a `--workspace` argument to set the active workspace, and I'm worried this might have been one of those and we no longer have the ability to do that).

src/main/java/bio/terra/cli/businessobject/Workspace.java
```diff
@@ -250,13 +253,15 @@ public List<Resource> listResourcesAndSync() {
...
    * Clone the current workspace into a new one
    *
+   * @param userFacingId - user-facing ID of the new workspace
    * @param name - name of the new workspace
    * @param description - description of the new workspace
    * @return - ClonedWorkspace structure with details on each resource
    */
-  public ClonedWorkspace clone(@Nullable String name, @Nullable String description) {
+  public ClonedWorkspace clone(
+      @Nullable String userFacingId, @Nullable String name, @Nullable String description) {
```
melissachang, line 262: Please change `@Nullable String userFacingId` to `String userFacingId`. When you run `terra workspace clone`, `--new-id` is required. `@Nullable` is only for optional fields. The variable will be null if user didn't pass in that option.

src/main/java/bio/terra/cli/command/shared/options/WorkspaceOverride.java
```diff
@@ -17,12 +17,12 @@ public class WorkspaceOverride {
   @CommandLine.Option(
       names = "--workspace",
       description = "Workspace id to use for this command only.")
-  private String userFacingId;
+  private String id;
```
melissachang, line 20: Let's add comment since this is easy to forget.

```
// Variable is `id` instead of `userFacingId` because user sees it when they run a command with missing required arguments
```

src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
```diff
@@ -381,16 +387,18 @@ public void enablePet(UUID workspaceId) {
    * workspace.
    *
    * @param workspaceId - workspace ID to clone
-   * @param displayName - optional name of new cloned workspace
+   * @param userFacingId - optional userFacingId of new cloned workspace
```
melissachang, line 390: Change `optional` to `required`.

melissachang (approved):
Thanks

src/main/java/bio/terra/cli/command/workspace/Clone.java
ginaygoogle: Yes. after I tested. the `--new-id` represent the new id. and the `--workspace` represent the old(current active) workspace id. See the link for testing.
https://paste.googleplex.com/4896313884803072

src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
ginaygoogle: done

ginaygoogle:
https://paste.googleplex.com/4896313884803072 
final version for testing the CLI without flag.

PR #263, opened 2022-06-01
hot fix on the reg expression bug

ginaygoogle:


src/test/java/unit/Group.java
```diff
@@ -315,6 +294,31 @@ private static void expectListedMemberWithPolicies(
...
+    String[] rowHead = rows[0].split("\\s+");
+    assertEquals("EMAIL", rowHead[0].trim().replace("\r", ""));
+    assertEquals("POLICIES", rowHead[1].trim().replace("\r", ""));
+
+    // test the email exsit in the table and the policy correct
+    Boolean emailExist = false;
+    for (int i = 1; i < rows.length; i = i + 1) {
+      String[] rowi = rows[i].split("\\s+", 2);
+      if (email.equalsIgnoreCase(rowi[0])) {
+        assertEquals("[MEMBER, ADMIN]", rowi[1].trim().replace("\r", ""));
```
melissachang, line 315: Instead of hard-coding `"[MEMBER, ADMIN]"`, you should use the `policies` that's passed into this method.

Instead of split, I recommend just doing this, to make test easier to read.

```
    for (int i = 1; i < rows.length; i = i + 1) {
      if (row[i].contains(email)) {
        emailExists = true;
        for (var policy : policies) {
          assert row[i] contains policy
        } 
    }
```

src/test/java/unit/Group.java
```diff
@@ -315,6 +294,31 @@ private static void expectListedMemberWithPolicies(
...
+    assertTrue(cmd.stdErr == null || cmd.stdErr.isEmpty());
+    assertEquals(0, cmd.exitCode, "group list-user returned successfully");
+    String[] rows = cmd.stdOut.split("\\n");
+    String[] rowHead = rows[0].split("\\s+");
+    assertEquals("EMAIL", rowHead[0].trim().replace("\r", ""));
+    assertEquals("POLICIES", rowHead[1].trim().replace("\r", ""));
+
+    // test the email exsit in the table and the policy correct
+    Boolean emailExist = false;
+    for (int i = 1; i < rows.length; i = i + 1) {
```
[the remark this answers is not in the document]
melissachang: This isn't done yet

src/test/java/unit/Group.java
```diff
@@ -315,6 +294,32 @@ private static void expectListedMemberWithPolicies(
...
+    // assert header is correct
+    assertTrue(cmd.stdErr == null || cmd.stdErr.isEmpty());
+    assertEquals(0, cmd.exitCode, "group list-user returned successfully");
+    String[] rows = cmd.stdOut.split("\\n");
+    String[] rowHead = rows[0].split("\\s+");
+    assertEquals("EMAIL", rowHead[0].trim().replace("\r", ""));
+    assertEquals("POLICIES", rowHead[1].trim().replace("\r", ""));
+
+    // test the email exist in the table and the policy correct
+    boolean emailExist = false;
```
melissachang, line 311: emailExists

src/test/java/unit/Group.java
```diff
@@ -315,6 +294,32 @@ private static void expectListedMemberWithPolicies(
...
+
+    // assert header is correct
+    assertTrue(cmd.stdErr == null || cmd.stdErr.isEmpty());
+    assertEquals(0, cmd.exitCode, "group list-user returned successfully");
+    String[] rows = cmd.stdOut.split("\\n");
+    String[] rowHead = rows[0].split("\\s+");
+    assertEquals("EMAIL", rowHead[0].trim().replace("\r", ""));
+    assertEquals("POLICIES", rowHead[1].trim().replace("\r", ""));
+
+    // test the email exist in the table and the policy correct
```
melissachang, line 310: // assert email and policies are correct

src/test/java/unit/Group.java
```diff
@@ -315,6 +294,32 @@ private static void expectListedMemberWithPolicies(
...
+    boolean emailExist = false;
+    for (int i = 1; i < rows.length; i = i + 1) {
+      if (rows[i].contains(email.toLowerCase())) {
```
melissachang, line 313: Delete .toLowerCase()

melissachang (approved):
Code looks good, but let's wait until terra-cli tests are back to normal (passing) before merging

src/test/java/unit/Group.java
```diff
@@ -315,6 +294,31 @@ private static void expectListedMemberWithPolicies(
...
+    String[] rows = cmd.stdOut.split("\\n");
+    String[] rowHead = rows[0].split("\\s+");
+    assertEquals("EMAIL", rowHead[0].trim().replace("\r", ""));
+    assertEquals("POLICIES", rowHead[1].trim().replace("\r", ""));
+
+    // test the email exsit in the table and the policy correct
+    Boolean emailExist = false;
+    for (int i = 1; i < rows.length; i = i + 1) {
+      String[] rowi = rows[i].split("\\s+", 2);
+      if (email.equalsIgnoreCase(rowi[0])) {
```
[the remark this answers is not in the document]
melissachang: Sorry I missed this earlier. I would take out `toLowerCase` and change these emails to lower-case:

https://github.com/DataBiosphere/terra-cli/blob/main/src/test/resources/testconfigs/broad.json

src/test/java/unit/Group.java
```diff
@@ -315,6 +294,32 @@ private static void expectListedMemberWithPolicies(
...
+    String[] rowHead = rows[0].split("\\s+");
+    assertEquals("EMAIL", rowHead[0].trim().replace("\r", ""));
+    assertEquals("POLICIES", rowHead[1].trim().replace("\r", ""));
+
+    // assert email and policies are correct
+    boolean emailExists = false;
+    for (int i = 1; i < rows.length; i++) {
+      if (rows[i].contains(email)) {
+        emailExists = true;
+        for (var policy : policies) {
```
jaycarlton, line 315: Looks like you are asserting that every row must contain every policy. I don't think it works that way in the general case.

src/test/java/unit/Group.java
```diff
@@ -315,6 +294,32 @@ private static void expectListedMemberWithPolicies(
...
+          assertTrue(rows[i].contains(policy.toString()));
```
jaycarlton, line 316: Whenever you use `assertTrue()` or `assertFalse()`, please put a description string saying what you're actually asserting. For example, 
```
assertTrue(rows[i].contains(policy.toString()), "each row contains all policies");
```

Note that in that case you'd probably want to use 

```
assertThat("Row has all policies.", rows[i], hasItems(policy.toString()));
```

src/test/java/unit/Group.java
```diff
@@ -315,6 +294,32 @@ private static void expectListedMemberWithPolicies(
         "test user has the right policies");
   }
 
+  private void expectListedMemberWithPoliciesTableFormat(
+      String name, String email, GroupPolicy... policies) throws JsonProcessingException {
+    // call `terra group list-users` in table format
+    TestCommand.Result cmd = TestCommand.runCommand("group", "list-users", "--name=" + name);
```
jaycarlton, line 300: If you want, you could change `TestCommand.runCommandExpectSuccess()` to return a `TestCommand.Result` instead of being `void` and then you wouldn't have to assert the stdErr and the exit code.

src/test/resources/testconfigs/broad.json
```diff
@@ -1,27 +1,27 @@
 {
   "testUsers": [
     {
-      "email": "Penelope.TwilightsHammer@test.firecloud.org",
+      "email": "penelope.twilightsHammer@test.firecloud.org",
```
jaycarlton, line 4: Why are you making these changes? You can just compare ignoring case. I think we should keep these the way they are specified in the Broad's spreadsheet. @zloery do you remember where that is?

jaycarlton (changes requested):

src/test/resources/testconfigs/broad.json
melissachang: I requested these changes in an earlier comment on this PR.

Spreadsheet: https://docs.google.com/spreadsheets/d/17PjpwWedN-hxeOlEbepNahLmRDqhvLc25yYxbz6Oj3E/edit#gid=0

I thought it would be cleaner to be lower-case here (one place), versus having to convert to lower case in various tests. Also, people don't normally upper-case emails; ie when they use terra cli, email will be lower-case. I don't think we need to copy spreadsheet exactly. But I don't feel too strongly.

If you change this back to upper-case: Please add a comment in the test:

// Convert to lower-case because email in broad.json is mixed case

If you leave this as lower-case: You forgot to lower-case the "H" in "Hammer".

jaycarlton: I'm _leaning_ toward not doing this only because we've had bugs where we needed to do lowercase comparisons on emails. You're right that it's very uncommon, but if it's a valid email, we should support it. Changing this file will affect all tests, and it will err on the side of making things pass that should have failed. I'll let Guang be the tie-breaking vote.

I'm not sure where in the stack we lose the case information. It might be Sam, or a GCP API.

src/test/java/unit/Group.java
ginaygoogle: I think we could not do it as void. because the header verification need the cmd.stdOut.

src/test/java/unit/Group.java
```diff
@@ -315,6 +294,32 @@ private static void expectListedMemberWithPolicies(
...
+
+    // assert header is correct
+    assertTrue(cmd.stdErr == null || cmd.stdErr.isEmpty());
+    assertEquals(0, cmd.exitCode, "group list-user returned successfully");
+    String[] rows = cmd.stdOut.split("\\n");
+    String[] rowHead = rows[0].split("\\s+");
+    assertEquals("EMAIL", rowHead[0].trim().replace("\r", ""));
```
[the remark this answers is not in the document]
ginaygoogle: done

src/test/java/unit/Group.java
ginaygoogle: Hi Jay. thanks for the leading. I will pay attention to the assertion next time. 
due to it could not use hasItems. So I use `assertThat("Row has all policies.", rows[i].contains(policy.toString()));`

src/test/java/unit/Group.java
ginaygoogle: done

src/test/java/unit/Group.java
ginaygoogle: I think the policy there is a parameter pass from the function. in this test, we test that one has 2 policies, so we pass 2 polices here and use for loop to iterate it.

src/test/java/unit/Group.java
```diff
@@ -315,6 +294,32 @@ private static void expectListedMemberWithPolicies(
...
+    for (int i = 1; i < rows.length; i++) {
+      // Convert to lower-case because email in broad.json is mixed case
+      if (rows[i].contains(email.toLowerCase())) {
+        emailExists = true;
+        for (var policy : policies) {
+          assertThat("Row has all policies.", rows[i].contains(policy.toString()));
+        }
+      }
+    }
+    assertTrue(emailExists);
```
jaycarlton, line 319: Put a message like "Email address for group member should be included in table output." in this assertion.

jaycarlton (approved):

src/test/java/unit/Group.java
```diff
@@ -317,6 +317,7 @@ private TestCommand.Result expectListedMemberWithPoliciesTableFormat(
       }
     }
     assertTrue(emailExists);
+    // email address for group member should be included in table output.
```
jaycarlton, line 320: I mean inside the `assertTrue` call put that message.

PR #266, opened 2022-06-10
PF-1722: resource description update add --new-

ginaygoogle:


src/main/java/bio/terra/cli/command/shared/options/ResourceUpdate.java
```diff
@@ -18,19 +18,18 @@ public class ResourceUpdate {
           "New name of the resource. Only alphanumeric and underscore characters are permitted.")
   public String newName;
 
-  @CommandLine.Mixin public ResourceDescription resourceDescriptionOption;
+  @CommandLine.Option(names = "--new-description", description = "New description of the resource.")
```
yuhuyoyo, line 21: why do you change this?

yuhuyoyo (approved):

src/main/java/bio/terra/cli/command/shared/options/ResourceUpdate.java
ginaygoogle: because there is a ticket before for making all updated parameters with "--new-", This one is which I missed last time...

jaycarlton (approved):

PR #267, opened 2022-06-13
[PF-1721] Fix Metadata command option

jaycarlton:
Metadata option declaration was using a single string with comma-separated values. We need to use an array of separate strings for this to work.

Added a small test to make sure command is accepted; we can only see the metadata created in the Cloud Console AFAIK.

Also a small .gitignore entry for jenv local file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,6 @@ out/
 ### Nextflow
 .nextflow
 .nextflow.log
+
+### Jenv directory-local version setting
+.java-version
--- a/src/main/java/bio/terra/cli/command/resource/create/GcpNotebook.java
+++ b/src/main/java/bio/terra/cli/command/resource/create/GcpNotebook.java
@@ -64,10 +64,10 @@ public class GcpNotebook extends BaseCommand {
   private String postStartupScript;
 
   @CommandLine.Option(
-      names = "-M, --metadata",
+      names = {"--metadata", "-M"},
       description =
           "Custom metadata to apply to this instance.\n"
-              + "specify multiple metadata in the format of --metadata=key1=value1 --Mkey2=value2.\n"
+              + "specify multiple metadata in the format of --metadata=key1=value1 -Mkey2=value2.\n"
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -42,3 +42,6 @@ out/
 ### Nextflow
 .nextflow
 .nextflow.log
+
+### Jenv directory-local version setting
+.java-version
```
ginaygoogle, line 47: any reason for this adding? (want to learn)

ginaygoogle (approved):

src/main/java/bio/terra/cli/command/resource/create/GcpNotebook.java
```diff
@@ -64,10 +64,10 @@ public class GcpNotebook extends BaseCommand {
   private String postStartupScript;
 
   @CommandLine.Option(
-      names = "-M, --metadata",
+      names = {"--metadata", "-M"},
```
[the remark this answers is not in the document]
jaycarlton: Yep.

yuhuyoyo (approved):
thanks

src/test/java/unit/GcpNotebookControlled.java
```diff
@@ -39,7 +39,12 @@ void listDescribeReflectCreateDelete() throws IOException {
...
-            UFGcpNotebook.class, "resource", "create", "gcp-notebook", "--name=" + name);
+            UFGcpNotebook.class,
+            "resource",
+            "create",
+            "gcp-notebook",
+            "--name=" + name,
+            "--metadata=foo=bar");
 
     // check that the name and notebook name match
     assertEquals(name, createdNotebook.name, "create output matches name");
```
ginaygoogle, line 50: do we need to add the assertion for the metadata?

jaycarlton: The metadata doesn't show up when we describe the resource; it's cloud-only. If we had a convenient way to get it from the cloud, we could assert on that.

PR #269, opened 2022-06-15
PF-1694: create workspace property

ginaygoogle:
Screen Shot 2022-06-15 at 11 45 43 AM
create the workspace property
Screen Shot 2022-06-15 at 5 38 53 PM
multiple entries add

This pr only contain the create property. and the property support multiple adds


jaycarlton:
Are we sure we want to override the properties on clone? I would've expected them to be merged with the source workspace. I'm actually surprised we want to specify them on clone at all. I thought the intent of the ticket was to just clone the existing properties to the new workspace.

src/main/java/bio/terra/cli/businessobject/Workspace.java
```diff
@@ -312,6 +325,18 @@ public String grantBreakGlass(
     return granteeProxyGroupEmail;
   }
 
+  public static Properties stringMapToProperties(Map<String, String> map) {
+    Properties propertyResult = new Properties();
+    if (map == null) {
```
melissachang, line 330: Can delete 330-332

src/main/java/bio/terra/cli/businessobject/Workspace.java
```diff
@@ -312,6 +325,18 @@ public String grantBreakGlass(
...
+      return propertyResult;
+    }
+    for (Map.Entry<String, String> entry : map.entrySet()) {
+      Property prop = new Property().key(entry.getKey()).value(entry.getValue());
```
melissachang, line 334: prop -> property. In Java, usually we don't cut off words

src/main/java/bio/terra/cli/businessobject/Workspace.java
```diff
@@ -312,6 +325,18 @@ public String grantBreakGlass(
     return granteeProxyGroupEmail;
   }
 
+  public static Properties stringMapToProperties(Map<String, String> map) {
+    Properties propertyResult = new Properties();
```
melissachang, line 329: propertyResult -> properties

src/main/java/bio/terra/cli/businessobject/Workspace.java
```diff
@@ -332,6 +357,10 @@ public String getGoogleProjectId() {
     return googleProjectId;
   }
 
+  public Properties getProperty() {
```
melissachang, line 360: getProperties. Please try to pay attention to singular (property) vs plural (properties). :)

melissachang (commented):
> Are we sure we want to override the properties on clone? I would've expected them to be merged with the source workspace. I'm actually surprised we want to specify them on clone at all. I thought the intent of the ticket was to just clone the existing properties to the new workspace.

I agree we should not override on clone, we should merge. I also agree that for simplicity, we don't need to support updating property on clone.

When you clone, you definitely need to set ID, and most likely want to set name (give new workspace different name from old workspace). We don't need to support every possible field with clone.

src/main/java/bio/terra/cli/serialization/persisted/PDWorkspace.java
```diff
@@ -88,6 +93,11 @@ public Builder googleProjectId(String googleProjectId) {
       return this;
     }
 
+    public Builder property(Properties property) {
```
jaycarlton, line 96: this should be `properties`. Note that this is our last chance to change it since it's a persisted field.

src/test/java/unit/CloneWorkspace.java
```diff
@@ -182,13 +182,16 @@ public void cloneWorkspace(TestInfo testInfo) throws Exception {
             "clone",
             "--new-id=cloned_id" + RANDOM.nextInt(Integer.MAX_VALUE),
             "--name=cloned_workspace",
-            "--description=A clone.");
+            "--description=A clone.",
+            "--property=key1=value1,key2=value2");
```
jaycarlton, line 186: So this usage makes me think maybe we should name the option `--properties`. When it's used multiple times, the singular makes sense.

src/main/java/bio/terra/cli/businessobject/Workspace.java
jaycarlton: Mark the map and the function `@Nullable` if you keep these lines.

src/main/java/bio/terra/cli/command/shared/options/WorkspaceProperty.java
```diff
@@ -0,0 +1,15 @@
+package bio.terra.cli.command.shared.options;
+
+import java.util.Map;
+import picocli.CommandLine;
+
+/** @CommandLine.Mixin class for workspace options */
```
melissachang, line 6: Delete

In general for comments, please either update or delete them

src/main/java/bio/terra/cli/command/shared/options/WorkspaceProperty.java
```diff
@@ -0,0 +1,15 @@
...
+public class WorkspaceProperty {
```
melissachang, line 7: WorkspaceProperties (since flag is --properties, not --property)

src/main/java/bio/terra/cli/command/workspace/Create.java
```diff
@@ -16,7 +17,9 @@ public class Create extends BaseCommand {
   // Variable is `id` instead of `userFacingId` because user sees it with `terra workspace create`
   private String id;
 
-  @CommandLine.Mixin WorkspaceNameAndDescription workspaceNameAndDescription;
+  @CommandLine.Mixin WorkspaceNameAndDescription workspaceNameAndDescriptionAndProperty;
```
melissachang, line 20: workspaceNameAndDescriptionAndProperty -> workspaceNameAndDescription

Tip: Before asking reviewers to take another look, you may want look at your changes, to see if anything needs updating :)

src/main/java/bio/terra/cli/serialization/userfacing/UFWorkspaceLight.java
```diff
@@ -59,6 +64,11 @@ public void print() {
     OUT.println("Name:              " + name);
     OUT.println("Description:       " + description);
     OUT.println("Google project:    " + googleProjectId);
+    OUT.println("Properties:");
+    for (int counter = 0; counter < properties.size(); counter++) {
```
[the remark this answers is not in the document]
melissachang: I don't think we need to print counter. Please do this comment?

src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
```diff
@@ -213,12 +214,16 @@ public WorkspaceDescriptionList listWorkspaces(int offset, int limit) {
    * @param userFacingId required user-facing ID
    * @param name optional display name
    * @param description optional description
+   * @param properties optional property
```
melissachang, line 217: property -> properties, since the param is called properties

src/test/java/unit/Workspace.java
```diff
@@ -124,11 +124,15 @@ void statusDescribeListReflectUpdate() throws IOException {
...
     String name = "statusDescribeListReflectUpdate";
     String description = "status list reflect update";
-    UFWorkspace createdWorkspace = WorkspaceUtils.createWorkspace(testUser, name, description);
+    String property = "key1=value1";
+
+    UFWorkspace createdWorkspace =
+        WorkspaceUtils.createWorkspace(testUser, name, description, property);
 
     // check the created workspace name and description are set
     assertNotNull(createdWorkspace.name, "create workspace name is defined");
```
melissachang, line 133: Usually we assert the exact value, not just that it's not null
```
    // check the created workspace name, description, properties are set
    assertEquals(name, createdWorkspace.name);
    assertEquals(description, createdWorkspace.description);
    // here, assert property is correct

```

src/main/java/bio/terra/cli/command/workspace/Create.java
ginaygoogle: sorry for this one, I use refactor to change the name, not sure why it still here and no fail during the check...

src/main/java/bio/terra/cli/serialization/userfacing/UFWorkspaceLight.java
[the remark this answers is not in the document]
ginaygoogle: Hi Melissa. I also used this way before. But it does not work. because the properties type is properties. 
image
I am not sure why it does not allow. 

I find the reason... I should do 
    for (Property property : properties) {
      OUT.println(
          " " + property.getKey() + ": " + property.getValue());
    }
    
    to specialize the property's type is property

src/test/java/unit/Workspace.java
ginaygoogle: Thanks for the clarification this is not intentionally do like this way. 
i correct the name, description also here.

src/main/java/bio/terra/cli/command/workspace/Create.java
```diff
@@ -18,14 +19,19 @@ public class Create extends BaseCommand {
 
   @CommandLine.Mixin WorkspaceNameAndDescription workspaceNameAndDescription;
 
+  @CommandLine.Mixin WorkspaceProperties workspaceProperty;
```
melissachang, line 22: workspaceProperty -> workspaceProperties

Since class is `WorkspaceProperties`

src/main/java/bio/terra/cli/serialization/userfacing/UFWorkspaceLight.java
```diff
@@ -59,6 +65,10 @@ public void print() {
     OUT.println("Name:              " + name);
     OUT.println("Description:       " + description);
     OUT.println("Google project:    " + googleProjectId);
+    OUT.println("Properties:");
+    for (Property property : properties) {
+      OUT.println(" " + property.getKey() + ": " + property.getValue());
```
melissachang, line 70: " " -> "  " (2 spaces)

melissachang (approved):

jaycarlton (approved):

PR #270, opened 2022-06-16
PF-1728: add the metadata display in the text/json

ginaygoogle:
image
update one metadata

Screen Shot 2022-06-16 at 7 58 20 PM
add multiple entries of metadata.

### DIFF
--- a/src/main/java/bio/terra/cli/serialization/userfacing/resource/UFGcpNotebook.java
+++ b/src/main/java/bio/terra/cli/serialization/userfacing/resource/UFGcpNotebook.java
@@ -67,9 +67,7 @@ public class UFGcpNotebook extends UFResource {
     OUT.println(prefix + "Instance name: " + (instanceName == null ? "(undefined)" : instanceName));
     OUT.println(prefix + "State:         " + (state == null ? "(undefined)" : state));
     OUT.println(prefix + "Metadata:");
-    for (Map.Entry<String, String> entry : metadata.entrySet()) {
-      OUT.println("   " + entry.getKey() + ": " + entry.getValue());
-    }
+    metadata.forEach((key, value) -> OUT.println("   " + key + ": " + value));
     OUT.println(prefix + "Proxy URL:     " + (proxyUri == null ? "(undefined)" : proxyUri));
     OUT.println(prefix + "Create time:   " + (createTime == null ? "(undefined)" : createTime));
   }
--- a/src/main/java/bio/terra/cli/serialization/userfacing/resource/UFGcpNotebook.java
+++ b/src/main/java/bio/terra/cli/serialization/userfacing/resource/UFGcpNotebook.java
@@ -7,6 +7,7 @@ import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
 import com.google.api.services.notebooks.v1.model.Instance;
 import java.io.PrintStream;
+import java.util.Map;
 import java.util.Optional;
 
 /**
@@ -23,6 +24,7 @@ public class UFGcpNotebook extends UFResource {
   public final String location;
   public final String instanceName;
   public final String state;
+  public final Map<String, String> metadata;
   public final String proxyUri;
   public final String createTime;
 
@@ -36,6 +38,7 @@ public class UFGcpNotebook extends UFResource {
     Optional<Instance> instance = internalObj.getInstance();
     this.instanceName = instance.map(Instance::getName).orElse(null);
     this.state = instance.map(Instance::getState).orElse(null);
+    this.metadata = instance.map(Instance::getMetadata).orElse(null);
     this.proxyUri = instance.map(Instance::getProxyUri).orElse(null);
     this.createTime = instance.map(Instance::getCreateTime).orElse(null);
   }
@@ -48,6 +51,7 @@ public class UFGcpNotebook extends UFResource {
     this.location = builder.location;
     this.instanceName = builder.instanceName;
     this.state = builder.state;
+    this.metadata = builder.metadata;
     this.proxyUri = builder.proxyUri;
     this.createTime = builder.createTime;
   }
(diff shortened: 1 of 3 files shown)


src/test/java/unit/GcpNotebookControlled.java
```diff
@@ -211,6 +213,12 @@ void overrideLocationAndInstanceId() throws IOException {
     // stored in or accessible via Workspace Manager.
     assertEquals(newName, updatedNotebook.name, "create output matches name");
     assertEquals(newDescription, updatedNotebook.description, "create output matches description");
+    String[] metadatas = newMetadata.split(",");
```
yuhuyoyo, line 216: when you use a map, the correct format is:
--new-metadata=newFoo=newBar --new-metadata=newFoo2=newBar2. you shouldn't need to do split yourself. See usage example: https://picocli.info/#_maps

jaycarlton: If you use the `split` argument, you can also do `--new-metadata=foo=bar,baz=quux`, but yes, you don't need to call `String#split`.

src/test/java/unit/GcpNotebookControlled.java
```diff
@@ -204,13 +207,19 @@ void overrideLocationAndInstanceId() throws IOException {
...
-            "--new-metadata=" + newMetadata);
+            "--new-metadata=" + newMetadata1 + "," + newMetadata2);
 
     // check that the properties match
     // the metadata supports multiple entries, we can't assert on the metadata because it's not
     // stored in or accessible via Workspace Manager.
     assertEquals(newName, updatedNotebook.name, "create output matches name");
     assertEquals(newDescription, updatedNotebook.description, "create output matches description");
+    assertTrue(
+        updatedNotebook.metadata.toString().contains(newMetadata1),
```
yuhuyoyo, line 218: i think i prefer `assertEquals(metadata1, updatedNotebook.metadata.get(NewMetadata1))` than converting map to string.

jaycarlton: ➕

src/test/java/unit/GcpNotebookControlled.java
```diff
@@ -206,19 +207,19 @@ void overrideLocationAndInstanceId() throws IOException {
             "--name=" + name,
             "--new-name=" + newName,
             "--new-description=" + newDescription,
-            "--new-metadata=" + newMetadata);
+            "--new-metadata=" + newMetadata1 + "," + newMetadata2);
```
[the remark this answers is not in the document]
jaycarlton: By adding the `split` option, now the user can do it either way. We should probably do this for the other key-value options.

src/main/java/bio/terra/cli/serialization/userfacing/resource/UFGcpNotebook.java
```diff
@@ -62,6 +66,10 @@ public void print(String prefix) {
     OUT.println(prefix + "Instance id:       " + instanceId);
     OUT.println(prefix + "Instance name: " + (instanceName == null ? "(undefined)" : instanceName));
     OUT.println(prefix + "State:         " + (state == null ? "(undefined)" : state));
+    OUT.println(prefix + "Metadata:");
+    for (Map.Entry<String, String> entry : metadata.entrySet()) {
+      OUT.println("   " + entry.getKey() + ":" + entry.getValue());
```
jaycarlton, line 71: Put a space after the colon like ": ". Also try a tab to see what looks better (":\t").

src/test/java/unit/GcpNotebookControlled.java
```diff
@@ -48,6 +48,8 @@ void listDescribeReflectCreateDelete() throws IOException {
 
     // check that the name and notebook name match
     assertEquals(name, createdNotebook.name, "create output matches name");
+    assertTrue(
+        createdNotebook.metadata.toString().contains("foo=bar"), "create output matches metadata");
```
jaycarlton, line 52: You don't need to convert the map to a string. Do a `get()` on it like `assertEquals("bar", createdNotebook.metadata.get("foo"))`

jaycarlton (approved):

src/test/java/unit/GcpNotebookControlled.java
```diff
@@ -194,7 +195,8 @@ void overrideLocationAndInstanceId() throws IOException {
     // new key-value pair will be appended, existing key-value pair will be updated.
     String newName = "NewOverrideLocationAndInstanceId";
     String newDescription = "\"new override default location and instance id\"";
-    String newMetadata = "NewMetadata1=metadata1,NewMetadata2=metadata2";
+    String newMetadata1 = "NewMetadata1=metadata1";
```
jaycarlton, line 198: I'd declare `String newKey1 = "NewMetadata1;`, etc, and then do `String newEntry1 = newKey1 + "=" + newValue1;`. Then when you go to do your assertions, you don't have to split to get the expected value for `newValue1`.

src/main/java/bio/terra/cli/serialization/userfacing/resource/UFGcpNotebook.java
```diff
@@ -62,6 +66,10 @@ public void print(String prefix) {
     OUT.println(prefix + "Instance id:       " + instanceId);
     OUT.println(prefix + "Instance name: " + (instanceName == null ? "(undefined)" : instanceName));
     OUT.println(prefix + "State:         " + (state == null ? "(undefined)" : state));
+    OUT.println(prefix + "Metadata:");
+    for (Map.Entry<String, String> entry : metadata.entrySet()) {
```
jaycarlton, line 70: You can also do 
```
metadata.forEach((key, value) -> OUT.println("   " + key + ": " + value));
```
I think it's cleaner than peeling off the entrySet when you don't have to.

src/test/java/unit/GcpNotebookControlled.java
ginaygoogle: I was also thinking that way to separate the key and value. but the concern will be we should have 4 more variables... I am not sure comparing the more redundant variables and a split sentence, which one will make the code better.

src/main/java/bio/terra/cli/serialization/userfacing/resource/UFGcpNotebook.java
ginaygoogle: great!

yuhuyoyo (approved):

PR #275, opened 2022-06-28
PF-1218: Pass nothing for PrivateResourceUser on controlled resource creation

mikenev:


src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
```diff
@@ -910,21 +906,6 @@ private static ControlledResourceCommonFields createCommonFields(
             .accessScope(createParams.accessScope)
             .managedBy(ManagedBy.USER);
```
melissachang, line 908: Looks like the following aren't used any more, can you delete them?

- CreateResourceParams.accessScope
- ControlledResourceCommonFields.accessScope

mikenev: They are used in a couple of spots. Might need a separate work item to clean those up.
The ControlledResourceCommonFields comes from a different library.

melissachang: Thanks. Sorry new to this code, CreateResourceParams.accessScope is still used.

melissachang (approved):

PR #278, opened 2022-07-07
[PF-1827] resolve return a json formatted string

yuhuyoyo:
terra resolve --name=foo_bucket --format=json
{"foo_bucket":"<project-id>/<bucket_name>"}

This command does not change by this PR: terra resolve --name=foo_bucket
project-id/bucket-name

in the following PR where i implement terra resolve --name=foo_datasource
it will be a JSONArray of JSONObjects
[
{"foo_bucket":"<project-id>/<bucket_name>"},
{"bar_dataset":"<path-to-dataset>"}
]

### DIFF
--- a/src/main/java/bio/terra/cli/command/resource/Resolve.java
+++ b/src/main/java/bio/terra/cli/command/resource/Resolve.java
@@ -67,11 +67,10 @@ public class Resolve extends BaseCommand {
   }
 
   private void printText(JSONObject object) {
-    Resource resource = Context.requireWorkspace().getResource(resourceNameOption.name);
-    OUT.println(object.get(resource.getName()));
+    OUT.println(object.get(resourceNameOption.name));
   }
 
   private void printJson(JSONObject object) {
-    OUT.println(object.toString());
+    OUT.println(object);
   }
 }
--- a/src/main/java/bio/terra/cli/command/resource/Resolve.java
+++ b/src/main/java/bio/terra/cli/command/resource/Resolve.java
@@ -11,6 +11,7 @@ import bio.terra.cli.command.shared.BaseCommand;
 import bio.terra.cli.command.shared.options.Format;
 import bio.terra.cli.command.shared.options.ResourceName;
 import bio.terra.cli.command.shared.options.WorkspaceOverride;
+import org.json.JSONObject;
 import picocli.CommandLine;
 import picocli.CommandLine.Command;
 
@@ -60,6 +61,17 @@ public class Resolve extends BaseCommand {
       default:
         cloudId = resource.resolve();
     }
-    formatOption.printReturnValue(cloudId);
+    JSONObject object = new JSONObject();
+    object.put(resource.getName(), cloudId);
+    formatOption.printReturnValue(object, this::printText, this::printJson);
+  }
+
+  public void printText(JSONObject object) {
+    Resource resource = Context.requireWorkspace().getResource(resourceNameOption.name);
+    OUT.println(object.get(resource.getName()));
+  }
+
+  public void printJson(JSONObject object) {
+    OUT.println(object.toString());
   }
 }
--- a/src/main/java/bio/terra/cli/command/resource/Resolve.java
+++ b/src/main/java/bio/terra/cli/command/resource/Resolve.java
@@ -66,12 +66,12 @@ public class Resolve extends BaseCommand {
     formatOption.printReturnValue(object, this::printText, this::printJson);
   }
 
-  public void printText(JSONObject object) {
+  private void printText(JSONObject object) {
     Resource resource = Context.requireWorkspace().getResource(resourceNameOption.name);
     OUT.println(object.get(resource.getName()));
   }
 
-  public void printJson(JSONObject object) {
+  private void printJson(JSONObject object) {
     OUT.println(object.toString());
   }
 }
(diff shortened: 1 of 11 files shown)


src/main/java/bio/terra/cli/command/resource/Resolve.java
```diff
@@ -60,6 +61,17 @@ protected void execute() {
       default:
         cloudId = resource.resolve();
     }
-    formatOption.printReturnValue(cloudId);
+    JSONObject object = new JSONObject();
```
yuhuyoyo, line 64: alternatively, I can create a JSON serializable resolve class for each individual resources. But i'd like the  JSON object's key to be the resource name.

The datasource resolve will need a jsonarray of resource _name: resource_cloud_id to be easily dereferenced.

src/main/java/bio/terra/cli/command/resource/Resolve.java
```diff
@@ -60,6 +61,17 @@ protected void execute() {
...
+    formatOption.printReturnValue(object, this::printText, this::printJson);
+  }
+
+  private void printText(JSONObject object) {
+    Resource resource = Context.requireWorkspace().getResource(resourceNameOption.name);
+    OUT.println(object.get(resource.getName()));
+  }
+
+  private void printJson(JSONObject object) {
+    OUT.println(object.toString());
```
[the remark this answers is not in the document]
zloery: I don't think you need an additional signature to do custom printJson + default printText, you could use the method specifying both but use the Format::printText since it's public and static (something like `formatOption.printReturnValue(object, Format::printText, this::printJsonValue);`).

It only saves one duplicate line in a method that's probably not going to change though, so either way is fine

zloery (approved):

melissachang (commented):
`terra resolve --name=foo_bucket` should not return json, it should just return a string, as it currently does.

`terra resolve --name=foo_bucket --format=json` should return json

`terra resolve --name=foo_datasource` can return json

melissachang:
> `terra resolve --name=foo_bucket` should not return json, it should just return a string, as it currently does.
> 
> `terra resolve --name=foo_bucket --format=json` should return json
> 
> `terra resolve --name=foo_datasource` can return json

Actually, could we be consistent with other commands?

`terra resolve --name=foo_datasource` prints a table, similar to `terra resource list` or `terra resource describe --name=foo_datasource`

`terra resolve --name=foo_datasource --format=json` prints json

I don't feel *as* strongly about this. If you really want, `terra resolve --name=foo_datasource` can print json. But I do feel strongly that `terra resolve --name=foo_bucket` should not print json. (In a notebook you just want the string, you don't want to fiddle with json.)

yuhuyoyo:
> `terra resolve --name=foo_bucket` should not return json, it should just return a string, as it currently does.
> 
> `terra resolve --name=foo_bucket --format=json` should return json
> 

yeah that is still the behavior. this change doesn't print json if not specified as json. see printText.

> `terra resolve --name=foo_datasource ` can return json

this.. i'm planning to print jsonstring for printText.

yuhuyoyo:
@melissachang I see why you thought that. I corrected the pr description.

ginaygoogle (approved):

src/test/java/harness/TestCommand.java
```diff
@@ -101,14 +101,23 @@ public static void runCommandExpectSuccess(String... args) {
     assertEquals(0, cmd.exitCode, "exit code = success");
   }
 
+  /**
+   * Helper method to run a command, check its exit code is 0=success, and return {@link Result}.
```
melissachang, line 105: This test code took me a while to understand:

```
    JSONObject resolved =
        new JSONObject(
            TestCommand.runAndGetResultExpectSuccess("resource", "resolve", "--name=" + name)
                .stdOut);
```

Instead could we have:

```
    JSONObject resolved =
            TestCommand.runAndGetJSONObjectExpectSuccess("resource", "resolve", "--name=" + name);
```

Ie put `new JSONObject` and `.stdOut` in this helper method?

yuhuyoyo: I thought about having a method to return JSONObject specifically but i like the flexibility of returning `Result` because later this will be a jsonarray of jsonobject instead of a jsonobject for datasource resolve. but since this run command have to succeed, i only need to return stdOut. I will change to that and return string.

yuhuyoyo: i don't follow. both JsonObject and JsonArray are JsonValue but I would still need to create either one of them in the helper method. How would i know which one to create?

melissachang: I was thinking, have the helper method return JsonValue. In the test, you can cast/convert to JsonObject or JsonArray.

Now I'm thinking, can you have two helper methods, one that returns JsonObject and one that returns JsonArray? I think it's worth it to make the tests easier to read.

```
Use this when there's no Java class (eg UDFGitRepo.class) for the returned JsonObject
runAndGetJsonObjectExpectSuccess

Use this when there's no Java class for the returned JsonArray
runAndGetJsonArrayExpectSuccess
```

yuhuyoyo: jsonvalue doesn't have a constructor.

i don't think having new JSONObject(string) is hard to read but i can do it this way if you like.

PR #279, opened 2022-07-07
[PF-1513] Add gradle test retry plugin

zloery:
Per gradle discussion, I'm adding this plugin to start tracking test flakes and automatically retry some. This plugin will retry failing tests up to twice, which should reduce the number of times we manually rerun tests on PRs.

This covers up test flakiness but doesn't fix it, if we want shorter PR/nightly tests we'll still need to actually fix the causes of failure. The flake data from this plugin should help, though.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -368,14 +368,12 @@ task getDockerImageTag {
 test {
     retry {
         // Max retries per test.
-        maxRetries = 2
+        maxRetries = 1
         // Max total test failures.
         // This is an estimate based on current failure rates, but if more than
         // 5 tests fail in a run it's likely a real source of failure and we
         // should stop retrying.
         maxFailures = 5
-        // TODO: until test flakiness is at a manageable level, accept flaky
-        //  tests as long as they pass once.
-        failOnPassedAfterRetry = false
+        failOnPassedAfterRetry = true
     }
 }
--- a/build.gradle
+++ b/build.gradle
@@ -367,13 +367,17 @@ task getDockerImageTag {
 
 test {
     retry {
-        // Max retries per test.
-        maxRetries = 1
-        // Max total test failures.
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -363,3 +364,18 @@ task getDockerImageTag {
...
+test {
+    retry {
+        // Max retries per test.
+        maxRetries = 2
+        // Max total test failures.
+        // This is an estimate based on current failure rates, but if more than
+        // 5 tests fail in a run it's likely a real source of failure and we
+        // should stop retrying.
+        maxFailures = 5
+        // TODO: until test flakiness is at a manageable level, accept flaky
```
yuhuyoyo, line 377: nit: attach ticket to TODO

yuhuyoyo (approved):

melissachang (commented):
Thanks

build.gradle
```diff
@@ -363,3 +364,16 @@ task getDockerImageTag {
         println("${project.properties['dockerImageTag']}")
     }
 }
+
+test {
```
melissachang, line 368: Forgot to mention, can we do "Limiting retry to CI builds"?

https://github.com/gradle/test-retry-gradle-plugin#limiting-retry-to-ci-builds

For local runs, often I'm working on a legitimately failing test

melissachang (approved):

melissachang (approved):
May want to give team a heads up -- When they're oncall, they'll still get email for test failures and they still should file verily_triage tickets. They can just skip the "rerun workflow and see if it passed" step.

melissachang (approved):
Sorry, thought of something else. Totally optional --

This is meant to be a common Terra gradle plugin:

https://github.com/DataBiosphere/terra-gradle-plugin

Currently doesn't do anything, and not used.

You could put this in there. Then the various repos could include the common plugin.

PR #281, opened 2022-07-12
[PF-1665] Poll destination workspace for clone workspace job ID

zloery:
In https://github.com/DataBiosphere/terra-workspace-manager/pull/710 WSM began validating that job IDs were inside the appropriate workspace for all calls getting job results. One quirk this uncovered was that for `cloneWorkspace` jobs, the job lives inside the destination workspace, not the source workspace. This is different behavior from cloneResource endpoints, which live inside the source workspace. The CLI has been calling this endpoint incorrectly the whole time, we just didn't notice because we did not check which workspace jobs lived in.

(This fixes the CloneWorkspace test which began failing with the WSM change. I'm still not exactly sure how the last night's tests passed after I merged the change yesterday afternoon)


yuhuyoyo (approved):
thanks for the quick fix Zach!

PR #284, opened 2022-07-19
[PF-1513] Move test retry config inside runTestsWithTag

zloery:
I thought we could have the `retry` block inside the `test` task since `runTestsWithTask` has type `Test`, but it looks like we never actually retried tests.

Tested this locally, the new configuration will actually retry tests as expected.


melissachang (approved):
Sweet!

yuhuyoyo (approved):

PR #288, opened 2022-07-22
[PF-1859] created date and last udpated date can be null

yuhuyoyo:
It a rare case but when it happens, we don't want to just throw NPE. createdate and lastupdatedate are not required field in the workspaceDescription.

### DIFF
--- a/src/main/java/bio/terra/cli/serialization/userfacing/resource/UFDataCollection.java
+++ b/src/main/java/bio/terra/cli/serialization/userfacing/resource/UFDataCollection.java
@@ -189,7 +189,7 @@ public class UFDataCollection extends UFResource {
 
   private void printDate(String prefix, String dateLabel, @Nullable OffsetDateTime dateTime) {
     if (dateTime == null) {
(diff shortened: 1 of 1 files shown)


src/main/java/bio/terra/cli/serialization/userfacing/resource/UFDataCollection.java
```diff
@@ -183,7 +187,11 @@ public UFDataCollection build() {
     public Builder() {}
   }
 
-  private void printDate(String prefix, String dateLabel, OffsetDateTime dateTime) {
+  private void printDate(String prefix, String dateLabel, @Nullable OffsetDateTime dateTime) {
+    if (dateTime == null) {
+      logger.warn(String.format("datetime for %s is null", dateLabel));
```
melissachang, line 192: opt Delete logger.warn, or add `This is expected for old workspaces`

melissachang (approved):

zloery (approved):
LGTM
please add a ticket number to the PR title or it'll make Gary sad

PR #296, opened 2022-07-28
[PF-1750] Alllow dashes in resource name

melissachang:
WSM allows dashes:

https://cs.github.com/DataBiosphere/terra-workspace-manager/blob/4eabe67ee705645c85cb9f522e739fb8b8a77228/service/src/main/java/bio/terra/workspace/service/resource/ResourceValidationUtils.java?q=resource_name#L103

CLI currently doesn't. Change CLI to match WSM.

Before:

```
~/terra-cli (mc/allow-dash): terra resource create gcp-notebook --post-startup-script=gs://melchangbucketwithnounderscore/post-startup.sh --name="resource-name-dash"
Invalid resource name specified. Name must be 1 to 1024 alphanumeric characters, underscores, and dashes and must not start with a dash or underscore.
```

After:

```
~/terra-cli (mc/allow-dash): terra resource create gcp-notebook --post-startup-script=gs://melchangbucketwithnounderscore/post-startup.sh --name="resource-name-dash"
Successfully added controlled GCP Notebook instance.
```

Also fix bug when resource name has dashes. Environment variables don't allow dashes.

Before:
- TERRA_ env variable wouldn't work because it has dash

After:
- TERRA_ env variable works, no longer has dash

### DIFF
--- a/src/main/java/bio/terra/cli/app/CommandRunner.java
+++ b/src/main/java/bio/terra/cli/app/CommandRunner.java
@@ -154,7 +154,7 @@ public abstract class CommandRunner {
   }
 
   // Note: `foo-bar` and `foo_bar` will have the same env variable. PF-1907 will fix this.
-  private static final String convertToEnvironmentVariable(String string) {
+  public static final String convertToEnvironmentVariable(String string) {
     return "TERRA_" + string.replace("-", "_");
   }
 
--- a/src/main/java/bio/terra/cli/app/CommandRunner.java
+++ b/src/main/java/bio/terra/cli/app/CommandRunner.java
@@ -123,7 +123,8 @@ public abstract class CommandRunner {
         .forEach(
             resource -> {
               if (Resource.Type.DATA_COLLECTION != resource.getResourceType()) {
-                terraReferences.put("TERRA_" + resource.getName(), resource.resolve());
+                String envVariable = convertToEnvironmentVariable(resource.getName());
+                terraReferences.put(envVariable, resource.resolve());
               } else {
                 Workspace dataCollectionWorkspace = null;
                 try {
@@ -139,14 +140,22 @@ public abstract class CommandRunner {
                           // This should NEVER happen but check here to prevent endless resolve.
                           r -> Resource.Type.DATA_COLLECTION != r.getResourceType())
                       .forEach(
-                          r ->
-                              terraReferences.put(
-                                  "TERRA_" + resource.getName() + "_" + r.getName(), r.resolve()));
+                          r -> {
+                            String envVariable =
+                                convertToEnvironmentVariable(
+                                    resource.getName() + "_" + r.getName());
+                            terraReferences.put(envVariable, r.resolve());
+                          });
                 }
               }
             });
 
     return terraReferences;
+  }
+
+  // Note: `foo-bar` and `foo_bar` will have the same env variable. PF-1907 will fix this.
+  private static final String convertToEnvironmentVariable(String string) {
+    return "TERRA_" + string.replace("-", "_");
   }
 
   /**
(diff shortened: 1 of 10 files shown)


src/main/java/bio/terra/cli/app/CommandRunner.java
```diff
@@ -139,16 +140,24 @@ private Map<String, String> buildMapOfTerraReferences() {
...
                 }
               }
             });
 
     return terraReferences;
   }
 
+  private static final String convertToEnvironmentVariable(String string) {
+    // Env variables can't have "-"
+    return string.replace("-", "_");
```
yuhuyoyo, line 158: oh sorry i forgot CLI didn't allow dashes because of environment variable.

but i don't think this would guarantee uniqueness of the env names. if there is a bq-1 and a bq_1, both of them will be TERRA_bq_1 envVariable.

melissachang: Replaced with double underscore instead of single underscore. Technically there could still be a conflict but chances are extremely low.

I'm concerned that double underscore is unintuitive -- people will try single underscore and wonder why it doesn't work. We'll see what happens.

src/main/java/bio/terra/cli/app/CommandRunner.java
```diff
@@ -123,7 +123,8 @@ private Map<String, String> buildMapOfTerraReferences() {
         .forEach(
             resource -> {
               if (Resource.Type.DATA_COLLECTION != resource.getResourceType()) {
-                terraReferences.put("TERRA_" + resource.getName(), resource.resolve());
+                String envVariable = convertToEnvironmentVariable("TERRA_" + resource.getName());
```
zloery, line 126: optional nit: move the `"TERRA_"+` inside `convertToEnvironmentVariable`?

zloery (approved):
LGTM

yuhuyoyo (approved):

yuhuyoyo:
could you attach ticket to PR title?

src/main/java/bio/terra/cli/app/CommandRunner.java
melissachang: Done

melissachang:
> could you attach ticket to PR title?

Not sure which ticket you're referring to. I put PF-1750. Initially I didn't put PF-1750 because this isn't really PF-1750 (which is a new feature), this is fixing a pre-existing bug.

melissachang:
Hmm the env variables look like:

```
TERRA_1000__genomes__ref4399_bucket_resource_name3182
```

That looks pretty weird to me. @yuhuyoyo could I replace with single underscore for now? It's unlikely that there will be 2 resources that are the same, except for - _.

While the - _ corner case is broken, this PR still fixes a couple major things:

- Allow creating a resource with `-` in name
- Before this PR, for resource name with `-`, all env variables broke. This PR fixes that.

melissachang:
> Hmm the env variables look like:
> 
> ```
> TERRA_1000__genomes__ref4399_bucket_resource_name3182
> ```
> 
> That looks pretty weird to me. I think our users might do `1000_genomes` by accident. @yuhuyoyo could I replace with single underscore for now? It's unlikely that there will be 2 resources that are the same, except for - _.
> 
> I think better fixes:
> 
> * Have WSM enforce resources can't have same name, except for -_. So you can't have 1000_genomes and 1000-genomes.
> * If user has 2 resource names that conflict, user change one of them
> 
> While the - _ corner case is broken, this PR still fixes a couple major things:
> 
> * Allow creating a resource with `-` in name
> * Before this PR, for resource name with `-`, all env variables broke. This PR fixes that.

Talked to Yu. Changed back to single underscore. Filed WSM PF-1907 to have WSM enforce you can't have `foo-bar` and `foo_bar`.

PR #301, opened 2022-08-17
[PF-1922] Apply Formatting to repo: terra-cli

dexamundsen:
Apply formatting to entire repo (no manual changes)

### DIFF
--- a/.github/workflows/cleanup-test-user-workspaces.yml
+++ b/.github/workflows/cleanup-test-user-workspaces.yml
@@ -14,74 +14,74 @@ jobs:
   cleanup-test-user-workspaces:
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout current code
-        id: checkout_code
-        uses: actions/checkout@v2
-        with:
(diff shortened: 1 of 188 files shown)


yuhuyoyo:
Could you kindly figure out why the lint check is not working?

mikenev:
What's driving this change? We do have spotless and the lint check, was it not applying rules or something?

dexamundsen:
@mikenev - looks like the google java format was not applied on the IDE. I applied this as part of JDK upgrade and lead to a whole lot of format changes. Hence split it to a separate PR
@yuhuyoyo - Fixed the check. There is an issue with single line comments: the rules for google java format and spotless check were conflicting

mikenev (approved):
Ok, that makes sense. Thanks for doing this in a separate review.

yuhuyoyo (approved):
this is impossible to review so i will just trust you :)

yuhuyoyo:
filed https://broadworkbench.atlassian.net/browse/PF-1924

PR #261, opened 2022-05-31
Revert "[PF-1654] Upgrade Gradle to 7.4.2"

ddietterich:
Reverts DataBiosphere/terra-cli#260

The change to Gradle broke the release version of the CLI. Tests failed since this was merged, so we are reverting until it can be properly debugged.

PR #303, opened 2022-08-17
[PF-1554] Gradle & JDK upgrade for terra-cli

dexamundsen:
Below changes are made to keep this consistent with other services
- Gradle upgrade from 6.x to 7.4.1
- JDK upgrade from 11 to 17 
- Add Gradle dependency lock in build.gradle
- Move lib version to settings.gradle (and related github workflow update)

Subsequent changes that were required
- Update Docker image used to JDK17
- Library / code updates
- Dependency version updates
- Update documentation

test release (pre-release, will be deleted after this PR is merged) - https://github.com/DataBiosphere/terra-cli/releases/tag/test123

notebook JDK update - https://github.com/DataBiosphere/terra-workspace-manager/pull/776


build.gradle
```diff
@@ -8,32 +8,31 @@ plugins {
...
-if (hasProperty('buildScan')) {
-    buildScan {
-        termsOfServiceUrl = 'https://gradle.com/terms-of-service'
-        termsOfServiceAgree = 'yes'
-    }
-}
+// If true, search local repository (~/.m2/repository/) first for dependencies.
+def useMavenLocal = false
 
-// The tools/publish-release.sh script depends on this version string being of the format "version = '1.2.3'"
```
melissachang, line 28: Curious why this moved to settings.gradle?

settings.gradle
```diff
@@ -16,3 +14,9 @@ gradleEnterprise {
         uploadInBackground = System.getenv("CI") == null
     }
 }
+
+rootProject.name = 'terra-cli'
+
+enableFeaturePreview('ONE_LOCKFILE_PER_PROJECT')
```
melissachang, line 20: nit blank line after this line. Actually, could this go somewhere higher in the file, since this is important?

gradle.lockfile
```diff
@@ -0,0 +1,206 @@
+# This is a Gradle generated file for dependency locking.
```
melissachang, line 1: What problem is this fixing? I'd lean towards not adding this extra complexity, if there's nothing to fix...

src/main/java/bio/terra/cli/command/shared/options/WorkspaceNameAndDescription.java
```diff
@@ -2,7 +2,9 @@
 
 import picocli.CommandLine;
 
-/** @CommandLine.Mixin class for workspace name and description options */
```
melissachang, line 5: I believe we're following Google Java style guide. Having it on one line is valid...

https://google.github.io/styleguide/javaguide.html#s7.2-summary-fragment:~:text=...%20or%20in%20this%20single%2Dline%20example%3A

melissachang (commented):
Thanks for working on this!

Looks like some tests are failing.

You may already be aware -- earlier Jay tried upgrading Gradle. We had to roll it back (https://github.com/DataBiosphere/terra-cli/pull/261). I think this was the failing run:

https://github.com/DataBiosphere/terra-cli/runs/6664717857?check_suite_focus=true

Can you make sure that test-nightly.yml works on your test release?


We have 2 modes -- TERRA_CLI_DOCKER_MODE is DOCKER_NOT_AVAILABLE or DOCKER_AVAILABLE. As a sanity check, could you try running some manual commands in both modes? (Our tests only test docker mode.)

src/main/java/bio/terra/cli/command/shared/options/WorkspaceNameAndDescription.java
dexamundsen: This was change was auto-included when formatting was applied, will recheck

build.gradle
```diff
@@ -42,93 +41,80 @@ repositories {
...
+    testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter', version: '5.9.0'
+    testImplementation group: 'org.hamcrest', name: 'hamcrest', version: '2.2'
+
+    // PubSub is used in tests to communicate with Janitor, but is not required for the CLI itself.
+    testImplementation group: 'com.google.cloud', name: 'google-cloud-pubsub', version: '1.120.11'
+
+    annotationProcessor group: 'info.picocli', name: 'picocli-codegen', version: '4.6.3'
+}
+
+dependencyLocking {
```
yuhuyoyo, line 108: thanks for adding this and reordering all the dependency... but i prefer having one PR to do one thing especially since github diff viewer is crazy...

gradle.lockfile
dexamundsen: 2 reasons I added this for, open to discuss and remove if the value add does not outweigh the added complexity

- Versions of dependencies matter here: incompatible versions might cause errors during compile / runtime. Some of the libraries did not specify the version to be included, so pulling the latest version might lead to breaking changes. Hence added this to avoid them. Example: some libraries in the build file refer to older library versions due to compatibility errors 
- Keeping it consistent with other newer repos (like WSM).. this repo is relatively older and does not align to the newer standards followed. Consistent structure eases maintenance 

(Will sync up when rest pr if ready - these changes are easy to remove)

settings.gradle
dexamundsen: Sure, the line was auto-added.
Will move project name to the top

src/test/java/harness/CRLJanitor.java
```diff
@@ -101,17 +102,16 @@ public static void registerWorkspaceForCleanup(UUID uuid, TestUser testUser) {
...
             .resourceUid(
                 new CloudResourceUid()
-                    .terraWorkspace(
-                        new TerraWorkspaceUid()
-                            .workspaceId(uuid)
-                            .workspaceManagerInstance(wsmInstance)))
-            .resourceMetadata(new ResourceMetadata().workspaceOwner(testUser.email))
-            .creation(OffsetDateTime.now())
-            .expiration(OffsetDateTime.now().plus(WORKSPACE_TIME_TO_LIVE));
+                    .googleProjectUid(
```
yuhuyoyo, line 110: can we please split this to a different PR? doesn't seem related to jdk upgrade.

build.gradle
dexamundsen: Keeping this consistent structure across other repos. 
I will further update the build files in this PR itself --> keeping it more modular.

.github/workflows/release-on-pr-merge.yml
```diff
@@ -2,8 +2,8 @@ name: Bump version and publish release
 on:
   workflow_dispatch: { }
   push:
-    branches:
-    - main
+    branches: [ master ]
```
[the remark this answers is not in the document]
dexamundsen: this is fixed already

docker/Dockerfile
```diff
@@ -1,5 +1,5 @@
-# Java 11 base image
-FROM openjdk:11.0-jre-slim
+# Java 17 base image
+FROM eclipse-temurin:17.0.4_8-jre-alpine
```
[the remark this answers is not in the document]
dexamundsen: OpenJDK is being deprecated and they recommend moving to other build - 
https://hub.docker.com/_/openjdk has this notice

Eclipse temurin is now maintained by Adoptium

build.gradle
dexamundsen: Sure, I plan to get All tests working and then Split the changes - this will make it much easier to avoid breaking any of the tests

src/test/java/harness/CRLJanitor.java
dexamundsen: This is actually required due to a breaking change when the library versions were upgraded (to keep up with Java 17 upgrade).

PR #306, opened 2022-08-22
fix the autoname bucket name bug

ginaygoogle:
This ticket solves 2 issues after the PF-1750 auto-generation cloud-native id.
Both changes in the unit test GcsBucketControlled.java. 
-  the generated bucket name is not contain "bucket". It contains the originalName.toLowerCase(). (the original name does not have any character, so only to lower case is enough)
- we do not allow the same bucket name in the same workspace. I changed the bucket name is unique. (feel a little confused each unit test, the resource will be delete in the end, it could not have same bucket name, maybe rerun the whole unit test can solve, I still change the name unique in order to be clear with each unit test with their individual specified bucket name.)


yuhuyoyo:
Can you file a ticket and attach it to the PR please?

yuhuyoyo (approved):

mikenev (approved):

PR #310, opened 2022-08-23
PF-1924: Add an action to run spotlessApply

yuhuyoyo:


.github/workflows/tests-nightly.yml
```diff
@@ -3,7 +3,8 @@ on:
   workflow_dispatch: { }
   schedule:
   - cron: '0 5 * * *' # 5AM UTC = 12AM EST
-
+env:
+  PR_REVIEWERS: "ddietterich,yuhuyoyo,zloery,mikenev,melissachang"
```
dexamundsen, line 7: could we notify the on-call directly? Refer to chat on devops group. 
this way we dont need to maintain this list.

.github/actions/spotless-apply/action.yml
```diff
@@ -0,0 +1,43 @@
...
+      commit-message: "Terra CLI spotless apply"
+      committer: terracli-bot <noreply@github.com>
+      author: ${{ github.actor }} <${{ github.actor }}@users.noreply.github.com>
+      title: "Terra CLI spotless apply"
+      branch: "actions/spotlessApply"
+      delete-branch: true
+      body: |
+        Run spotlessApply.
+        *Note: This PR was opened by the [nightly-test GitHub Actions workflow](https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}).*
+      reviewers: ${{ inputs.reviewers }}
```
dexamundsen, line 43: nit: please add new line at the end

.github/workflows/tests-nightly.yml
yuhuyoyo: actually we are already notifying slack about nightly outcome so I don't think sending email is necessary.

dexamundsen (approved):

melissachang (commented):
I don't understand the problem. Our repo is configured so that you can't merge if spotlessCheck fails.

First, lint-and-static-analysis runs `./gradlew spotlessCheck`:

image

Second, the run test task runs spotlessApply:

image

yuhuyoyo:
for some reason it has broken before. Dex may be able to speak more about why but he recently had to manually do a spotlessapply and that PR has hundreds of files.

melissachang:
I still have the questions I asked on Aug 24. @dexamundsen , how did you generate your formatting PR?

I also have the same questions that I had for #1938. This GHA is highly unusual. Usually for a formatting PR:

1. Someone runs formatting command
2. Someone sends PR with just output from formatting command

The output of the normal process and the output of this action (the resulting PR) are the same. There are many kids of PRs where you do one thing, then you send out a PR for review. We don't create actions for all those things. I don't understand the benefit of this action.

yuhuyoyo:
> I still have the questions I asked on Aug 24. @dexamundsen , how did you generate your formatting PR?
> 
https://github.com/DataBiosphere/terra-cli/pull/301
this is the PR I'm referring to. 

> I also have the same questions that I had for #1938. This GHA is highly unusual. Usually for a formatting PR:
> 
> 1. Someone runs formatting command
> 2. Someone sends PR with just output from formatting command
> 
> The output of the normal process and the output of this action (the resulting PR) are the same. There are many kids of PRs where you do one thing, then you send out a PR for review. We don't create actions for all those things. I don't understand the benefit of this action.
maybe they all should be automated. My principle here is that if it's not reviewable, it shouldn't need to get reviewed. And a human created PR should be reviewed.


I didn't look at the 186 files to stamp it. So this kind of PR that can be done automatically shouldn't require human action. i could accidentally get some exploration changes into this PR and no one would notice it (186 files!) Google has those bots that generate PR to clean up dead code, etc. This is not a frequent thing. If spotlessCheck passes, nothing will be done.

dexamundsen:
Apologies for the delayed response - on PR https://github.com/DataBiosphere/terra-cli/pull/301 -> I set the TDE to use google formatting and applied formatting manually (right click, format). Since the gradle build (both local and as part of PR) did not change any further files, did not second guess the IDE applied formatting

melissachang:
> Apologies for the delayed response - on PR #301 -> I set the TDE to use google formatting and applied formatting manually (right click, format). Since the gradle build (both local and as part of PR) did not change any further files, did not second guess the IDE applied formatting

Ah, that's not the usual way we do formatting. As I said on Aug 24, we run spotlessApply/spotlessCheck on all code before merging. We don't need an action to run spotlessApply. It would do nothing, since spotlessApply was already run on the code.

FYI - google-java-format produces slightly different results from spotlessApply. If you want Intellij to run spotlessApply instead of google-java-format, [here's how](https://github.com/diffplug/spotless/issues/200#issuecomment-1063519752).

PR #307, opened 2022-08-23
[PF-1932] Upgrade gradle to 7.4.1

dexamundsen:
Upgrade gradle to 7.4.1 in repo: terra-cli
Target version choice based on version used in other services (keeping it consistent)
Value add - support for Java 17

All changes in this PR are auto-generated. 
Command used for upgrade --> 

`>> ./gradlew wrapper --gradle-version 7.4.1`

Split from https://github.com/DataBiosphere/terra-cli/pull/303

Nightly test run successful - https://github.com/DataBiosphere/terra-cli/runs/7964039465?check_suite_focus=true

Prev attempt - https://github.com/DataBiosphere/terra-cli/pull/261
was rolled back due to nightly test failure - https://github.com/DataBiosphere/terra-cli/runs/6664717857?check_suite_focus=true

PR #311, opened 2022-08-24
PF-1938: Add an action to generate PR for gradle upgrade

yuhuyoyo:
This is currently a manual trigger only GHA where the version needs to be specified. intentional manual update is preferred because sometimes gradle update requires a human eye to debug.

Because the version upgrade usually makes some automatic changes to the gradle files which makes it hard to review, having a GHA to generate the PR is easier. An example PR https://github.com/DataBiosphere/terra-cli/pull/307 where we update gradlew. It's purely generated code but it's not simply a number change (4 files are touched). So it adds challenge to the reviewer to tell what is changed by the command. 

With this PR, if someone needs to 1. run GHA, 2. pull it and change more file, 3. push it again, it's clear which part is auto-generated, which part is a manual fix as commit leaves a trace.
without PR, if someone just run the command and make more changes, it hard to tell which part is generated.


melissachang (commented):
Why do we need an action? Can't someone just run `./gradlew wrapper --gradle-version` when they want to upgrade? This repo has been around for 1.5 years and we've only wanted to upgrade once.

melissachang:
> Because the version upgrade usually makes some automatic changes to the gradle files which makes it hard to review, having a GHA to generate the PR is easier.

If the PR is only for gradle upgrade (eg #307), I think it's easy to review. You just ignore changes to files other than `[gradle-wrapper.properties](https://github.com/DataBiosphere/terra-cli/pull/307/files#diff-40640fe1078ece83d7ea8fb67daacd77923a86d13447baf9769660b3b46f2ece)`. Also, I don't see how having an action helps make it easier to review.

yuhuyoyo:
> > Because the version upgrade usually makes some automatic changes to the gradle files which makes it hard to review, having a GHA to generate the PR is easier.
> 
> If the PR is only for gradle upgrade (eg #307), I think it's easy to review. You just ignore changes to files other than `[gradle-wrapper.properties](https://github.com/DataBiosphere/terra-cli/pull/307/files#diff-40640fe1078ece83d7ea8fb67daacd77923a86d13447baf9769660b3b46f2ece)`. Also, I don't see how having an action helps make it easier to review.

I'm sure no one would be malicious about it but if we leaves it up to a human, there can always be files mixed up and there could be possibilities that undesired change sneak in. So instead of running the command you just trigger an action which is an equal amount if not less work.  From my perspectives, example PR you pointed out touches 4 files. When i was reviewing it, I don't really look at them closely and just approved assuming that it's generated by the command.

dexamundsen:
Would recommend against adding this as a github action, or automating any part of version upgrades. 
reason - any change of version leads to concerns of compatibility errors. We would want the engineer to verily versions, update plugins and potentially have extra tests that are not covered by automated tests. 

for example - 

- I am debugging a Class path error that surfaces only when Terra cli is run in docker mode. The root cause is that Grade 6->7 upgrade changes the underlying java set up in the auto-generated gradle wrappers
- found another issue where the images generated on laptop failed on docker, while the ones on github release were successful for the same source code (rc: image platform arch)
- Changing the gradle version will require some of the plugins to be updated as well, so the PR will likely need manual updates

in summary - such version upgrades benefit from manual checks. Besides, they are pretty rare and we typically don't want to pick the latest version especially if there are known bugs or vulnerabilities.

yuhuyoyo:
> Would recommend against adding this as a github action, or automating any part of version upgrades. reason - any change of version leads to concerns of compatibility errors. We would want the engineer to verily versions, update plugins and potentially have extra tests that are not covered by automated tests.
> 
> for example -
> 
> * I am debugging a Class path error that surfaces only when Terra cli is run in docker mode. The root cause is that Grade 6->7 upgrade changes the underlying java set up in the auto-generated gradle wrappers
> * found another issue where the images generated on laptop failed on docker, while the ones on github release were successful for the same source code (rc: image platform arch)
> * Changing the gradle version will require some of the plugins to be updated as well, so the PR will likely need manual updates
> 
> in summary - such version upgrades benefit from manual checks. Besides, they are pretty rare and we typically don't want to pick the latest version especially if there are known bugs or vulnerabilities.

I would hope test will catch the failure when the pr is generated. This GHA is not a cron job and can only be triggered by manual action. so it would be like your flow except instead of manually run the command, you go to GHA to create the PR.

melissachang (approved):
> so it would be like your flow except instead of manually run the command, you go to GHA to create the PR.

I still don't understand how this PR helps. Sounds like you're worried about accidental files in this PR unrelated to this PR. Without this PR:

1. `./gradlew wrapper --gradle-version`
2.  Do manual testing, possibly add more files to PR.

With this PR:

1. Run action
2. Do manual testing, possibly add more files to PR.

In both cases, someone could accidentally add a file in step 2. How does it matter what step 1 is? (Also, someone could accidentally add a file for any PR. Review is supposed to catch this.)

If you really want this action, please add a comment at the top explaining why, because I don't understand. :)

yuhuyoyo:
Sure, added more to description.

melissachang:
> Sure, added more to description.

Thanks, do you mind also adding comment to [update-gradle-version.yml](https://github.com/DataBiosphere/terra-cli/pull/311/files#diff-f33abcbf86283f56c1dde6e7ab2014624608d145d44c3e4694affd5f9a26d114) ? That's the first place I would look.

dexamundsen (approved):

PR #312, opened 2022-09-02
[PF-1932] Add dependency locking to repo terra-cli

dexamundsen:
Add dependency locking to repo terra-cli, keeping it consistent with other repositories in terra
No changes to functionalities or library versions


yuhuyoyo (approved):
thanks!

ginaygoogle (approved):

PR #316, opened 2022-09-08
For cloud context creation, retry for 3 mins instead of 2 mins

melissachang:
We retry for 2 mins. For one of the flakes last night, cloud context creation took 2 mins 28 sec. So increase to 3 mins.

Also:

- Move constants inline, so I don't have to scroll to top of file to see values
- We were retrying clone workspace for 1 hr, which seems excessive. Changed to 5 min


yuhuyoyo (approved):
just a sidenote that we should start implementing random time backoff. https://cloud.google.com/storage/docs/retry-strategy

zloery (approved):

PR #318, opened 2022-09-13
[PF-1998] Check that 'test worker' property is present before using.

jmczerk:
The `cleanupTestUserWorkspaces` uses the CLI Test Harness to call commands outside of a test context.  This leads to an NPE when a NULL String is passed to the `Path.resolve()` method.

### DIFF
--- a/src/main/java/bio/terra/cli/businessobject/Context.java
+++ b/src/main/java/bio/terra/cli/businessobject/Context.java
@@ -125,7 +125,10 @@ public class Context {
     // parallel without clobbering context across runners.
     String isTest = System.getProperty(CommandRunner.IS_TEST);
     if (isTest != null && isTest.equals("true")) {
-      contextPath = contextPath.resolve(System.getProperty("org.gradle.test.worker"));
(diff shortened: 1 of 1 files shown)


zloery (approved):
nice find, thanks for this!

src/main/java/bio/terra/cli/businessobject/Context.java
```diff
@@ -125,7 +125,10 @@ public static Path getContextDir() {
     // parallel without clobbering context across runners.
     String isTest = System.getProperty(CommandRunner.IS_TEST);
     if (isTest != null && isTest.equals("true")) {
-      contextPath = contextPath.resolve(System.getProperty("org.gradle.test.worker"));
+      String testWorker = System.getProperty("org.gradle.test.worker");
+      if (testWorker != null) {
```
melissachang, line 129: // cleanupTestUserWorkspaces uses CLI Test Harness to call commands outside of test context. IS_TEST is true and testWorker is null.

melissachang (approved):

PR #319, opened 2022-09-15
[PF-1945] terra auth policy based on the server. Broad servers use broad policy, others use broad policy.

ginaygoogle:
In the ticket [PF-1945](https://broadworkbench.atlassian.net/browse/PF-1945). We have new policy logic based on the which server it is.

- If on the broad server, the auth page policy would be broad policy
- If on the verily server, the auth page policy would be verily policy
- If on other personal server, the auth page policy would be default broad policy

The manually tests are showing below: 
**Set the server to broad server** 
```
ginay-macbookpro:terra-cli ginay$ terra server set --name=broad-dev
Switching the server will clear the current login credentials and workspace. Are you sure you want to proceed (y/N)? y
Terra server is set to broad-dev (CHANGED).
ginay-macbookpro:terra-cli ginay$ terra auth login
Please open the following address in your browser:
  https://accounts.google.com/o/oauth2/auth?access_type=offline&approval_prompt=force&client_id=436986732241-nvjfkarirkhmcq7m3r1uemdau518uvdo.apps.googleusercontent.com&redirect_uri=http://localhost:56166/Callback&response_type=code&scope=openid%20email%20profile
Attempting to open that address in the default browser now...
Login successful: ginay@google.com
```
Auth page shows:
Screen Shot 2022-09-15 at 2 28 22 AM

**Set the server to verily server** 
```
ginay-macbookpro:terra-cli ginay$ terra server set --name=verily
Switching the server will clear the current login credentials and workspace. Are you sure you want to proceed (y/N)? y
Terra server is set to verily (CHANGED).
ginay-macbookpro:terra-cli ginay$ terra auth login
Please open the following address in your browser:
  https://accounts.google.com/o/oauth2/auth?access_type=offline&approval_prompt=force&client_id=590500654757-hdnr1eh33s0gn4hg050cs4gp8cdddup4.apps.googleusercontent.com&redirect_uri=http://localhost:56392/Callback&response_type=code&scope=openid%20email%20profile
Attempting to open that address in the default browser now...
Login successful: ginay@google.com
```
Auth page shows:
Screen Shot 2022-09-15 at 2 27 39 AM

**Code Explain**

- Adding the new secret json file, and making the name clear, so I make the name with "broad" and "verily"
- Considering the auth check can happen at anytime, not only at the beginning of login, also user can switch to another server on the fly, I created a helper function to exchange/update the client secret. 
- We do not have the unit test or integration test in the code change, only do the manual test. The reviewers should also pull the branches and test it.


src/main/java/bio/terra/cli/service/GoogleOauth.java
```diff
@@ -53,8 +54,8 @@ public final class GoogleOauth {
   private static final Logger logger = LoggerFactory.getLogger(GoogleOauth.class);
   // google OAuth client secret file
   // (https://developers.google.com/adwords/api/docs/guides/authentication#create_a_client_id_and_client_secret)
-  private static final String CLIENT_SECRET_FILENAME = "client_secret.json";
-  private static final GoogleClientSecrets clientSecrets = readClientSecrets();
+  private static String CLIENT_SECRET_FILENAME = "broad_secret.json";
+  private static GoogleClientSecrets clientSecrets = readClientSecrets();
```
yuhuyoyo, line 58: you can just remove these two fields

src/main/java/bio/terra/cli/service/GoogleOauth.java
```diff
@@ -371,6 +379,16 @@ public void onTokenErrorResponse(Credential credential, TokenErrorResponse token
     }
   }
 
+  /** Get the client secrets by server. */
+  private static GoogleClientSecrets UpdateClientSecret() {
```
yuhuyoyo, line 383: just implement this in getClientSecret. have readClientSecrets takes in clientSecretFileName

src/main/java/bio/terra/cli/service/GoogleOauth.java
```diff
@@ -371,6 +379,16 @@ public void onTokenErrorResponse(Credential credential, TokenErrorResponse token
...
+    if (Context.getServer().getName().startsWith("broad")) {
```
cbookg, line 384: assuming the server name seems a bit brittle.  how about instead of using the server name, we add a property to the server config that indicates which client config to use?

alternatively (less good IMO) would be to do something like:
if (startsWith("broad")) {
  // use broad config
} else if (startsWith("verily")) {
 // use verily config
} else {
  fail
}
so someone adding a new server pattern (say microsoft-devel) wouldn't fallback to verily config?

ginaygoogle: Yes. Thanks for the comments.  I also considered whether we will have other servers in the future. But supposed it should followed the verily policy due to our ownership? 

I am ok with both ideas. From the Code Sustainability, add a property like "serverType" in [businessobject/server](https://github.com/DataBiosphere/terra-cli/blob/main/src/main/java/bio/terra/cli/businessobject/Server.java) will be better, but if we only supports 2 kinds of servers even in the future, we can use the if statement.

yuhuyoyo: I like Chris's first idea. You can add another fields in these server json file  here https://github.com/DataBiosphere/terra-cli/tree/main/src/main/resources/servers. For all the personal server, let's just use broad config.

src/main/resources/servers/verily.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "verily",
+  "clientFile": "verily_secret.json",
```
yuhuyoyo, line 3: clientCredentialsFile

yuhuyoyo (approved):

cbookg (approved):

PR #323, opened 2022-09-20
Add example group usage to README

wnojopra:

PR #326, opened 2022-09-26
Add example group usage to README

wnojopra:
I had always assumed Terra groups worked this way, but didn't see any documentation in the CLI to confirm/deny. Fortunately we were able to test and confirmed that it does indeed work this way.

Created this because tests weren't running in #323

### DIFF
--- a/README.md
+++ b/README.md
@@ -531,6 +531,8 @@ not `mygroup@mydomain.com`:
 > terra group list-users --name=mygroup
 ```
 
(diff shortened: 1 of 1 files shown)


melissachang:
Force merging. ([Previous PR](https://github.com/DataBiosphere/terra-cli/pull/323) had 2 approvals.) Dex will fix things so this doesn't happen in the future.

PR #327, opened 2022-09-26
[PF-2049] terra-cli: Remove paths ignore from Github workflows

dexamundsen:
Issue - If changeset only has .md files (readme.md for example) the PR check (github workflow) does not report the status as success  skipped, thus leading to a PR that can't be merged

As a workaround remove this config for now
Further work will be carried out in https://broadworkbench.atlassian.net/browse/PF-2048

### DIFF
--- a/.github/workflows/tests-on-pr-push-and-merge.yml
+++ b/.github/workflows/tests-on-pr-push-and-merge.yml
@@ -3,10 +3,8 @@ on:
   workflow_dispatch: { }
   push:
     branches: [ main ]
-    paths-ignore: [ '**.md' ]
   pull_request:
     branches: [ '**' ]
-    paths-ignore: [ '**.md' ]
 
 jobs:
   lint-and-static-analysis:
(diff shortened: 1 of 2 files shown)


.github/workflows/tests-on-pr-push-and-merge.yml
```diff
@@ -3,10 +3,8 @@ on:
   workflow_dispatch: { }
   push:
     branches: [ main ]
-    paths-ignore: [ '**.md' ]
   pull_request:
     branches: [ '**' ]
-    paths-ignore: [ '**.md' ]
```
melissachang, line 9: Thanks for fixing!

FYI technically you only need to remove for `pull_request`, not `push`. See https://github.com/DataBiosphere/terra-workspace-manager/blob/main/.github/workflows/test.yml#L11

Let's future-proof this. Add comment:

```
# Branch settings require status checks before merging, so don't add paths-ignore.
```

melissachang (approved):

.github/workflows/tests-on-pr-push-and-merge.yml
dexamundsen: Done, thanks

wnojopra (approved):
Thanks!

PR #332, opened 2022-09-29
PF-2060: disable notebook test temporarily

yuhuyoyo:
notebook on create get stuck in provisioning. it's not a CLI bug. Disabling the test temporarily until the root issue is resolved.


melissachang (approved):
Yikes

zloery (approved):
SGTM to unblock PRs, though we should definitely remember to remove this once resolved

PR #337, opened 2022-10-06
[PF-963] Fix cleanup test user action

zloery:
The cleanup user action is broken in the Verily deployment - for some reason this command usage doesn't work as expected. This is the only command where an argument with a value is provided without a `=` character, though I don't understand why this is an issue in Verily but not the Broad deployment.


yuhuyoyo (approved):

yuhuyoyo:
actually.. `terra server set --name verily-devel` this works fine for me

zloery:
Yeah, it also works when I run it (and seems to work for the action here too). That line apparently fails when run through `TestCommand.runCommandExpectSuccess` in Verily though, e.g. https://github.com/verily-src/terra-tool-cli/actions/runs/3199083952

zloery:
Actually testing this more it seems like there's something else going on, I'm not sure this fix will work

PR #338, opened 2022-10-11
[PF-1931] Remove gradle enterprise plugins

zloery:
This does not remove `maxParallelForks` which is a non-enterprise feature, so CI tests will continue to run in parallel on the 2 cores available to GHA runners.

### DIFF
--- a/.github/workflows/tests-nightly.yml
+++ b/.github/workflows/tests-nightly.yml
@@ -67,21 +67,21 @@ jobs:
       run: |
         echo "Running unit tests for server: ${{ matrix.testServer }}"
         mkdir -p ~/logs-unit
-        ./gradlew runTestsWithTag -PtestTag=unit --scan -Pserver=${{ matrix.testServer }} -PcontextDir=$HOME/logs-unit
+        ./gradlew runTestsWithTag -PtestTag=unit -Pserver=${{ matrix.testServer }} -PcontextDir=$HOME/logs-unit
     - name: Run integration tests against source code
       id: run_integration_tests_against_source_code
       if: always()
       run: |
         echo "Running integration tests against source code for server: ${{ matrix.testServer }}"
         mkdir -p ~/logs-integration-source
-        ./gradlew runTestsWithTag -PtestTag=integration --scan -Pserver=${{ matrix.testServer }} -PcontextDir=$HOME/logs-integration-source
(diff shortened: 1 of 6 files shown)


.github/workflows/tests-nightly.yml
```diff
@@ -67,21 +67,21 @@ jobs:
       run: |
         echo "Running unit tests for server: ${{ matrix.testServer }}"
         mkdir -p ~/logs-unit
-        ./gradlew runTestsWithTag -PtestTag=unit --scan -Pserver=${{ matrix.testServer }} -PcontextDir=$HOME/logs-unit
+        ./gradlew runTestsWithTag -PtestTag=unit -Pserver=${{ matrix.testServer }} -PcontextDir=$HOME/logs-unit
```
yuhuyoyo, line 70: sad that the non-enterprise version doesn't have scan anymore

yuhuyoyo (approved):
what's the context? is this because the broad doesn't want gradle enterprise?

.github/workflows/tests-nightly.yml
melissachang: Can we keep the --scan? (Same for rest of PR)

build.gradle
```diff
@@ -5,7 +5,6 @@ plugins {
 
     id 'com.diffplug.spotless' version '6.10.0'
     id 'com.github.spotbugs' version '5.0.12'
-    id 'org.gradle.test-retry' version '1.4.1'
```
melissachang, line 8: I don't think test-retry requires GE, can we keep that? That was very useful.

gradle/test.gradle
```diff
@@ -103,21 +100,6 @@ task runTestsWithTag(type: Test) {
     // run the install-for-testing.sh script before any tests
 
     outputs.upToDateWhen { false } // force tests to always be re-run
-
-    retry {
```
melissachang, line 107: Keep?

melissachang (approved):

.github/workflows/tests-nightly.yml
zloery: good point, we can continue using the public scan functionality!

build.gradle
zloery: looks like you're right, adding that back in

zloery:
> what's the context? is this because the broad doesn't want gradle enterprise?

We decided not to go through with GE after the trial. I'm writing up a quick summary doc, I'll share that out with the team

PR #339, opened 2022-10-11
[PF-1930] Add COPY_REFERENCE as option for cloning controlled resource

melissachang:


yuhuyoyo (approved):

src/main/java/bio/terra/cli/command/shared/options/CloningInstructionsForUpdate.java
```diff
@@ -15,9 +15,9 @@ public class CloningInstructionsForUpdate {
   @CommandLine.Option(
       names = "--new-cloning",
       // For what is allowed, see:
-      // https://github.com/DataBiosphere/terra-workspace-manager/blob/main/service/src/main/java/bio/terra/workspace/service/resource/ResourceValidationUtils.java#L448
+      // https://github.com/DataBiosphere/terra-workspace-manager/blob/main/service/src/main/java/bio/terra/workspace/service/resource/ResourceValidationUtils.java#L478
```
yuhuyoyo, line 18: use perm link? https://github.com/DataBiosphere/terra-workspace-manager/blob/792780ab6d1bf38ee9e4b811811952135974c7f2/service/src/main/java/bio/terra/workspace/service/resource/ResourceValidationUtils.java#L478

yuhuyoyo (approved):

src/main/java/bio/terra/cli/command/shared/options/CloningInstructionsForUpdate.java
melissachang: Sometimes I use perm link. In this case, I think it's better to link to latest commit. If logic changes, we want CLI developer to see the latest logic.

zloery (approved):

PR #341, opened 2022-10-12
[PF-2101] Check java version in CLI download-install.sh

dexamundsen:
Problem statement - Terra CLI needs java 17 installed. There is no check during cli installation, but execution fails
Fix - add a check for java version during installation

with this change, install will fail if Java is set to <17. 
terra command would still fail if the java version is changed after install. Keeping this use case out of scope for now

Testing  - install fails with java set to version 11

```
clicheck-test >> ls
download-install.sh
clicheck-test >> jenv versions
...
* 11.0 (set by /Users/dexamundsen/clicheck-test/.java-version)
...
clicheck-test >> java --version
openjdk 11.0.16.1 2022-07-19 LTS
OpenJDK Runtime Environment Zulu11.58+23-CA (build 11.0.16.1+1-LTS)
OpenJDK 64-Bit Server VM Zulu11.58+23-CA (build 11.0.16.1+1-LTS, mixed mode)
clicheck-test >> ./download-install.sh 
--  Checking if installed Java version is 17 or higher
Java version set to 11, exiting...
```

Testing  - install succeeds with java set to version 17
```
clicheck-test >> jenv local 17.0.3
clicheck-test >> jenv versions
...
* 17.0.3 (set by /Users/dexamundsen/.jenv/version)
...
clicheck-test >> java --version
openjdk 17.0.3 2022-04-19
OpenJDK Runtime Environment Temurin-17.0.3+7 (build 17.0.3+7)
OpenJDK 64-Bit Server VM Temurin-17.0.3+7 (build 17.0.3+7, mixed mode)
clicheck-test >> ./download-install.sh 
....
--   Install complete
....
clicheck-test >> 
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -16,7 +16,7 @@ plugins {
 
 // TODO(PF-1981) - Move version to settings.gradle
 // The tools/publish-release.sh script depends on this version string being of the format "version = '1.2.3'"
-version = '0.258.0'
+version = '0.259.0'
 group = 'terra-cli'
 
 // If true, search local repository (~/.m2/repository/) first for dependencies.
--- a/build.gradle
+++ b/build.gradle
@@ -16,7 +16,7 @@ plugins {
 
 // TODO(PF-1981) - Move version to settings.gradle
 // The tools/publish-release.sh script depends on this version string being of the format "version = '1.2.3'"
-version = '0.259.0'
+version = '0.260.0'
 group = 'terra-cli'
 
 // If true, search local repository (~/.m2/repository/) first for dependencies.
--- a/src/test/resources/testscripts/DeleteWorkspace.sh
+++ b/src/test/resources/testscripts/DeleteWorkspace.sh
@@ -6,8 +6,7 @@ terra status
 terra auth status
 
 isLoggedIn=$(terra auth status --format=json | jq .loggedIn)
-if [[ "true" = "$isLoggedIn" ]]
-then
+if [[ "true" = "$isLoggedIn" ]]; then
   currentUser=$(terra auth status --format=json | jq .userEmail)
   echo "User $currentUser is logged in. Deleting the current workspace."
   terra workspace delete --quiet
--- a/src/test/resources/testscripts/DeleteWorkspace.sh
+++ b/src/test/resources/testscripts/DeleteWorkspace.sh
@@ -6,7 +6,7 @@ terra status
 terra auth status
 
 isLoggedIn=$(terra auth status --format=json | jq .loggedIn)
-if [[ "true" = "$isLoggedIn" ]]; then
+if [[ "true" == "$isLoggedIn" ]]; then
   currentUser=$(terra auth status --format=json | jq .userEmail)
   echo "User $currentUser is logged in. Deleting the current workspace."
   terra workspace delete --quiet
--- a/tools/build-docker.sh
+++ b/tools/build-docker.sh
@@ -9,21 +9,21 @@ set -e
 ##        ./tools/build-docker.sh test123 terracli/branchA   --> builds local Docker image terracli/branchA:test123
 
 ## The script assumes that it is being run from the top-level directory "terra-cli/".
-if [ $(basename $PWD) != 'terra-cli' ]; then
+if [[ $(basename $PWD) != 'terra-cli' ]]; then
   >&2 echo "ERROR: Script must be run from top-level directory 'terra-cli/'"
   exit 1
 fi
 
 # generate a tag from the commit hash if no tag was provided
 localImageTag=$1
-if [ -z "$localImageTag" ]; then
+if [[ -z "$localImageTag" ]]; then
   echo "Generating an image tag from the Git commit hash"
   localImageTag=$(git rev-parse --short HEAD)
 fi
 
 # set the local image name if no name was provided
 localImageName=$2
-if [ -z "$localImageName" ]; then
+if [[ -z "$localImageName" ]]; then
     localImageName="terra-cli/local"
 fi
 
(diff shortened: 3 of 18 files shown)


tools/download-install.sh
```diff
@@ -6,6 +6,22 @@ set -e
...
 
+readonly REQ_JAVA_VERSION=17
+echo "--  Checking if installed Java version is ${REQ_JAVA_VERSION} or higher"
+if [[ -n "$(which java)" ]];
+then
+  # Get the current major version of Java: "11.0.12" => "11"
+  readonly CUR_JAVA_VERSION="$(java -version 2>&1 | awk -F\" '{ split($2,a,"."); print a[1]}')"
+  if [[ "${CUR_JAVA_VERSION}" -lt ${REQ_JAVA_VERSION} ]];
+  then
+    echo "Java version set to ${CUR_JAVA_VERSION}, exiting..."
```
mikenev, line 17: nit: I'd note this as an error

mikenev (approved):

tools/download-install.sh
dexamundsen: Done!

jmczerk (approved):

tools/download-install.sh
```diff
@@ -6,6 +6,22 @@ set -e
 ## Usage: ./download-install.sh                            --> installs the latest version
 ##        TERRA_CLI_VERSION=0.1.0 ./download-install.sh    --> installs version 0.1.0
 
+readonly REQ_JAVA_VERSION=17
+echo "--  Checking if installed Java version is ${REQ_JAVA_VERSION} or higher"
+if [[ -n "$(which java)" ]];
+then
```
mbookman, line 12: Plase have "then" on the same line as the "if".
See go/shell-style#control-flow.

tools/download-install.sh
```diff
@@ -6,6 +6,22 @@ set -e
...
+  # Get the current major version of Java: "11.0.12" => "11"
+  readonly CUR_JAVA_VERSION="$(java -version 2>&1 | awk -F\" '{ split($2,a,"."); print a[1]}')"
+  if [[ "${CUR_JAVA_VERSION}" -lt ${REQ_JAVA_VERSION} ]];
+  then
+    echo "Error with Java version: set to ${CUR_JAVA_VERSION}, exiting..."
```
mbookman, line 17: The info message above indicates what is being checked (java version minimum), but it's a good practice to have the error as informative as possible in case the end user gets only one line of information.

```
echo "ERROR: Java version detected (${CUR_JAVA_VERSION}) is less than required (${REQ_JAVA_VERSION})
```

I also like to send errors to stderr instead of stdout, though it isn't done in this file. I'd prefer to see all of the error messages be written as:


```
>&2 echo "ERROR: Java version detected (${CUR_JAVA_VERSION}) is less than required (${REQ_JAVA_VERSION})"
```

mbookman (changes requested):

build.gradle
```diff
@@ -16,7 +16,7 @@ plugins {
 
 // TODO(PF-1981) - Move version to settings.gradle
 // The tools/publish-release.sh script depends on this version string being of the format "version = '1.2.3'"
-version = '0.258.0'
```
dexamundsen, line 19: Changes in non-shell files are due to rebase. They belong to 2 previous PRs and will not cause merge issues
since this PR already had comments, retaining it.

dexamundsen:
Changes in non-shell files are due to rebase. They belong to 2 previous PRs and will not cause merge issues
since this PR already had comments, retaining it.

tools/download-install.sh
dexamundsen: Done update all files in this repo

tools/download-install.sh
dexamundsen: done, changes in entire repo

src/test/resources/testscripts/DeleteWorkspace.sh
```diff
@@ -6,8 +6,7 @@ terra status
 terra auth status
 
 isLoggedIn=$(terra auth status --format=json | jq .loggedIn)
-if [[ "true" = "$isLoggedIn" ]]
-then
+if [[ "true" = "$isLoggedIn" ]]; then
```
mbookman, line 9: Could you also (opportunistically) replace single "=" with double "=="?

From https://google.github.io/styleguide/shellguide.html#testing-strings

_For clarity, use == for equality rather than = even though both work. The former encourages the use of [[ and the latter can be confused with an assignment._

tools/build-docker.sh
```diff
@@ -23,8 +23,7 @@ fi
 
 # set the local image name if no name was provided
 localImageName=$2
-if [ -z "$localImageName" ]
-  then
+if [ -z "$localImageName" ]; then
```
mbookman, line 26: Could you (opportunistically) replace single brace expressions with double-brace?

https://google.github.io/styleguide/shellguide.html#test----and---

> [[ … ]] is preferred over [ … ], test and /usr/bin/[.
> 
> [[ … ]] reduces errors as no pathname expansion or word splitting takes place between [[ and ]]. In addition, [[ … ]] allows for regular expression matching, while [ … ] does not.

mbookman (changes requested):
Thanks for doing the additional work on shell script consistency!

src/test/resources/testscripts/DeleteWorkspace.sh
dexamundsen: Done

tools/build-docker.sh
dexamundsen: done

mbookman (approved):
LGTM

PR #347, opened 2022-10-25
[PF-2138] Lower-case email stored in User, WorkspaceUser

melissachang:
This way, callers of User.getEmail() or WorkspaceUser.getEmail() don't have to lower-case themselves, which is error-prone.

Ran PassthroughApps.workspaceEnvVars multiple times, still not flaky.


mbookman (commented):
This looks good to me. I have a few questions.

1- Should the same be done in the constructor? While it may be lowercasing something known to be lower-case, it's not computationally expensive and asserts that email addresses are lower-case.
2- Should the same be done for the service account email?

A variant for 1 is to have a private function:

```
private void setEmail(userEmail) {
  email = userEmail.toLowerCase()
}
```

that fetchUserInfo() and the constructor would call.

mbookman (approved):
LGTM

yuhuyoyo (approved):

PR #348, opened 2022-10-25
[PF-2144] Fix `terra workspace set-property` failure

melissachang:
Before this PR:

```
~/terra-ui (mc/fix-group): terra workspace set-property
[ERROR] An unexpected error occurred in java.lang.NullPointerException: Cannot invoke "java.util.Map.entrySet()" because "propertyMap" is null
See /Users/melchang/.terra/logs/terra.log for more information
```

After this PR:

```
~/terra-cli (mc/set-property): terra workspace set-property
Missing required option: '--properties=<String=String>'
Usage: terra workspace set-property [--format=<format>] [--workspace=<id>]
                                    --properties=<String=String>[,
                                    <String=String>...]
                                    [--properties=<String=String>[,
                                    <String=String>...]]...
Set the workspace properties.
      --format=<format>   Set the format for printing command output: JSON,
                            TEXT. Defaults to the config format property.
                            Default: null
      --properties=<String=String>[,<String=String>...]
                          Workspace properties. Example:
                            --properties=key=value. For multiple properties,
                            use ",": --properties=key1=value1,key2=value2
      --workspace=<id>    Workspace id to use for this command only.
```

For `terra workspace set-property`, `--properties` is a required flag. If a required flag is missing, convention is to display help message like above. In order to display help message, `--properties` must be required. In WorkspaceProperties.java, `--properties` is optional. So I deleted `WorkspaceProperties.java` and hard-coded the flag in Create.java (--properties is optional) and SetProperty.java (--properties is required).


ginaygoogle (approved):

src/main/java/bio/terra/cli/service/WorkspaceManagerService.java
```diff
@@ -1561,6 +1561,9 @@ private <T> T handleClientExceptions(
   }
 
   public List<Property> buildProperties(Map<String, String> propertyMap) {
+    if (propertyMap == null) {
```
yuhuyoyo, line 1564: annotate nullable

yuhuyoyo: nit, since you are here, also change the method to private static

yuhuyoyo (approved):

PR #349, opened 2022-10-27
[PF-2167] Add new verily environments

dexamundsen:
Add new verily environments - mc-dev and mc-flagged

### DIFF
--- a/src/main/resources/servers/all-servers.json
+++ b/src/main/resources/servers/all-servers.json
@@ -11,6 +11,8 @@
   "verily.json",
   "verily-autopush.json",
   "verily-devel.json",
+  "verily-mc-dev.json",
+  "verily-mc-flagged.json",
   "verily-preprod.json",
   "verily-staging.json"
 ]
--- a/src/main/resources/servers/verily-mc-dev.json
+++ b/src/main/resources/servers/verily-mc-dev.json
@@ -1,0 +1,11 @@
+{
+  "name": "verily-mc-dev",
+  "clientCredentialsFile": "verily_secret.json",
+  "description": "Verily mc-dev environment",
+  "samUri": "https://terra-mc-dev-sam.api.verily.com",
+  "samInviteRequiresAdmin": true,
+  "workspaceManagerUri": "https://terra-mc-dev-wsm.api.verily.com",
+  "wsmDefaultSpendProfile": "wm-default-spend-profile",
+  "externalCredsUri": "https://terra-mc-dev-ecm.api.verily.com",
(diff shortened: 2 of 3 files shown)


yuhuyoyo:
what does mc-flagged mean?

dexamundsen:
> what does mc-flagged mean?

- mc-dev: development environment
- mc-flagged: development environment with all features blocked via a feature flag. for end users this should look the same as the regular dev / staging environment. The main goal of having this is to ensure we dont break any other existing functionality

yuhuyoyo:
what's the benefit of having another environment vs just having the dev and you can switch flags on-off?

less deployment overhead and more flexibility (e.g. having a subset of flags on)

dexamundsen:
> what's the benefit of having another environment vs just having the dev and you can switch flags on-off?
> 
> less deployment overhead and more flexibility (e.g. having a subset of flags on)

The general idea is -(from deployment design doc) 
Integration tests run against a “dev” environment that enables new features, and a “flagged off” environment that disables new functionality via feature flags/config parameters to expose regressions in existing code paths caused by our changes.

yuhuyoyo:
That is a valuable testing feature (i listed it under nice to have in the feature requirement doc but I guess I should move it up :) ) I wasn't able to find any documentation for out-of-box testing framework support for this.

WSM integration test set config for different environments like this https://github.com/DataBiosphere/terra-workspace-manager/tree/main/integration/src/main/resources/configs/integration. Say we are using LD. Can we instead create an all-flagged-on environment and all-flagged-off environment in LD and have the test points to their client ids (different per environment)?

dexamundsen:
> That is a valuable testing feature (i listed it under nice to have in the feature requirement doc but I guess I should move it up :) ) I wasn't able to find any documentation for out-of-box testing framework support for this.
> 
> WSM integration test set config for different environments like this https://github.com/DataBiosphere/terra-workspace-manager/tree/main/integration/src/main/resources/configs/integration. Say we are using LD. Can we instead create an all-flagged-on environment and an all-flagged-off environment in LD and have the test points to their client ids (different per environment)?
> 
> or just separate config files when there's no feature flagging yet

Agree on the alternate to accomplishing this. However since the MC design and implementation may have near-term changes, created this second env.  This flagged env will essentially be removed later once the concept of feature flagging is set up for all services as needed. In the meantime, this is the easiest path to getting there (we don't have feature flags or configs set up on all services in the manner we need). Happy to plan the EOL of this env once we have it set up.

zloery (approved):

PR #351, opened 2022-11-06
Switch default behavior and add a FULL_PATH_SQL option.

gjuggler:


src/main/java/bio/terra/cli/businessobject/resource/BqResolvedOptions.java
```diff
@@ -2,16 +2,24 @@
 
 /** This enum specifies the possible ways to resolve a BigQuery resource. */
 public enum BqResolvedOptions {
-  FULL_PATH, // For data table: [project id].[dataset id].[data table id if applicable]
+  FULL_PATH, // For data table: [project id]:[dataset id].[data table id if applicable]
```
melissachang, line 5: Suggest FULL_PATH -> FULL_PATH_BQ and some comment changes

```
// project:dataset.table, used for bq CLI
FULL_PATH_BQ,
// project.dataset.table, used for SQL
FULL_PATH_SQL,
```
(I left out `if applicable` for brevity..)

src/test/java/harness/utils/ExternalBQDatasets.java
```diff
@@ -217,14 +217,25 @@ public static BigQuery getBQClient(GoogleCredentials credentials) {
         .getService();
   }
 
-  /** Gets cloud identifier for Dataset in full-path: [project id].[dataset id] */
+  /** Gets cloud identifier for Dataset in full-path: [project id]:[dataset id] */
   public static String getDatasetFullPath(String projectId, String datasetId) {
```
melissachang, line 221: getDatasetFullPathBq ?

melissachang (approved):
I'm not sure which should be default. It could be that a user rarely runs `bq`, and often runs SQL queries (eg in a notebook) against data collection BQ resources. If you want, I'm happy to shoot an email to solutions. The email could kill 2 birds with 1 stone: ask for their preference, and possibly notify them of breaking change.

If we keep bq as default, I can update UI to match. (In the future, this logic should be in FrontendServer.)

gjuggler:
> I'm not sure which should be default. It could be that a user rarely runs `bq`, and often runs SQL queries (eg in a notebook) against data collection BQ resources. If you want, I'm happy to shoot an email to solutions. The email could kill 2 birds with 1 stone: ask for their preference, and possibly notify them of breaking change.
> 
> If we keep bq as default, I can update UI to match. (In the future, this logic should be in FrontendServer.)

👍  I emailed the solutions team, and will come back to this PR once we have a clearer signal on which default to prefer.

PR #358, opened 2022-11-23
[PF-2174] check version is out of date for the commands talk to server

ginaygoogle:
This PR comes from the bug ticket [PF-2174](https://broadworkbench.atlassian.net/browse/PF-2174)

Before the ticket, we do the version check every 30 minutes and do not check the version for every command. 
Because we do not support the version out of date. We should throw error to user when they request the server but the version is old.

The ticket scope:
- create the subclass which extend from the parent class `BaseCommand`. and can specify part of the command use the subclass. and rest of them still use the parent class.
- only affect the resource commands and workspace commands because they talk to the server.
- remove the existing expire time check in `src/main/java/bio/terra/cli/app/utils/VersionCheckUtils.java` 

Manually test:

the version is up to date:
```
ginay-macbookpro:terra-cli ginay$ terra workspace list
   ID                                        NAME                            GOOGLE PROJECT                  DESCRIPTION                             
   propertytest                              (unset)                         terra-vdevel-halcyon-onion-400  (unset)                                 
   junhtest                                  (unset)                         terra-vdevel-smart-mango-3426   (unset)                                 
   a-6bc4a502-8600-4449-b224-0104d80e444b    (unset)                         terra-vdevel-speedy-lemon-8435  (unset)                                 
   clonetestname                             (unset)                         terra-vdevel-cold-lime-8223     (unset)                                 
   newidfortestnameclone                     (unset)                         terra-vdevel-happy-coffee-83    (unset)                                 
   newidupdated                              (unset)                         terra-vdevel-genial-grape-9381  (unset)      
```

manually set the version to old version:
```
ginay-macbookpro:terra-cli ginay$ terra workspace list
Warning: Version 0.27.0 of the CLI has expired. Functionality may not work as expected. To install the latest version: curl -L https://github.com/DataBiosphere/terra-cli/releases/latest/download/download-install.sh | bash ./terra
If you have added the CLI to your $PATH, this step will need to be repeated after the installation is complete.
```
It does not affect on the command which not request the server
```
ginay-macbookpro:terra-cli ginay$ terra config list
OPTION                VALUE                                          DESCRIPTION                                                 
app-launch            DOCKER_CONTAINER                               app launch mode                                             
browser               AUTO                                           browser launch for login                                    
image                 gcr.io/terra-cli-dev/terra-cli/0.2.0:stable    docker image id                                             
resource-limit        1000                                           max number of resources to allow per workspace              
console-logging       OFF                                            logging level for printing directly to the terminal         
file-logging          INFO                                           logging level for writing to files/Users/ginay/.terra/logs  
server                verily-devel                                   (unset)                                                     
workspace             testid                                         (unset)                                                     
format                TEXT                                           output format       
```

### DIFF
--- a/src/main/java/bio/terra/cli/app/utils/VersionCheckUtils.java
+++ b/src/main/java/bio/terra/cli/app/utils/VersionCheckUtils.java
@@ -5,55 +5,32 @@ import bio.terra.cli.businessobject.VersionCheck;
 import bio.terra.cli.service.WorkspaceManagerService;
 import bio.terra.workspace.model.SystemVersion;
 import java.lang.module.ModuleDescriptor.Version;
-import java.time.Duration;
 import java.time.OffsetDateTime;
-import java.util.Optional;
 import javax.annotation.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class VersionCheckUtils {
   private static final Logger logger = LoggerFactory.getLogger(VersionCheckUtils.class);
-  private static final Duration VERSION_CHECK_INTERVAL = Duration.ofMinutes(30);
 
   /** Query Workspace Manager for the oldest supported */
   public static boolean isObsolete() {
-    if (checkIntervalElapsed()) {
-      // update last checked time in the context file
-      VersionCheck updatedVersionCheck = new VersionCheck(OffsetDateTime.now());
-      Context.setVersionCheck(updatedVersionCheck);
+    // update last checked time in the context file
+    VersionCheck updatedVersionCheck = new VersionCheck(OffsetDateTime.now());
+    Context.setVersionCheck(updatedVersionCheck);
 
-      // The oldest supported version is exposed on the main WSM /version endpoint
-      SystemVersion wsmVersion =
-          WorkspaceManagerService.unauthenticated(Context.getServer()).getVersion();
-      String oldestSupportedVersion = wsmVersion.getOldestSupportedCliVersion();
-      String currentCliVersion = bio.terra.cli.utils.Version.getVersion();
-      boolean result = isOlder(currentCliVersion, oldestSupportedVersion);
-      logger.debug(
-          "Current CLI version {} is {} than the oldest supported version {}",
-          currentCliVersion,
-          result ? "older" : "newer",
-          oldestSupportedVersion);
-      return result;
-    } else {
-      return false;
-    }
-  }
-
-  /**
-   * We don't want to hit the version endpoint on the server with every command invocation, so check
-   * if a certain amount of time has passed since the last time we checked (or the last time is
-   * null/never).
-   *
-   * @return true if we should do the version check again
-   */
-  public static boolean checkIntervalElapsed() {
-    Optional<OffsetDateTime> lastCheckTime =
-        Context.getVersionCheck().map(VersionCheck::getLastVersionCheckTime);
-    return (lastCheckTime.isEmpty()
-        || Duration.between(lastCheckTime.get(), OffsetDateTime.now())
-                .compareTo(VERSION_CHECK_INTERVAL)
-            > 0);
+    // The oldest supported version is exposed on the main WSM /version endpoint
+    SystemVersion wsmVersion =
+        WorkspaceManagerService.unauthenticated(Context.getServer()).getVersion();
+    String oldestSupportedVersion = wsmVersion.getOldestSupportedCliVersion();
+    String currentCliVersion = bio.terra.cli.utils.Version.getVersion();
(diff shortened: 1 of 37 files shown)


src/main/java/bio/terra/cli/app/utils/VersionCheckUtils.java
```diff
@@ -5,55 +5,32 @@
...
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class VersionCheckUtils {
   private static final Logger logger = LoggerFactory.getLogger(VersionCheckUtils.class);
-  private static final Duration VERSION_CHECK_INTERVAL = Duration.ofMinutes(30);
 
   /** Query Workspace Manager for the oldest supported */
   public static boolean isObsolete() {
-    if (checkIntervalElapsed()) {
```
ginaygoogle, line 21: Based on the change, we should check the version once the command request the server, so the `checkIntervalElapsed` is not need. If we keep this, when the version is out of date. it will show nothing until 30 minutes later. Like
```
ginay-macbookpro:terra-cli ginay$ terra workspace list
ginay-macbookpro:terra-cli ginay$ 
```
This is also not clear for the reason why this is nothing print out, Maybe because of the server does not have workspace, or some error hit.

src/main/java/bio/terra/cli/command/shared/WsmBaseCommand.java
```diff
@@ -0,0 +1,52 @@
...
+import java.io.PrintStream;
+import org.slf4j.LoggerFactory;
+
+public abstract class WsmBaseCommand extends BaseCommand {
+  private static final org.slf4j.Logger logger = LoggerFactory.getLogger(WsmBaseCommand.class);
+  protected static PrintStream OUT;
+  protected static PrintStream ERR;
+
+  @Override
+  public Integer call() {
```
melissachang, line 18: Can 18-51 be replaced with

```
  @Override
  public Integer call() {
    // Check if this version of the CLI is out of date
    if (VersionCheckUtils.isObsolete()) {
      ERR.printf(
          "Warning: Version %s of the CLI has expired. Functionality may not work as expected. To install the latest version: curl -L https://github.com/DataBiosphere/terra-cli/releases/latest/download/download-install.sh | bash ./terra\n"
              + "If you have added the CLI to your $PATH, this step will need to be repeated after the installation is complete.%n",
          bio.terra.cli.utils.Version.getVersion());
      return 0;
    }

    return super.call();
  }
```

src/main/java/bio/terra/cli/command/shared/WsmBaseCommand.java
```diff
@@ -0,0 +1,52 @@
...
+import bio.terra.cli.app.utils.VersionCheckUtils;
+import bio.terra.cli.businessobject.Context;
+import bio.terra.cli.businessobject.User;
+import bio.terra.cli.command.Main;
+import bio.terra.cli.utils.Logger;
+import bio.terra.cli.utils.UserIO;
+import java.io.PrintStream;
+import org.slf4j.LoggerFactory;
+
+public abstract class WsmBaseCommand extends BaseCommand {
```
melissachang, line 12:
```
/**
 * This class prints a warning if CLI version is too old for WSM version.
 * 
 * Commands which call WSM should extend this instead of BaseCommand. (This way, if WSM happens to be down, commands that don't call WSM will still succeed.)
 */
```
In general, if a class isn't obvious, there should be a comment. :)

src/main/java/bio/terra/cli/command/shared/WsmBaseCommand.java
ginaygoogle: Wow. Thank you, Melissa. Just a heads-up, It also need necessary initialization before calling VersionCheckUtils, so I add L21-25. 
I also keep the L26-27 with the L13. How do you think it? if we do not need this logger, I can remove these lines.

melissachang (approved):

src/main/java/bio/terra/cli/app/utils/VersionCheckUtils.java
```diff
@@ -5,55 +5,32 @@
...
-      // update last checked time in the context file
-      VersionCheck updatedVersionCheck = new VersionCheck(OffsetDateTime.now());
-      Context.setVersionCheck(updatedVersionCheck);
+    // update last checked time in the context file
+    VersionCheck updatedVersionCheck = new VersionCheck(OffsetDateTime.now());
+    Context.setVersionCheck(updatedVersionCheck);
```
yuhuyoyo, line 20: i don't think we need to set this time anymore since we are not checking the version based on the interval elapse.

yuhuyoyo (approved):
thanks!

PR #359, opened 2022-11-24
[PF-2257] Allow specifying alternate spend profiles

iliadsh:
`spend enable`, `disable`, and `list-users` accept an additional (optional) `--profile` and `--enable-spend` in `user invite` accepts an (optional) parameter for the profile as well.

Should be used with caution since the CLI and UI do **not** yet store and read the given spend profile on invite -- the WSM default is still assumed.


src/main/java/bio/terra/cli/command/spend/Enable.java
```diff
@@ -17,15 +19,20 @@ public class Enable extends BaseCommand {
 
   /** Print this command's output in text format. */
   private static void printText(UFSpendProfileUser returnValue) {
-    OUT.println("User enabled on the default spend profile.");
+    OUT.println("User enabled on the spend profile.");
```
cbookg, line 22: print the specific profile here (like with disable)?

src/main/java/bio/terra/cli/command/user/Invite.java
```diff
@@ -17,20 +19,27 @@ public class Invite extends BaseCommand {
...
+          spendProfile.get().isEmpty()
+              ? Context.getServer().getWsmDefaultSpendProfile()
+              : spendProfile.get();
+
       SpendProfileUser spendProfileUser =
-          SpendProfileUser.enable(email, SpendProfileManagerService.SpendProfilePolicy.USER);
-      OUT.println("User enabled on the default spend profile.");
+          SpendProfileUser.enable(
+              email, SpendProfileManagerService.SpendProfilePolicy.USER, profile);
+      OUT.println("User enabled on the spend profile.");
```
cbookg, line 42: print the profile?

src/main/java/bio/terra/cli/command/spend/Disable.java
```diff
@@ -12,14 +14,20 @@
 public class Disable extends BaseCommand {
   @CommandLine.Mixin bio.terra.cli.command.shared.options.SpendProfileUser spendProfileUserOption;
 
-  /** Remove access to the WSM default spend profile for the given email. */
+  /** Remove access to a spend profile for the given email. */
   @Override
```
cbookg, line 18: i assume tests exists for these modified commands.  we should add tests to cover the value being provided.

src/main/java/bio/terra/cli/command/spend/Enable.java
iliadsh: It's being printed as part of the UFSpendProfileUser.print(). For ex:
```
> terra spend enable --email=ishaghag@google.com --policy=user --profile=wm-alt-spend-profile

User enabled on the spend profile.
ishaghag@google.com: wm-alt-spend-profile(OWNER, USER)
```

src/main/java/bio/terra/cli/command/spend/Disable.java
iliadsh: Yeah true, wasn't sure how to make the test generic for both broad and verily servers though...

we could add a value to the test configs for an alt spend profile in the env?

cbookg: Can this not be tested without the profile existing on a real server?

iliadsh: my understanding is that all the tests make calls to real servers, so I'm unsure how to meaningfully test this without the profile existing

src/main/java/bio/terra/cli/command/spend/Enable.java
cbookg: sg

src/main/java/bio/terra/cli/command/user/Invite.java
```diff
@@ -17,20 +19,27 @@ public class Invite extends BaseCommand {
 
   @CommandLine.Option(
       names = "--enable-spend",
-      defaultValue = "false",
-      description = "Also enable the email as a user on the default spend profile.")
-  private boolean enableSpend;
+      arity = "0..1",
```
melissachang, line 22: Why did this (lines 20-25) change? I find old code easier to read

src/main/java/bio/terra/cli/command/spend/Disable.java
cbookg: Having a test run potentially leak unique spend profiles isn't ideal, nor is deleting a single 'test' spend profile that would never allow tests to run in parallel.

Seems like using a static 'alt' profile in all envs is the least-of-all-evils?

melissachang: How is leaking spend profiles different from leaking workspaces, resources, SAM groups, etc?

If you want a static alt profile, we should document setting it up https://github.com/DataBiosphere/terra-cli/blob/main/ADMIN.md#setup-spend-profile . 

Note that CreateProfile, DeleteProfile should be updated as part of this ticket, even if not part of this PR, and not tested. (Right?)

melissachang: My 2 cents is that it's fine to leak things in non-prod environments. In general, I think it's fine for data in non-prod environments to be "bad". Otherwise, we wouldn't be able to test creating/deleting X.

Your proposal SG. My vote is not just hard-code "test-spend-profile", not add a server setting (line in server .json files).

cbookg: Did you mean "My vote is *to* hard-code..."?

melissachang: Yes :)

src/main/java/bio/terra/cli/command/user/Invite.java
iliadsh: Wanted to keep the argument structure the same, so `--enable-spend` can still be used as a flag OR you can specify a profile to go with it like `--enable-spend=test-spend`. 

Is there anything that jumps out as less readable I should address?

melissachang: I prefer to leave --enable-spend as a boolean; it makes the code easier to understand. If --enable-spend is set, then --spend-profile will be used. (If --spend-profile isn't set on command-line, default of wm-default-spend-profile will be used.)

(@cbookg another thing we could do is delete --enable-spend option and always add to spend profile. In practice, we always set --enable-spend.)

src/main/java/bio/terra/cli/command/shared/options/SpendProfileUser.java
```diff
@@ -20,4 +20,7 @@ public class SpendProfileUser {
       required = true,
       description = "Spend policy: ${COMPLETION-CANDIDATES}.")
   public SpendProfileManagerService.SpendProfilePolicy policy;
+
+  @CommandLine.Option(names = "--profile", description = "Name of the spend profile.")
+  public String spendProfile;
```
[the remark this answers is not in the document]
iliadsh: sounds good!

src/main/java/bio/terra/cli/command/user/Invite.java
iliadsh: fair enough. kept the --enable-spend flag for now but agree it should just be the default so can remove if needed

ADMIN.md
```diff
@@ -34,16 +31,15 @@ there will be a dedicated Spend Profile Manager service that will allow users to
...
 access to users.
 
-Until that new service is build, Workspace Manager recognizes a single spend
-profile per deployment. This single spend profile corresponds to a SAM resource.
-The name of this resource is specified in the Workspace Manager deployment
-configuration (e.g. Helm charts) but needs to be setup and managed manually in
-SAM (i.e. there are no WSM endpoints for managing spend profiles, you have to
-talk to SAM directly). The CLI provides convenience commands for this setup and
-management.
+Until that new service is build, Workspace Manager must hardcode spend profiles
```
melissachang, line 34: Change 34-39:

Until that new service is built, setting up a new spend profile entails:

- Adding WSM configuration (helm charts) which associates spend profile name with a billing account ID
- Creating the SAM spend profile resource with `terra spend create-profile`

src/main/java/bio/terra/cli/command/spend/CreateProfile.java
```diff
@@ -2,17 +2,23 @@
...
 
 /** This class corresponds to the third-level "terra spend create-profile" command. */
-@Command(
-    name = "create-profile",
-    description = "Create the Workspace Manager default spend profile.")
+@Command(name = "create-profile", description = "Create a spend profile.")
 public class CreateProfile extends BaseCommand {
-  /** Create the WSM default spend profile. */
+
+  @CommandLine.Option(
```
melissachang, line 12: When an option is needed in more than one command, we usually create a shared option. Let's

- Move this option to shared/options/SpendProfile.java
- This this option out of shared/options/SpendfProfileUser.java
- Use shared/options/SpendProfile.java everywhere

melissachang (commented):
I thought this PR was ready for review so I took another look. I don't see tests, so maybe it wasn't ready... :)

src/test/java/unit/SpendProfileUser.java
```diff
@@ -108,6 +120,53 @@ void listUsersReflectsEnableDisable() throws IOException {
...
+    String testEmail = generateSamGroupForEmail();
+
+    // use a test user that is an owner to grant spend access to the test email
+    TestUser spendProfileOwner = TestUser.chooseTestUserWithOwnerAccess();
+    spendProfileOwner.login();
+
+    // create test profile if it doesn't exist already
+    TestCommand.runCommand("spend", "create-profile", "--profile=" + TEST_SPEND_PROFILE);
+    // ensure cli spend owners have access
+    TestCommand.runCommand(
```
melissachang, line 136: 135-141 seems like a duplicate of 143-149. I would delete 135-141.

melissachang (approved):

cbookg (approved):

PR #364, opened 2022-11-30
[PF-2268] fix the currentVersion is null cause NPE 

ginaygoogle:
Background:

- This PR comes from the ticket [PF-2268](https://broadworkbench.atlassian.net/browse/PF-2268). The [PF-2174](https://broadworkbench.atlassian.net/browse/PF-2174). 
- The error only happens in Verily server. and shows the current version is null. see [details](https://scans.gradle.com/s/tt2z6g6zxrn4w/tests/overview)
- We dive deep into the code, and find the reason is https://github.com/DataBiosphere/terra-cli/blob/main/src/main/java/bio/terra/cli/app/utils/VersionCheckUtils.java#L65 verily server set the oldest version, so it could not go into the if statement there, when the current version is null. then in L72 it could not parse. 

The way for solve the problem is which learn from `Config.java` https://github.com/DataBiosphere/terra-cli/blob/main/src/main/java/bio/terra/cli/businessobject/Config.java#L51. we add one more if sentence to deal with this situation.

### DIFF
--- a/gradle/test.gradle
+++ b/gradle/test.gradle
@@ -37,7 +37,7 @@ task runTestsWithTag(type: Test) {
             // used to get the default docker image id.
             systemProperty("TERRA_JAR_IMPLEMENTATION_VERSION",
                     "${project.properties['dockerRepoPath']}/${project.properties['dockerImageName']}/${project.version}:${project.properties['dockerImageTag']}")
-            systemProperty("TERRA_VERSION", "${project.version}")
+            systemProperty("TERRA_CLI_VERSION", "${project.version}")
         } else if (testTag == "integration") {
             // [for integration tests] specify the install location for the terra application (i.e. launch script)
             // for an installation directly from source code, this points to the build/install/terra-cli/bin directory (i.e. the output of ./gradlew install)
--- a/gradle/test.gradle
+++ b/gradle/test.gradle
@@ -37,6 +37,7 @@ task runTestsWithTag(type: Test) {
             // used to get the default docker image id.
             systemProperty("TERRA_JAR_IMPLEMENTATION_VERSION",
                     "${project.properties['dockerRepoPath']}/${project.properties['dockerImageName']}/${project.version}:${project.properties['dockerImageTag']}")
+            systemProperty("TERRA_VERSION", "${project.version}")
         } else if (testTag == "integration") {
             // [for integration tests] specify the install location for the terra application (i.e. launch script)
             // for an installation directly from source code, this points to the build/install/terra-cli/bin directory (i.e. the output of ./gradlew install)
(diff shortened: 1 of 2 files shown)


src/main/java/bio/terra/cli/utils/Version.java
```diff
@@ -5,6 +5,10 @@ public class Version {
   /** Getter for the Terra CLI version of the current JAR. */
   public static String getVersion() {
     // read from the JAR Manifest file
-    return Version.class.getPackage().getSpecificationVersion();
+    String version = Version.class.getPackage().getSpecificationVersion();
+    if (version != null) {
+      return version;
+    }
+    return System.getProperty("TERRA_VERSION");
```
yuhuyoyo, line 12: Similar to Config.java, add a comment and a warning log https://github.com/DataBiosphere/terra-cli/blob/main/src/main/java/bio/terra/cli/businessobject/Config.java#L60

yuhuyoyo (approved):

gradle/test.gradle
```diff
@@ -37,6 +37,7 @@ task runTestsWithTag(type: Test) {
             // used to get the default docker image id.
             systemProperty("TERRA_JAR_IMPLEMENTATION_VERSION",
                     "${project.properties['dockerRepoPath']}/${project.properties['dockerImageName']}/${project.version}:${project.properties['dockerImageTag']}")
+            systemProperty("TERRA_CLI_VERSION", "${project.version}")
```
melissachang, line 40: // Normally CLI version comes from jar. For unit tests, there is no jar, so pass version this way.

melissachang (approved):
Nice. My understanding from an offline chat.

- DataBiosphere WSM Unauthenticated.serviceVersion() endpoint doesn't return oldestSupportedCliVersion (for both CLI and live envs)
- verily-src WSM Unauthenticated.serviceVersion() endpoint returns oldestSupportedCliVersion (for both CLI and live envs)

Guang's first PR (#362) passed in DataBiosphere. WSM doesn't return oldestSupportedCliVersion, so [VersionCheckUtils.isOlder()](https://github.com/DataBiosphere/terra-cli/blob/27f5128a62d03c1f880cd6f274e706851757f03c/src/main/java/bio/terra/cli/app/utils/VersionCheckUtils.java#L69) exitted early.

#362 failed in verily-src. WSM returned oldestSupportedCliVersion, so VersionCheckUtils.isOlder() got further. Then [this line](https://github.com/DataBiosphere/terra-cli/pull/364/files#diff-4a10d0dcb7bb1a5912ae919e811fcda23f1cc0a545db6b83721fbdb9390e7c26R12) tried to get version from jar, but failed because there is no jar.

Fix is to pass CLI version via gradle file, similar to docker image.

PR #366, opened 2022-12-01
[PF-2273],[TERRA-200],[TERRA-230],[TERRA-231] Add optional platform parameter to Terra-cli

dexamundsen:
- Add optional platform parameter to Terra-cli create workspace
- Corresponding test utils are updated to use only GCP
- Since parameter is optional, and default value is set to GCP, no impact to users
- Reason for this change: incremental changes are easier to manage, primarily intended for test update
- If an unsupported platform is explicitly chosen, error message is displayed

testing ---> 
```
terra-cli (dexamundsen/2273) >> terra server status
Current server: verily-autopush (Verily auto-push environment)
Server status: OKAY
terra-cli (dexamundsen/2273) >> terra workspace create 
Missing required option: '--id=<id>'
....
      --platform=<cloudPlatform>
                          Set the Cloud platform: AZURE, GCP.
....
terra-cli (dexamundsen/2273) >> terra workspace create --id=dex-ws-1207-1
Test log: --> server: verily-autopush, supportedCloudPlatforms:[GCP], chosen cloudPlatform: GCP
Workspace successfully created.
....
terra-cli (dexamundsen/2273) >> terra workspace create --id=dex-ws-1207-2 --platform=AWS
Invalid value for option '--platform': expected one of [AZURE, GCP] (case-insensitive) but was 'AWS'
....
terra-cli (dexamundsen/2273) >> terra workspace create --id=dex-ws-1207-2 --platform=AZURE
Test log: --> server: verily-autopush, supportedCloudPlatforms:[GCP], chosen cloudPlatform: AZURE
CloudPlatform AZURE not supported for current user dexamundsen@google.com on the current server verily-autopush
terra-cli (dexamundsen/2273) >> 
terra-cli (dexamundsen/2273) >> terra workspace create --id=dex-ws-1207-2 --platform=GCP
Test log: --> server: verily-autopush, supportedCloudPlatforms:[GCP], chosen cloudPlatform: GCP
Workspace successfully created.
```


src/main/java/bio/terra/cli/command/resource/Resolve.java
```diff
@@ -18,7 +18,6 @@
 /** This class corresponds to the third-level "terra resource resolve" command. */
 @Command(name = "resolve", description = "Resolve a resource to its cloud id or path.")
 public class Resolve extends BaseCommand {
-
```
ginaygoogle, line 21: Thanks for the help. It seems come from the last time I manually solve the conflict.

ginaygoogle (approved):

melissachang (commented):
Summarizing an offline conversation :)

- For the foreseeable future, TVC customers will only have one option, GCP. It would be confusing for them to see the `--platform` flag. It would also be inconvenient for them to have to set it every time.
- In the server .json files, we could put which platforms a server supports. In the far future, we may want to customize on a per-customer or per-organization basis. However, per-server should work for a while.
TVC servers only support GCP. broad-dev will support GCP and soon Azure. CWB customers don't use GCP. but CLI repo uses broad-dev for testing.
- It would be nice if --platform flag *only* appeared if the server supported more than one platform.

dexamundsen:
> Summarizing an offline conversation :)
> 
> * For the foreseeable future, TVC customers will only have one option, GCP. It would be confusing for them to see the `--platform` flag. It would also be inconvenient for them to have to set it every time.
> * In the server .json files, we could put which platforms a server supports. In the far future, we may want to customize on a per-customer or per-organization basis. However, per-server should work for a while.
>   TVC servers only support GCP. MC servers only support AWS. broad-dev will support GCP and soon Azure. CWB customers don't use GCP. but CLI repo uses broad-dev for testing.
> * It would be nice if --platform flag _only_ appeared if the server supported more than one platform.

Thanks for the feedback. As discussed, this PR will be on hold. Tickets created - 
https://verily.atlassian.net/browse/TERRA-230
https://verily.atlassian.net/browse/TERRA-231

src/main/java/bio/terra/cli/businessobject/Workspace.java
```diff
@@ -41,25 +41,18 @@
...
-  private OffsetDateTime createdDate;
-  private OffsetDateTime lastUpdatedDate;
+  private final UUID uuid;
+  private final String userFacingId;
+  private final String name; // not unique
+  private final String description;
+  private final CloudPlatform cloudPlatform;
+  private final String googleProjectId;
+  private final Map<String, String> properties;
+  private final String serverName; // name of the server where this workspace exists
```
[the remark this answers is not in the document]
dexamundsen: done

dexamundsen:
> 

Summary - It is not a concern if users w/o access to AWS are shown AWS-related commands, as long as there is graceful error handling + messages. The primary reason is the complexity needed to dynamically show/hide options & subcommands not only based on the server but workspace, workspace override, user permissions (or policy) based on multi-tenancy, or just server changes on the CLI 

 

Much of the command spec is static today. Manual changes to Help messages, options, and subcommands might open up unwarranted complexity in the code. The control flow of the current implementation would also need to be changed. 

hence it was suggested - 1) no need to hide AWS related options (and Azure in the future), add relevant error messages 2) alternate: build specific binaries / packages that only support required cloud platforms (not considered at this time)

src/main/resources/servers/broad-dev-jcarlton.json
```diff
@@ -7,5 +7,6 @@
   "samUri": "https://sam.dsde-dev.broadinstitute.org",
```
melissachang, line 7: This person's no longer on the team, this file can be deleted. :)

Same for wchamber

src/main/java/bio/terra/cli/businessobject/Group.java
```diff
@@ -17,7 +17,6 @@
  */
 public class Group {
   private static final Logger logger = LoggerFactory.getLogger(Group.class);
```
melissachang, line 19: I prefer having a blank line between logger and other variables. That's what most of our code does. Can these be reverted?

src/main/java/bio/terra/cli/command/workspace/Create.java
```diff
@@ -28,13 +28,20 @@ public class Create extends WsmBaseCommand {
   // Variable is `id` instead of `userFacingId` because user sees it with `terra workspace create`
   private String id;
 
+  @CommandLine.Option(
```
melissachang, line 31: As discussed, let's set default to GCP

```
  @CommandLine.Option(
      names = "--platform",
      description = "Set the Cloud platform: ${COMPLETION-CANDIDATES}.",
      defaultValue = "GCP")
```

src/main/java/bio/terra/cli/businessobject/Workspace.java
```diff
@@ -48,16 +48,12 @@ public class Workspace {
   private CloudPlatform cloudPlatform;
   private String googleProjectId;
   private Map<String, String> properties;
-
```
melissachang, line 51: Revert, these newlines make code easier to read.

src/main/java/bio/terra/cli/businessobject/Workspace.java
```diff
@@ -71,6 +67,8 @@ private Workspace(WorkspaceDescription wsmObject) {
       this.cloudPlatform = CloudPlatform.GCP;
     } else if (wsmObject.getAzureContext() != null) {
       this.cloudPlatform = CloudPlatform.AZURE;
+    } else {
```
melissachang, line 70: cloudPlatform should never be null, right? Can you throw an exception instead?

src/main/java/bio/terra/cli/utils/CommandUtils.java
```diff
@@ -0,0 +1,20 @@
...
+import bio.terra.cli.businessobject.Context;
+import bio.terra.cli.exception.UserActionableException;
+import bio.terra.workspace.model.CloudPlatform;
+
+public class CommandUtils {
+  public static void checkPlatformSupport(CloudPlatform cloudPlatform)
+      throws UserActionableException {
+    if (!Context.getServer().getSupportedCloudPlatforms().contains(cloudPlatform)) {
+      throw new UserActionableException(
+          "CloudPlatform "
```
melissachang, line 12:
```
          "Cloud platform "
              + cloudPlatform
              + " not supported for server "
              + Context.getServer().getName());
```

src/test/java/harness/utils/WorkspaceUtils.java
```diff
@@ -23,12 +24,16 @@ public static String createUserFacingId() {
    * @param workspaceCreator The user who owns the workspace. This user will be impersonated to in
    *     the WSM workspaceDelete request.
    */
-  public static UFWorkspace createWorkspace(TestUser workspaceCreator)
+  public static UFWorkspace createWorkspace(TestUser workspaceCreator, CloudPlatform cloudPlatform)
```
melissachang, line 27: Hmm the point of these helper methods is to get rid of boilerplate. Now we have `CloudPlatform.GCP` in a bunch of places. What do you think:

Rename this to createGcpWorkspace. Either hard-code to GCP, or omit (will default to GCP).

melissachang (commented):
Nice!

I get errors

```
~/terra-cli (dexamundsen/2273): terra workspace create --id=test-platform
[ERROR] An unexpected error occurred in java.lang.NullPointerException: Cannot invoke "java.util.Set.contains(Object)" because the return value of "bio.terra.cli.businessobject.Server.getSupportedCloudPlatforms()" is null

~/terra-cli (dexamundsen/2273): terra workspace create --id=test-platform --platform=AZURE
[ERROR] An unexpected error occurred in java.lang.NullPointerException: Cannot invoke "java.util.Set.contains(Object)" because the return value of "bio.terra.cli.businessobject.Server.getSupportedCloudPlatforms()" is null
```

melissachang:
FYI @rtitle

src/main/resources/servers/broad-dev-cli-testing.json
```diff
@@ -7,5 +7,6 @@
   "samUri": "https://sam.dsde-dev.broadinstitute.org",
   "workspaceManagerUri": "https://workspace.verilycli.integ.envs.broadinstitute.org",
   "wsmDefaultSpendProfile": "wm-default-spend-profile",
-  "externalCredsUri": "https://terra-devel-ecm.api.verily.com"
+  "externalCredsUri": "https://terra-devel-ecm.api.verily.com",
+  "supportedCloudPlatforms": ["GCP"]
```
melissachang, line 11: Can the broad server have AZURE in addition to GCP?

src/test/java/harness/utils/WorkspaceUtils.java
dexamundsen: This change was a quick change until we add tests for AWS, at which time the same functions could have been used with multiple clouds. 

however with addition of GCP as default value, this entire change in tests can be removed

dexamundsen:
> getSupportedCloudPlatforms

We dont need this to error out, since it on the CommandSpec option defaults to GCP

src/main/java/bio/terra/cli/command/workspace/Create.java
```diff
@@ -28,13 +28,21 @@ public class Create extends WsmBaseCommand {
...
+      names = "--platform",
+      description = "Set the Cloud platform: ${COMPLETION-CANDIDATES}.",
+      defaultValue = "GCP")
```
melissachang, line 34: I noticed we weren't showing default value with `terra workspace create`. Looks like you need `showDefaultValue`:

```
  @CommandLine.Option(
      names = "--platform",
      description = "Set the Cloud platform: ${COMPLETION-CANDIDATES}.",
      showDefaultValue = CommandLine.Help.Visibility.ALWAYS,
      defaultValue = "GCP")
  private CloudPlatform cloudPlatform;
```

src/main/java/bio/terra/cli/businessobject/Server.java
```diff
@@ -8,13 +8,15 @@
 import bio.terra.cli.utils.FileUtils;
 import bio.terra.cli.utils.JacksonMapper;
 import bio.terra.datarepo.model.RepositoryStatusModel;
+import bio.terra.workspace.model.CloudPlatform;
```
melissachang, line 11: I'm still seeing [these errors](https://github.com/DataBiosphere/terra-cli/pull/366#pullrequestreview-1209326445)

src/main/java/bio/terra/cli/command/workspace/Create.java
dexamundsen: Yes, noted and added.

src/main/java/bio/terra/cli/businessobject/Server.java
dexamundsen: This should ideally never show up as long as the server config file is set up properly, and errors with the file would show up during deployment (which might not be convenient). 

The assumption here was that GCP, even if default has to be specified on the server config as supported. This way we decouple the config from the implementation. 
For the sake of UX we pick a default value, but that might not be expected on the infra. 

To avoid exceptions as these, initializing it to empty set. 

Could you share the server name used for this? and confirm that the corresponding server has the platforms enabled?

src/main/java/bio/terra/cli/businessobject/Server.java
```diff
@@ -228,5 +232,9 @@ public boolean getSupportsIdToken() {
 
   public Set<CloudPlatform> getSupportedCloudPlatforms() {
     return supportedCloudPlatforms;
+
+    // this.supportedCloudPlatforms = builder.supportedCloudPlatforms != null ?
```
melissachang, line 236: Delete ?

src/main/java/bio/terra/cli/businessobject/Server.java
melissachang: Hmm I was doing `./gradlew installDist`. If I do `source tools/local-dev.sh` (as you suggested), it works, thanks.

melissachang (approved):

src/main/java/bio/terra/cli/businessobject/Server.java
dexamundsen: Done

PR #369, opened 2022-12-05
PF-2302: check for stale resource list

yuhuyoyo:
Created the `bucket1` and `dataset1` resource through UI. They are reflected in `terra app execute env`.
Before the change, you need to call `terra resource list` again to resync the cache.
```
yuhuyoyo-macbookpro3% terra app execute env | sort | grep "TERRA"
Dec 05, 2022 5:20:05 PM com.google.auth.oauth2.DefaultCredentialsProvider warnAboutProblematicCredentials
WARNING: Your application has authenticated using end user credentials from Google Cloud SDK. We recommend that most server applications use service accounts instead. If your application continues to use end user credentials from Cloud SDK, you might receive a "quota exceeded" or "API not enabled" error. For more information about service accounts, see https://cloud.google.com/docs/authentication/.
TERRA_USER_EMAIL=yuhuyoyo@google.com
TERRA_bucket1=gs://bucket1-terra-vdevel-happy-cherry-2812
TERRA_dataset1=terra-vdevel-happy-cherry-2812.dataset1
TERRA_notebook=projects/terra-vdevel-happy-cherry-2812/locations/us-central1-a/instances/notebook
```


melissachang (approved):
This is fine. Note PF-1889 will fix the whole class of resource caching bugs, including PF-2302. I hope to work on it in the next month.

I have a slight preference for reverting your test changes, since they will no longer be necessary after PF-1889 is done.

cbookg (approved):

PR #378, opened 2022-12-13
[PF-2348] Wrap command tokens with quotes

iliadsh:
Previously, running a command like 
`terra gcloud compute ssh --zone us-central1-a jupyter@newenv --command "ps aux"`
 would fail to parse, as the space information in the quoted part would be lost when the command is built.

### DIFF
--- a/src/main/java/bio/terra/cli/app/CommandRunner.java
+++ b/src/main/java/bio/terra/cli/app/CommandRunner.java
@@ -8,6 +8,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.stream.Collectors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


melissachang (commented):
Do single quotes work?

https://github.com/DataBiosphere/terra-cli/#:~:text=%2D%2Dcommand%20%27echo%20%22Hello%20World%22%20%3E%20%22%24%7BOUT%7D%22%27%20%5C

If single quotes work, let's revert this PR

cbookg:
> 

But double-quotes should work too.  Though it doesn't make sense to me that these would be different?  The flag parsing library should be providing the underlying value regardless?

iliadsh:
> Do single quotes work?

Just tested it with the ssh command and it doesn't appear to work. I doubt the dsub example works as intended either.. but haven't tried.

melissachang:
Good point that double quotes should work, and that flag-parsing library (picocli) should handle this.

https://github.com/remkop/picocli/issues/851

https://stackoverflow.com/questions/63629789/how-to-escape-chars-from-string-option

Is it possible to write a test that fails before this PR and passes after this PR?

cbookg (approved):

PR #394, opened 2023-01-06
PF-2376: add a ssh-key add command 

yuhuyoyo:
There is a user friction when a user created an AI notebook for the first time and didn't set up an ssh key. Or they regenerate a new ssh key and deleted the old one from their github account after a notebook is already created. 

Right now there's no easy way to get and add the new terra ssh key to the notebook.

Adding a `terra user ssh-key add` command to get and save the key and add it to the ssh-agent.

Also did a refactor to achieve the same with `terra user ssh-key generate --save-to-file`.

```
yuhuyoyo-macbookpro3% terra user ssh-key add
Saved the terra ssh key to ~/.ssh/terra_id_rsa and ~/.ssh/terra_id_rsa.pub
Agent pid 39723
Identity added: /Users/yuhuyoyo/.ssh/terra_id_rsa (/Users/yuhuyoyo/.ssh/terra_id_rsa)
ssh-rsa <my-key>
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -747,6 +747,24 @@ for more details.
 
 #### User
 
+#### ssh-key
+
+`terra user ssh-key` is how Terra do source control in a notebook environment. 
+It handles the ssh key of the current user. There is one single Terra ssh key
+per user in a given server (e.g. broad-dev). With this SSH key, you can perform
+source control in a terra-managed notebook instance using git.
+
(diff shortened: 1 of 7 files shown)


dexamundsen (approved):

emmarogge (commented):
LGTM! Thank you for your quick action on this, @yuhuyoyo -- it makes a significant difference in reducing user friction!

PR #400, opened 2023-01-11
[PF-2333] Read Janitor SA key properly

zloery:
Yesterday I updated the Janitor SA key in Vault to no longer be base64 encoded so that it's readable by the terraform-GHA mechanism for the CLI, but forgot to update this script for local rendering.


dexamundsen (approved):

zloery:
This still doesn't work because the script wants the key as a top-level JSON object in vault instead of being in a single field. For now, I'm reverting the value in vault back to b64 and fixing the GHA.

We have three different formats for storing things in vault, I'll file a ticket for standardizing a bit

PR #409, opened 2023-01-26
Check Java version in local-dev.sh

wnojopra:
Similar to the check we have in [download-install.sh](https://github.com/DataBiosphere/terra-cli/blob/main/tools/download-install.sh).

Some questions that may come up:

> This code looks almost exactly copy-pasted from `download-install.sh`. Why not move it to a common util script?

In the main Readme we suggest users call the download-install script by running `curl -L https://github.com/DataBiosphere/terra-cli/releases/latest/download/download-install.sh | bash`. Moving the common code out to a utility script would break that flow.  

> Why isn't the code exactly copy-pasted? What changes did you make?

In the failure case, instead of doing `exit 1` I do a `return 1`. This is because the usage of this script is `source local-dev.sh`, which means we run it in the current shell. I don't want to exit from the current shell.

### DIFF
--- a/tools/local-dev.sh
+++ b/tools/local-dev.sh
@@ -1,13 +1,28 @@
 #!/bin/bash
+set -e
 
 ## This script sets up the environment for local development.
 ## Dependencies: docker, chmod
 ## Usage: source tools/local-dev.sh
 
+REQ_JAVA_VERSION=17
+echo "--  Checking if installed Java version is ${REQ_JAVA_VERSION} or higher"
+if [[ -n "$(which java)" ]]; then
+  # Get the current major version of Java: "11.0.12" => "11"
+  CUR_JAVA_VERSION="$(java -version 2>&1 | awk -F\" '{ split($2,a,"."); print a[1]}')"
+  if [[ "${CUR_JAVA_VERSION}" -lt ${REQ_JAVA_VERSION} ]]; then
+    >&2 echo "ERROR: Java version detected (${CUR_JAVA_VERSION}) is less than required (${REQ_JAVA_VERSION})"
+    return 1
+  fi
+else
+  >&2 echo "ERROR: No Java installation detected"
+  return 1
+fi
+
 ## The script assumes that it is being run from the top-level directory "terra-cli/".
 if [[ "$(basename "$PWD")" != 'terra-cli' ]]; then
   >&2 echo "ERROR: Script must be run from top-level directory 'terra-cli/'"
-  exit 1
+  return 1
 fi
 
 echo "Building Java code"
--- a/tools/local-dev.sh
+++ b/tools/local-dev.sh
@@ -22,6 +22,7 @@ function check_java_version() {
 }
 
 if ! check_java_version; then
+  unset check_java_version
   return 1
 fi
 
--- a/tools/local-dev.sh
+++ b/tools/local-dev.sh
@@ -25,7 +25,6 @@ if ! check_java_version; then
   unset check_java_version
   return 1
 fi
-
 unset check_java_version
 
 ## The script assumes that it is being run from the top-level directory "terra-cli/".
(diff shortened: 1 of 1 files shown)


dexamundsen (approved):

tools/local-dev.sh
```diff
@@ -1,9 +1,24 @@
 #!/bin/bash
+set -e
 
 ## This script sets up the environment for local development.
 ## Dependencies: docker, chmod
 ## Usage: source tools/local-dev.sh
 
+REQ_JAVA_VERSION=17
```
mbookman, line 8: Since this is being sourced, this will set a shell variable REQ_JAVA_VERSION that will persist after the script runs. I think that would be an unintended side-effect.

tools/local-dev.sh
```diff
@@ -1,9 +1,24 @@
 #!/bin/bash
+set -e
```
mbookman, line 2: Since this is being sourced, this will set errexit in the shell, even after the script exits. I think this would be an unintended side-effect.

tools/local-dev.sh
```diff
@@ -1,9 +1,24 @@
...
+    return 1
+  fi
+else
+  >&2 echo "ERROR: No Java installation detected"
+  return 1
+fi
+
 ## The script assumes that it is being run from the top-level directory "terra-cli/".
 if [[ "$(basename "$PWD")" != 'terra-cli' ]]; then
   >&2 echo "ERROR: Script must be run from top-level directory 'terra-cli/'"
```
mbookman, line 24: Existing problem, but if this fails, won't it exit the current shell?

mbookman (changes requested):

tools/local-dev.sh
mbookman: Not sure the *best* pattern to follow, but here is a suggestion to avoid side-effects.

```
function check_java_version() {
  local REQ_JAVA_VERSION=17

  echo "--  Checking if installed Java version is ${REQ_JAVA_VERSION} or higher"
  if [[ -n "$(which java)" ]]; then
    # Get the current major version of Java: "11.0.12" => "11"
    local CUR_JAVA_VERSION="$(java -version 2>&1 | awk -F\" '{ split($2,a,"."); print a[1]}')"
    if [[ "${CUR_JAVA_VERSION}" -lt ${REQ_JAVA_VERSION} ]]; then
      >&2 echo "ERROR: Java version detected (${CUR_JAVA_VERSION}) is less than required (${REQ_JAVA_VERSION})"
      return 1
    fi
  else
    >&2 echo "ERROR: No Java installation detected"
    return 1
  fi
}

if ! check_java_version; then
  unset check_java_version
  return 1
fi
```

mbookman (changes requested):

tools/local-dev.sh
wnojopra: Yes, it does. I'm changing it from exit to return.

tools/local-dev.sh
wnojopra: Thanks, I totally missed this and it was causing me problems. I've removed the `set -e`.

tools/local-dev.sh
wnojopra: Thanks, I do like this better. I've included this change.

tools/local-dev.sh
```diff
@@ -4,10 +4,32 @@
...
+  else
+    >&2 echo "ERROR: No Java installation detected"
+    return 1
+  fi
+}
+
+if ! check_java_version; then
+  unset check_java_version
+  return 1
+fi
```
mbookman, line 27: Just realized that to be clean, we should unset on success too.

mbookman (changes requested):

tools/local-dev.sh
wnojopra: Thanks, fixed.

mbookman (approved):
LGTM.

PR #411, opened 2023-01-27
Basic support for Bash tab-completion

wnojopra:
Basic support for Bash tab-completion. Makes the Terra users' life a bit easier.

image

A future PR may add support for additional parameters and commands.

### DIFF
--- a/bash_tab_complete
+++ b/bash_tab_complete
@@ -23,7 +23,10 @@ _go_with_options()
 _magic_options_for_terra()
 {
     local cmdline_options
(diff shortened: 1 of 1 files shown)


bash_tab_complete
```diff
@@ -0,0 +1,33 @@
+# To use, in the Bash shell:
```
mbookman, line 1: I think we want to use the built-in capabilities that come with the picocli. I'll see if I can do some testing with this later:

https://picocli.info/autocomplete.html

wnojopra:
Closing in favor of #423 instead

PR #417, opened 2023-01-31
[PF-2333] Continued test fixes

zloery:
I'll leave this branch open for a while to see PR runs and keep adding more fixes over time as I see more failures pop up in nightly tests.

So far:

- Bump Nextflow sleep from 15m to 20m, as the test user's pet frequently does not have bucket permission after 15m.
- Retry `gcloud builds submit` test command directly instead of polling `list bucket` first as permissions can flip-flop

### DIFF
--- a/src/test/java/unit/GcloudBuildsSubmit.java
+++ b/src/test/java/unit/GcloudBuildsSubmit.java
@@ -53,20 +53,11 @@ public class GcloudBuildsSubmit extends SingleWorkspaceUnitGcp {
 
     // `terra resource create gcs-bucket --name=$name --format=json`
     String bucketResourceName = "resourceName";
-    UFGcsBucket createdBucket =
-        TestCommand.runAndParseCommandExpectSuccess(
+    TestCommand.runAndParseCommandExpectSuccess(
             UFGcsBucket.class, "resource", "create", "gcs-bucket", "--name=" + bucketResourceName);
 
-    // Poll until the test user can fetch the bucket, which may be delayed.
-    Storage ownerClient =
-        ExternalGCSBuckets.getStorageClient(
-            workspaceCreator.getCredentialsWithCloudPlatformScope());
-    Page<Blob> bucketContents =
-        CrlUtils.callGcpWithPermissionExceptionRetries(
-            () -> ownerClient.get(createdBucket.bucketName).list());
-
     // `builds submit --async --gcs-bucket-resource=bucketName --tag=$tag`
-    TestCommand.runCommandExpectSuccess(
+    TestCommand.runCommandExpectSuccessWithRetries(
         "gcloud",
         "builds",
         "submit",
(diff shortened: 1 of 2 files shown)


src/test/resources/testscripts/NextflowRnaseqSetup.sh
```diff
@@ -0,0 +1,34 @@
...
+echo "resourceName: $resourceName, bucketName: $bucketName"
+terra resource create gcs-bucket --name=$resourceName --bucket-name=$bucketName
+terra resource describe --name=$resourceName
+
+# I will use an example Nextflow workflow from a GitHub repository [show webpage], and checkout a tag that I have tested beforehand. This is the same example workflow that is used on the GCP + Nextflow tutorial [show webpage].
+
+git clone https://github.com/nextflow-io/rnaseq-nf.git
+cd rnaseq-nf
+git checkout v2.0
+cd ..
```
dexamundsen, line 34: nitty: new line

src/test/java/integration/Nextflow.java
```diff
@@ -56,7 +56,15 @@ void nextflowFromGLSTutorial() throws IOException {
     assertEquals(0, exitCode, "workspace created without errors");
 
     // run the script that downloads the NF workflow from GH and runs it
-    exitCode = TestBashScript.runScript("NextflowRnaseq.sh");
+    TestBashScript.runScript("NextflowRnaseqSetup.sh");
+
+    // Try running the script once, which will likely (but not definitely) fail due to missing
+    // permissions on the bucket.
+    TestBashScript.runScript("NextflowRnaseqWorkflow.sh");
```
dexamundsen, line 63: Would it help to retry only if the first run failed? i.e. check the exit code and retry if needed. 
A retry on success wouldn't cause a problem here since Idempotence is not required for this script

src/test/resources/testscripts/NextflowRnaseqWorkflow.sh
```diff
@@ -48,6 +11,7 @@ terra nextflow -version
 #      google.lifeSciences.network = 'network'
 #      google.lifeSciences.subnetwork = 'subnetwork'
 mv rnaseq-nf/nextflow.config rnaseq-nf/nextflow.config_original
+resourceName="terraclitesting"
```
dexamundsen, line 14: Since same value of resourceName should be used in setup & workflow scripts, might be easier to maintain if it is passed as an arg in Nextflow.java

dexamundsen (approved):

src/test/java/integration/Nextflow.java
zloery: Good call, that potentially saves 10-15 minutes each run

src/test/resources/testscripts/CreateWorkspace.sh
```diff
@@ -11,9 +11,6 @@ if [[ "true" == "$isLoggedIn" ]]; then
   echo "User $currentUser is logged in. Creating a new workspace."
   terra workspace create --id=my-workspace-$RANDOM
```
yuhuyoyo, line 12: this step could timeout if wsm workspace creation takes too long

yuhuyoyo (approved):

PR #423, opened 2023-02-03
Add generate-completion subcommand.

mbookman:
Supercedes #411.

First commit to add support for bash tab completion. This simply adds a sub-command to the terra-cli such that it can generate its own tab-completion script, as per https://picocli.info/autocomplete.html#_distribution.

This PR does not attempt to install the tab completion script. We'll do that in a follow-up.


wnojopra (approved):
Cool! So I imagine one needs to run something like `terra generate-completion` and then source the output? 

LGTM assuming merge and tests are good.

src/main/java/bio/terra/cli/command/Main.java
```diff
@@ -112,6 +114,10 @@ public static int runCommand(String... args) {
     cmd.setColorScheme(colorScheme);
     cmd.setCaseInsensitiveEnumValuesAllowed(true);
 
+    // Hide the "generate-completion" subcommand
+    CommandLine gen = cmd.getSubcommands().get("generate-completion");
+    gen.getCommandSpec().usageMessage().hidden(true);
```
yuhuyoyo, line 119: i'm confused how this should be used, given you "hide" it?

mbookman: Thanks, Yu. Good question.

My plan was to separate out the addition of the command (this PR) with how we end up using it in installs and other set up. What I've observed so far is that actually installing/configuring tab completion likely has some complexity to navigate depending on environment. I thought we'd get it set up on a cloud environment first, by doing the right thing in the [WSM post-startup script](https://github.com/DataBiosphere/terra-workspace-manager/blob/main/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh), and then work on setup in other environments.

yuhuyoyo: so post-startup.sh will set it up by calling generate-completion and the end-user shouldn't need to worry about this?

mbookman: Right. Hiding the command hides it from the generated tab-completion script, but doesn't prevent a user from calling it.

yuhuyoyo (approved):

PR #440, opened 2023-02-23
PF-2500: Update ecm client version and handle null string in `writeToFile`.

yuhuyoyo:


zloery (approved):

src/main/java/bio/terra/cli/command/user/sshkey/Get.java
```diff
@@ -16,7 +17,12 @@ public class Get extends BaseCommand {
   @Override
   protected void execute() {
     ExternalCredentialsManagerService ecmService = ExternalCredentialsManagerService.fromContext();
-    var sshKeyPair = ecmService.getSshKeyPair(SshKeyPairType.GITHUB);
+    // Post-startup.sh is parsing the private key for the json and store it in the notebook.
+    // https://github.com/DataBiosphere/terra-workspace-manager/blob/c8c17e456142577d49fc8879c43ef9f034feb884/service/src/main/java/bio/terra/workspace/service/resource/controlled/cloud/gcp/ainotebook/post-startup.sh#L185
+    var sshKeyPair =
+        ecmService.getSshKeyPair(
+            SshKeyPairType.GITHUB,
+            /*includePrivateKey=*/ formatOption.getEffectiveFormatOption() == FormatOptions.JSON);
```
dexamundsen, line 25: Checking the format option (i.e. output format) might not be the best option here. 
The output format might as well be used in other scripts or CLI where we don't need the private key. 

alternate options to consider - 
1. If it is a given that CLI will always need private key, we could hard code it to true -> unlikely the need here
2. Add a user input option to include private key, default=false. This will be a breaking change if users expect private key by default. If mostly used in scripts (as in post-startup.sh) we can update them. For users - we add it as part of the changelog / readme. 
3. Any other mechanism to detect who / what script is calling the CLI. We don't currently have the support to detect it, and would be a bigger design / discussion

src/main/java/bio/terra/cli/command/user/sshkey/Get.java
```diff
@@ -13,10 +13,16 @@
 public class Get extends BaseCommand {
   @CommandLine.Mixin Format formatOption;
 
+  @CommandLine.Option(names = "--include-private-key", description = "User email.")
```
dexamundsen, line 16: minor: description needs update

dexamundsen (approved):

PR #447, opened 2023-03-14
PF-2554: change clone to duplicate

yuhuyoyo:


dexamundsen (commented):
Changes may be needed here - 
- https://github.com/DataBiosphere/terra-cli/blob/main/src/test/java/unit/Workspace.java#L58

are there plans to change the WSM api? 
- https://github.com/DataBiosphere/terra-cli/blob/main/src/main/java/bio/terra/cli/service/WorkspaceManagerService.java#L629

yuhuyoyo:
> Changes may be needed here -
> 
> * https://github.com/DataBiosphere/terra-cli/blob/main/src/test/java/unit/Workspace.java#L58
> 
Ah good catch.
> are there plans to change the WSM api?
> 
> * https://github.com/DataBiosphere/terra-cli/blob/main/src/main/java/bio/terra/cli/service/WorkspaceManagerService.java#L629
No, we will keep using clone in the backend. we just want the client to be consistent.

dexamundsen (approved):

zloery (approved):
LGTM, can you also update the references to `clone` in the README?

PR #451, opened 2023-03-28
Add optional concurrent-job-limit flag to terra cromwell generate-config

wnojopra:
Previously, concurrent-job-limit was hard coded to 10. This means larger workflows are bottlenecked by only having 10 VMs up at a time. 

This PR adds an optional `concurrent-job-limit` to configure this value. I have it default to 10, but I'm open to opinions to change it. Cromwell by default does not set this value, so it's technically unlimited. I think 10 is a fine default, as it would give time for newer users to cancel an accidentally massive workflow.

### DIFF
--- a/src/main/java/bio/terra/cli/command/cromwell/GenerateConfig.java
+++ b/src/main/java/bio/terra/cli/command/cromwell/GenerateConfig.java
@@ -30,6 +30,13 @@ public class GenerateConfig extends BaseCommand {
       description = "Directory to put generated cromwell.conf in. Defaults to current directory.")
   public String dir;
 
+  @CommandLine.Option(
+      names = "--concurrent-job-limit",
+      defaultValue = "10",
+      required = false,
+      description = "Optional limits on the number of concurrent jobs. Defaults to 10.")
+  public String concurrent_job_limit;
+
   // We create an exclusive ArgGroup because we require one of either
   // workspace_bucket_name or google_bucket_name.
   @CommandLine.ArgGroup(exclusive = true, multiplicity = "1")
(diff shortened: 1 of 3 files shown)


src/main/resources/configs/generate-cromwell-config.sh
```diff
@@ -40,7 +48,7 @@ backend {
 
         genomics {
           auth = "application_default"
-          compute-service-account = "$3"
+          compute-service-account = "$PET_SA_EMAIL"
```
mbookman, line 51:
```
          compute-service-account = "${PET_SA_EMAIL}"
```

src/main/java/bio/terra/cli/command/cromwell/GenerateConfig.java
```diff
@@ -30,6 +30,13 @@ public class GenerateConfig extends BaseCommand {
       description = "Directory to put generated cromwell.conf in. Defaults to current directory.")
   public String dir;
 
+  @CommandLine.Option(
```
mbookman, line 33: I *think* this is a good flag to add.

This will be interesting to see if we find we are slowly but surely adding more and more flags to this command because Cromwell is a big application with lots of possible settings.

The bucket name flag makes a lot of sense because it is a requirement. This one isn't required, but I agree that it is nice to have some built-in guardrails for our users.

mbookman (changes requested):

src/main/java/bio/terra/cli/command/cromwell/GenerateConfig.java
wnojopra: I think the good news is our CLI doesn't block users from configuring Cromwell however they want (users can fairly easily just change the config file directly). I think the main reason to add flags is for automated scripts and Getting Started notebooks that need programmatic access to the config file. 

I would like to add one more flag for configuring cromwell to communicate with a local database. And I can see users possibly needing options for things like call-caching, log directories, possibly others. But hopefully Terra has a better answer for workflows before "too many cli options" is actually a problem.

mbookman (approved):
LGTM

wnojopra:
@yuhuyoyo or @harrismendell Would either of you be willing to review? Or perhaps add another reviewer who would be appropriate to review? Thanks!

PR #455, opened 2023-04-10
[PF-2656] Add terra resource mount command

rogerwangcs:
This change adds `terra resources mount` and `terra resources unmount` commands:
- Supports mounting and unmounting all workspace bucket resources.
- Also supports `--disable-cache` flag.

Features to add in followup:
- Setting / toggling permissions on mount
- mounting/unmounting individual files

Note: `terra resource mount` only works in no docker mode (which is what we need on cloud environments).

Example Outputs:
```
❯ terra resource mount
One or more resources failed to unmount.
❯ tree -L 3 ~/workspace
/Users/rogerwangcs/workspace
├── Datasets
│   ├── bad-ref_NOT_FOUND
│   └── cloud-bucket-1
├── FakeFolderFile
├── clinvar
│   ├── README.txt
│   ├── disease_names.txt
│   ├── variant_summary.txt
│   └── variant_summary.txt.gz
├── empty-folder
├── example-r-markdown-rmd-referenced-object
├── file-preview-bucket
├── genomics-public-data
│   ├── NA12878.chr20.sample.DeepVariant-0.7.2.vcf
│   ├── NA12878.chr20.sample.bam
│   ├── README
│   ├── clinvar
│   │   ├── README.txt
│   │   ├── disease_names.txt
│   │   ├── variant_summary.txt
│   │   └── variant_summary.txt.gz
│   ├── precision-fda
│   │   └── README.txt
│   └── ucsc
│       └── annotations
├── invalid-bucket-object-ref
├── notebookObject
├── private-bucket-reference_NO_ACCESS
└── service-openapi-yaml
```

```
❯ terra resource unmount
Unmounted workspace resources.
```


src/main/java/bio/terra/cli/utils/MountUtils.java
```diff
@@ -0,0 +1,241 @@
...
+  }
+
+  /**
+   * Recursively deletes all empty subdirectories in the WORKSPACE_DIR, excluding the WORKSPACE_DIR
+   * itself.
+   *
+   * <p>Throws UserActionableException if a non-empty directory is encountered. Throws
+   * SystemException if there is an error during deletion.
+   */
+  public static void deleteResourceDirectories() {
```
rogerwangcs, line 195: A similar implementation is in the test context: https://github.com/DataBiosphere/terra-cli/blob/main/src/test/java/harness/TestContext.java#L90. Am curious if there's a pattern to share utility functions between testing and main.

docker/Dockerfile
```diff
@@ -33,6 +36,13 @@ ENV NXF_MODE google
 RUN curl -s https://get.nextflow.io | bash \
     && mv nextflow /usr/local/bin
 
+# install gcsfuse for mount tests
+RUN export GCSFUSE_REPO=gcsfuse-$(lsb_release -c -s) \
+    && echo "deb https://packages.cloud.google.com/apt $GCSFUSE_REPO main" | tee /etc/apt/sources.list.d/gcsfuse.list \
+    && curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key add -
+RUN apt-get update \
+    && apt-get install -y gcsfuse
```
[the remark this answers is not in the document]
yuhuyoyo: it's not only for test but also for docker mode. so i think you can remove the tests in the comment to be more accurate.

src/main/java/bio/terra/cli/app/utils/LocalProcessLauncher.java
```diff
@@ -76,6 +77,12 @@ public void launchProcess(
     }
   }
 
+  public Process launchSilentProcess(List<String> command, Map<String, String> envVars, Path workingDirectory) {
+    launchProcess(command, null, null);
```
yuhuyoyo, line 81: i don't follow. why passing in envVars and workingDirectory but not using them?

src/main/java/bio/terra/cli/app/utils/LocalProcessLauncher.java
```diff
@@ -76,6 +77,12 @@ public void launchProcess(
     }
   }
 
+  public Process launchSilentProcess(List<String> command, Map<String, String> envVars, Path workingDirectory) {
```
yuhuyoyo, line 80: add java doc for public method

unclear to me why this is called silent process.

src/main/java/bio/terra/cli/businessobject/ResourcePropertyNames.java
```diff
@@ -0,0 +1,15 @@
+package bio.terra.cli.businessobject;
+
+public enum ResourcePropertyNames {
+  FolderId("terra-folder-id");
```
yuhuyoyo, line 4: java enums are usually all CAP. FOLDER_ID.

src/main/java/bio/terra/cli/businessobject/resource/GcsObject.java
```diff
@@ -112,7 +117,7 @@ protected void deleteControlled() {
 
   /** Resolve a GCS bucket object resource to its cloud identifier. */
   public String resolve() {
-    return resolve(/*includeUrlPrefix=*/ true);
```
yuhuyoyo, line 115: can you not remove comment?

src/main/java/bio/terra/cli/businessobject/resource/GcsObject.java
```diff
@@ -125,6 +130,27 @@ public String resolve(boolean includeUrlPrefix) {
...
+   * with a trailing slash. There will always be an object present to represent the directory/prefix
+   * itself.
+   */
+  public boolean isPrefix() throws SystemException {
+    try {
+      BucketCow bucketCow =
+          CrlUtils.createStorageCow(Context.requireUser().getPetSACredentials()).get(bucketName);
+      Iterable<BlobCow> objects = bucketCow.list(BlobListOption.prefix(objectName)).getValues();
+      for (BlobCow object : objects) {
+        if (object.getBlobInfo().getName().endsWith("/")) {
```
yuhuyoyo, line 144: blobInfo has an isDirectory method https://cloud.google.com/java/docs/reference/google-cloud-storage/latest/com.google.cloud.storage.BlobInfo#com_google_cloud_storage_BlobInfo_isDirectory__

src/main/java/bio/terra/cli/businessobject/resource/GcsObject.java
```diff
@@ -125,6 +130,27 @@ public String resolve(boolean includeUrlPrefix) {
         : bucketName + "/" + objectName;
   }
 
+  /**
+   * Check if the GCS bucket object is a directory by checking for the presence of a child object
+   * with a trailing slash. There will always be an object present to represent the directory/prefix
+   * itself.
+   */
+  public boolean isPrefix() throws SystemException {
```
yuhuyoyo, line 138: isPrefix -> isDirectory?

src/main/java/bio/terra/cli/command/resource/ListTree.java
```diff
@@ -24,7 +25,6 @@ public class ListTree extends WsmBaseCommand {
   private static final HashMap<UUID, String> ID_TO_NAME = new HashMap<>();
   private static final HashMap<UUID, Boolean> IS_FOLDER = new HashMap<>();
   private static final UUID ROOT = UUID.randomUUID();
-  private static final String TERRA_FOLDER_ID_PROPERTY_KEY = "terra-folder-id";
```
yuhuyoyo, line 27: i don't see the value of changing this to a enum yet. when there are 3+ resource properties that we want to iterate through yes, but given there's only one and we are specifically looking for it, static field is enough.

src/main/java/bio/terra/cli/utils/MountUtils.java
yuhuyoyo: yeah test can use this one too if you want to migrate those usage.

src/main/java/bio/terra/cli/utils/MountUtils.java
```diff
@@ -0,0 +1,241 @@
...
+    resourceMountPaths.forEach(
+        (id, mountPath) -> {
+          Resource r = ws.getResource(id);
+          ResourceMountHandler handler = getMountHandler(r, mountPath, null);
+          handler.unmount();
+        });
+
+    // Delete mount point directories if they are empty, so we do not unintentionally delete
+    // bucket files or local files stored at mounted directories
+    deleteResourceDirectories();
```
yuhuyoyo, line 84: instead of adding the comment, just rename the method to deleteResourceDirectoriesIfEmpty

src/main/java/bio/terra/cli/businessobject/mounthandler/GcsFuseMountHandler.java
```diff
@@ -0,0 +1,124 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * This class handles mounting a GCS bucket or prefix object using the GCS FUSE driver.
+ */
+public class GcsFuseMountHandler extends ResourceMountHandler {
+  Logger logger = LoggerFactory.getLogger(GcsFuseMountHandler.class);
+  protected final String FUSE_MOUNT_COMMAND = "gcsfuse";
+  protected String bucketName;
+  protected String subDir;
```
yuhuyoyo, line 22: nullable

src/main/java/bio/terra/cli/utils/MountUtils.java
```diff
@@ -0,0 +1,241 @@
...
+   * Helper method to get mountable resources in the workspace, any file containing resource or
+   * reference to a file containing resource.
+   */
+  public static List<Resource> getMountableResources() {
+    return Context.requireWorkspace().listResources().stream()
+        .filter(
+            r -> {
+              if (r.getResourceType() == Resource.Type.GCS_BUCKET) {
+                return true;
+              } else if (r.getResourceType() == Resource.Type.GCS_OBJECT) {
```
yuhuyoyo, line 123: don't need else if, use if instead,  because 122 return.

src/main/java/bio/terra/cli/utils/MountUtils.java
```diff
@@ -0,0 +1,241 @@
...
+        .toList();
+  }
+
+  /**
+   * Helper method to get workspace folder paths. Used to build the mount paths for mountable
+   * resources.
+   *
+   * @return A map of folder IDs to folder paths.
+   */
+  private static Map<UUID, Path> getFolderPaths() {
```
yuhuyoyo, line 143: getFolderIdToFolderPathMap

docker/Dockerfile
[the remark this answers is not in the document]
yuhuyoyo: but also, please add test.

src/main/java/bio/terra/cli/app/utils/LocalProcessLauncher.java
rogerwangcs: Oops thanks. I'm removing this method since its unneeded abstraction, using `launchProcess` and `waitForTerminate` instead.

src/main/java/bio/terra/cli/utils/MountUtils.java
```diff
@@ -0,0 +1,241 @@
...
+        }
+      }
+    }
+  }
+
+  /**
+   * Recursively deletes all empty subdirectories in the WORKSPACE_DIR, excluding the WORKSPACE_DIR
+   * itself.
+   *
+   * <p>Throws UserActionableException if a non-empty directory is encountered. Throws
```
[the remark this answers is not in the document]
rogerwangcs: Thanks for pointing this out. So this error would most likely be thrown when the user already has non empty directories in `~/workspace` and the user tries to run `terra resource mount`. This is unlikely to happen on CE's when this command will run in post startup. The other rare case is if users manually mess with the mounts after already mounting, e.g. unmounting a resource and then adding files to the resource's directory, then this will get thrown on remount.

In either case, we expect that there are no local files`~/workspace` and I think we should bring the user's attention to them if there are.

src/main/java/bio/terra/cli/command/resource/ListTree.java
rogerwangcs: Sg.

src/main/java/bio/terra/cli/businessobject/resource/GcsObject.java
rogerwangcs: I had to also pass the `BlobStorageOption.currentDirectory()` option to`list` for it to behave as intended. Thank you!

src/main/java/bio/terra/cli/utils/mount/MountController.java
```diff
@@ -0,0 +1,285 @@
...
+import java.util.UUID;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Stream;
+
+/** This class provides utility methods for mounting and unmount workspace resources */
+public abstract class MountController {
+
+  // Directory to mount workspace resources under
+  private static final Path WORKSPACE_DIR = Paths.get(System.getProperty("user.home"), "workspace");
```
zloery, line 37: Use `Context.getContextDir()` as the base to respect the context dir override env var. That will also add `.terra` to the path so we're not working directly in `~/workspace`

src/main/java/bio/terra/cli/utils/OSFamily.java
```diff
@@ -0,0 +1,27 @@
...
+  public static final String LINUX = "linux";
+  public static final String WINDOWS = "windows";
+  public static final String UNKNOWN = "unknown";
+
+  /**
+   * Returns the OS family.
+   *
+   * @return the OS family string
+   */
+  public static String getOSFamily() {
```
zloery, line 15: This is currently unused, should MountControllerFactory call this?

src/main/java/bio/terra/cli/utils/mount/LinuxMountController.java
```diff
@@ -0,0 +1,10 @@
+package bio.terra.cli.utils.mount;
+
+import java.util.regex.Pattern;
+
+public class LinuxMountController extends MountController {
+
+  protected Pattern getMountEntryPattern() {
+    return Pattern.compile("^(\\S+)\\s+on\\s+([^\\(]+)\\s+(type[^\\)]++\\))");
```
zloery, line 8: Can you add an explanation or example of what should/shouldn't match here? I'm having a hard time parsing this

src/test/java/harness/baseclasses/ClearContextUnit.java
```diff
@@ -22,7 +22,8 @@
  */
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
 public class ClearContextUnit {
-  protected static final TestUser workspaceCreator = TestUser.chooseTestUserWithSpendAccess();
+  // protected static final TestUser workspaceCreator = TestUser.chooseTestUserWithSpendAccess();
+  protected static final TestUser workspaceCreator = TestUser.chooseTestUserWithOwnerAccess();
```
zloery, line 26: out of curiosity, why this change?

src/main/java/bio/terra/cli/utils/mount/handlers/BaseMountHandler.java
```diff
@@ -0,0 +1,122 @@
...
+    addErrorStateToMountPoint("_MOUNT_FAILED");
+  }
+
+  /**
+   * Adds an error state to the mount point.This is done by renaming the mount point to
+   * mountPoint_errorString.
+   *
+   * @param errorString the error string to add to the mount point
+   */
+  private void addErrorStateToMountPoint(String errorString) {
```
zloery, line 112: Why move the mount point instead of just throwing/logging an appropriate error?

src/main/java/bio/terra/cli/businessobject/resource/GcsObject.java
```diff
@@ -125,6 +131,26 @@ public String resolve(boolean includeUrlPrefix) {
...
+  public boolean isDirectory() throws SystemException {
+    try {
+      BucketCow bucketCow =
+          CrlUtils.createStorageCow(Context.requireUser().getPetSACredentials()).get(bucketName);
+      Iterable<BlobCow> objects =
+          bucketCow
+              .list(BlobListOption.currentDirectory(), BlobListOption.prefix(objectName))
+              .getValues();
+      return StreamSupport.stream(objects.spliterator(), false)
+          .anyMatch(object -> object.getBlobInfo().isDirectory());
```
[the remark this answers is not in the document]
rogerwangcs: Oh thanks! It would behave the same that way.

So I went with listing the objects with the prefix because I needed to do something similar in the UI to handle a rare edge case where a directory created manually in gcs does not get recognized as a directory. Just tested that case and looks like I wasn't handling it anyways. Making a change to handle nonempty user created directories. It's not possible to handle empty folders created by users since its ambigious given what we have (false for isDirectory, 1 object in the list which is the same for any regular file objects).

src/main/java/bio/terra/cli/utils/mount/handlers/BaseMountHandler.java
rogerwangcs: The main use case for this command is for it to run in the post startup script. We make users aware of errors that occur with their mounts by appending strings to the mountpoint directories. Otherwise users would have to dig into the logs.

src/main/java/bio/terra/cli/utils/mount/MountControllerFactory.java
```diff
@@ -0,0 +1,16 @@
...
+
+public class MountControllerFactory {
+  public static MountController getMountController() {
+    String os = OSFamily.getOSFamily();
+    if (os.equals(OSFamily.MAC)) {
+      return new MacMountController();
+    } else if (os.equals(OSFamily.LINUX)) {
+      return new LinuxMountController();
+    } else {
+      throw new UnsupportedOperationException("Unsupported operating system.");
```
yuhuyoyo, line 13: this can't work on window?

rogerwangcs:
@yuhuyoyo @zloery Added unit tests for `MountController` `mountResources()` and `unmountResources()` and for `BaseMountHandler`. For the time being, I moved these mockito unit tests into a separate test tag `real-unit`. Thoughts on this decision?

src/main/java/bio/terra/cli/utils/FileUtils.java
```diff
@@ -63,4 +67,65 @@ public static Path writeStringToFile(File outputFile, @Nullable String fileConte
 
     return Files.writeString(outputFile.toPath(), fileContents == null ? "" : fileContents);
   }
+
```
rogerwangcs, line 70: This method is moved from `TestContext` to here so it can be shared with `MountController`

src/main/java/bio/terra/cli/command/resource/Mount.java
```diff
@@ -0,0 +1,33 @@
...
+
+/** This class corresponds to the "terra resource mount" command. */
+@Command(name = "mount", description = "Mounts workspace resources.")
+public class Mount extends BaseCommand {
+
+  @CommandLine.Mixin WorkspaceOverride workspaceOption;
+
+  @CommandLine.Option(
+      names = "--disable-cache",
+      description = "Disable cache",
```
yuhuyoyo, line 18: need more description here. what does it mean to disable cache?

src/main/java/bio/terra/cli/utils/mount/handlers/GcsFuseMountHandler.java
```diff
@@ -0,0 +1,77 @@
...
+    super(mountPoint, disableCache);
+    this.bucketName = gcsObject.getBucketName();
+    this.subDir = gcsObject.getObjectName();
+  }
+
+  /** Implements the mount method for a GCS bucket or prefix object. */
+  public void mount() throws SystemException {
+    // Build mount command
+    List<String> command = new ArrayList<>(List.of(FUSE_MOUNT_COMMAND));
+    if (disableCache) {
```
yuhuyoyo, line 34: add comment about what they are

src/main/java/bio/terra/cli/utils/mount/handlers/BaseMountHandler.java
```diff
@@ -0,0 +1,132 @@
...
+   * Appends an error string to the mount point directory. This is done by renaming the mount point
+   * to mountPoint_errorString.
+   *
+   * @param errorString the error string.
+   */
+  private void addErrorStateToMountPoint(String errorString) {
+    // Add error state to mount point
+    Path newPath = mountPoint.resolveSibling(mountPoint.toAbsolutePath() + errorString);
+    try {
+      Files.move(mountPoint, newPath);
```
yuhuyoyo, line 126: what are you moving here?

src/test/java/unit/MountControllerTest.java
```diff
@@ -0,0 +1,171 @@
...
+  /** Utility method to check if a directory exists */
+  public void verifyDirectory(Path path) {
+    File file = path.toFile();
+    assert (file.exists() && file.isDirectory());
+  }
+
+  @Test
+  @DisplayName("mountController mounts buckets on linux")
+  void testMountResources() {
+    MockedStatic<Context> mockStaticContext = mockStatic(Context.class);
```
yuhuyoyo, line 56: https://www.baeldung.com/mockito-mock-static-methods#mocking-a-no-argument-static-method

yuhuyoyo (commented):
looks fine in general, not familiar with mounting myself.

could you also attach an example output in the description?

yuhuyoyo (approved):

src/main/java/bio/terra/cli/utils/mount/handlers/GcsFuseMountHandler.java
rogerwangcs: Adding comment to `@CommandLine.Option`

src/main/java/bio/terra/cli/utils/mount/handlers/BaseMountHandler.java
rogerwangcs: This is just renaming the directory. Let me rename variables to make it more clear.

src/test/java/unit/MountControllerTest.java
rogerwangcs: qq: Is it okay to do this instead of try-with-resource as long as I close the mock at the end of the test? My only reason is to avoid unneeded indentation though.

src/main/java/bio/terra/cli/utils/mount/MountController.java
zloery: Got it. I'm a little wary about writing to `~/workspace` which doesn't have any indication it's associated with Terra, but if we're generally expecting people to use this in notebooks rather than locally, that's less of a concern

src/test/java/unit/MountControllerTest.java
```diff
@@ -0,0 +1,171 @@
+package unit;
+
+import static org.mockito.Mockito.*;
```
zloery, line 3: For Terra repos, we avoid using * imports for clarity. If intellij did that automatically, there's a setting to stop it from doing so (something like "Class count to use import with '*'" to 99

src/test/java/unit/MountControllerTest.java
zloery: It's generally better to use a try-with-resources in case the test exits before getting to the end (e.g. due to a failed assertion)

zloery (approved):

src/test/java/unit/MountControllerTest.java
rogerwangcs: Oh I see.

src/main/java/bio/terra/cli/businessobject/resource/GcsObject.java
```diff
@@ -125,6 +133,35 @@ public String resolve(boolean includeUrlPrefix) {
...
+   * itself.
+   */
+  public boolean isDirectory() throws SystemException {
+    try {
+      // get all objects with prefix of objectName
+      BucketCow bucketCow =
+          CrlUtils.createStorageCow(Context.requireUser().getPetSACredentials()).get(bucketName);
+      Iterable<BlobCow> objects =
+          bucketCow
+              .list(BlobListOption.currentDirectory(), BlobListOption.prefix(objectName))
```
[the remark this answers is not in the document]
rogerwangcs: I didn't consider that. Added.

.github/workflows/tests-on-pr-push-and-merge.yml
```diff
@@ -30,7 +30,7 @@ jobs:
   tests-against-source-code:
     strategy:
       matrix:
-        testTag: [ "unit", "integration" ]
+        testTag: [ "unit", "integration", "real-unit" ]
```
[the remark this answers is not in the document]
rogerwangcs: I see. Combining into "unit".

src/main/java/bio/terra/cli/command/resource/Unmount.java
```diff
@@ -0,0 +1,22 @@
...
+import bio.terra.cli.utils.mount.MountControllerFactory;
+import picocli.CommandLine.Command;
+
+/** This class corresponds to the "terra resource unmount" command. */
+@Command(name = "unmount", description = "Unmounts workspace resources.")
+public class Unmount extends BaseCommand {
+
+  @Override
+  protected void execute() {
+    MountController mountController = MountControllerFactory.getMountController();
```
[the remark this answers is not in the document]
rogerwangcs: Yeah thats a good callout. I'll flag this behavior on the doc.

PR #461, opened 2023-04-21
[TERRA-495], [PF-2631] Remove client_id and client_secret

dexamundsen:
- Remove client_id and client_secret from secrets file
- Fetch these values from env during release publish and insert into secrets file

```
# export variables BROAD_CLIENT_ID BROAD_CLIENT_SECRET VERILY_CLIENT_ID VERILY_CLIENT_SECRET
>>  ./tools/client-credentials.sh "src/main/resources/broad_secret.json" $BROAD_CLIENT_ID $BROAD_CLIENT_SECRET
# verified that files have the client id / secret updated

>> terra-cli (dexamundsen/terra-495) >> ./tools/publish-release.sh dex-test-release
# pre-release created successfully
Downloaded the prerelease and verified that json files include client credentials.

>> ./tools/render-config
# verified files have the client id / secret updated
```

- Credentials are needed in PR tests, a successful PR check implies this is reflected in CI/CD runs

### DIFF
--- a/src/main/resources/verily_secret.json
+++ b/src/main/resources/verily_secret.json
@@ -6,8 +6,6 @@
     "auth_provider_x509_cert_url": "https://www.googleapis.com/oauth2/v1/certs",
     "redirect_uris": [
       "https://terra.verily.com/authcode"
-    ],
-    "client_id": "590500654757-l8ripukp65bpdr7kc7nq0o4bv8kmtku0.apps.googleusercontent.com",
-    "client_secret": "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"
+    ]
   }
 }
--- a/src/main/resources/verily_secret.json
+++ b/src/main/resources/verily_secret.json
@@ -1,11 +1,9 @@
 {
   "installed": {
-    "client_id": "590500654757-l8ripukp65bpdr7kc7nq0o4bv8kmtku0.apps.googleusercontent.com",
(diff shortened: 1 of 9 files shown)


src/main/resources/verily_secret.json
```diff
@@ -1,11 +1,11 @@
 {
   "installed": {
-    "client_id": "590500654757-l8ripukp65bpdr7kc7nq0o4bv8kmtku0.apps.googleusercontent.com",
     "project_id": "terra-ots0-prod",
     "auth_uri": "https://accounts.google.com/o/oauth2/auth",
     "token_uri": "https://oauth2.googleapis.com/token",
     "auth_provider_x509_cert_url": "https://www.googleapis.com/oauth2/v1/certs",
-    "client_secret": "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",
```
jwtracy, line 8: Does this need to be removed from Git history too or are you going to rotate the creds?

jwtracy (approved):

zloery (approved):

src/main/resources/verily_secret.json
dexamundsen: Ideally this needs to be rotated, however rotating ti will break existing installations. There is a long term plan however, sharing it offline

dexamundsen:
Downloaded the latest version and verified that client Id was passed in the auth call.

PR #468, opened 2023-04-28
[PF-2677] Add mount individual resources with the `--name` flag

rogerwangcs:
Last PR for the `terra resource mount` command. Allows for mounting and unmounting single resources by specifying the resource name. Can be combined with existing `--read-only` and `disable-cache` flags.

Examples:
```
❯ terra resource mount --name=clinvar
Successfully mounted resource clinvar.
❯ ls ~/workspace/clinvar
README.txt             disease_names.txt      variant_summary.txt    variant_summary.txt.gz
```
```
❯ terra resource unmount --name=clinvar
Successfully unmounted resource clinvar.
❯ terra resource unmount --name=clinvar
Resource clinvar is not mounted.
```
```
❯ terra resource mount --name=private-bucket-reference
Failed to mount resource private-bucket-reference
```


src/main/java/bio/terra/cli/command/resource/Mount.java
```diff
@@ -28,28 +29,46 @@ public class Mount extends BaseCommand {
   private Boolean disableCache;
 
   /**
-   * For controlled resources created by the calling user, the mount is always read-write regardless
-   * of the value of this flag.
+   * Whether to mount the resource(s) as read-only or read-write, overriding default mount
+   * permissions.
+   *
+   * <p>By default, for controlled resources created by the calling user, the mount is always
```
zloery, line 35: "regardless of the value of this flag" makes it sound like this is ignored for controlled resources creating by the calling user. If I mount a resource I've created with `--read-only=true`, will that actually mount it read-write?

src/main/java/bio/terra/cli/utils/mount/MountController.java
```diff
@@ -96,9 +100,90 @@ public int mountResources(boolean disableCache, @Nullable Boolean readOnly) {
         .sum();
   }
 
+  public void mountResource(String resourceName, boolean disableCache, @Nullable Boolean readOnly) {
```
zloery, line 103: This method is doing a lot of the same logic as `mountResources` above, can they be combined? Either having `mountResources` call `mountResource` for each resource, using a shared `mountResourceWorker` method to handle the implementation, or somethine else?

src/main/java/bio/terra/cli/command/resource/Mount.java
rogerwangcs: This contradicts the top line and does correctly describe the actual behavior. `--read-only=true` will mount with only read perm. Updating the comment.

src/main/java/bio/terra/cli/utils/mount/MountController.java
rogerwangcs: Done!

zloery (approved):

yuhuyoyo (approved):
oops forgot to approve

PR #474, opened 2023-05-11
[TERRA-320], [TERRA-565] AWS sagemaker support on CLI

dexamundsen:
- Add support for AWS sagemaker notebook create, delete, credential, start, stop
- rename Type.S3_STORAGE_FOLDER to Type.AWS_S3_STORAGE_FOLDER
- Minor comment / document updates
- [TERRA-565] test to create, delete, start, stop, getCredentials, resolve

Testing
```
terra-cli (dexamundsen/320-create) >> terra resource create sagemaker-notebook --name=dex-note-512-1 --instance-name=dex-note-512-1-dex-ws-512-1
CLI timed out waiting for the job to complete. It's still running on the server.

terra-cli (dexamundsen/320-create) >> terra resource list 
NAME                            RESOURCE TYPE         STEWARDSHIP TYPE      DESCRIPTION                             
dex-note-512-1                  AWS_SAGEMAKER_NOT...  CONTROLLED            (unset)                                 
dex-note-512-2                  AWS_SAGEMAKER_NOT...  CONTROLLED            (unset)                                 
dex-note-512-3                  AWS_SAGEMAKER_NOT...  CONTROLLED            (unset)                                 
dex-s3-512-1                    S3_STORAGE_FOLDER     CONTROLLED            (unset)  

terra-cli (dexamundsen/320-create) >> terra resource describe --name=dex-note-512-3
Name:         dex-note-512-3
Description:  
Type:         AWS_SAGEMAKER_NOTEBOOK
Stewardship:  CONTROLLED
Cloning:      COPY_RESOURCE
Access scope: PRIVATE_ACCESS
Managed by:   USER
Region:       us-east-1
Private user: dexamundsen@google.com
Properties:   class Properties {
    []
}
SageMaker Notebook instance name:   dex-note-512-3-name
SageMaker Notebook instance type:   ml.t2.medium
SageMaker Notebook instance status: InService

terra-cli (dexamundsen/320-create) >> terra resource credentials --name=dex-note-512-3
class AwsCredential {
    version: 1
    accessKeyId: …
    secretAccessKey: …
    sessionToken: …
    expiration: 2023-05-12T19:19:47Z
}

terra-cli (dexamundsen/320-create) >> terra notebook start --name=dex-s3-512-1
Only able to use notebook commands on AWS SageMaker Notebook resources, but specified resource is S3_STORAGE_FOLDER

terra-cli (dexamundsen/320-create) >> terra notebook start --name=dex-note-512-3
Notebook instance started  # already started

[...]

PR #475, opened 2023-05-12
add AWS prefix to S3 type, Set resource type in businessObject.resources

dexamundsen:
- Rename Type.S3_STORAGE_FOLDER to Type.AWS_S3_STORAGE_FOLDER  (in line with WSM definition)
- Minor comment and doc updates
- Set resource type in businessObject.resources
- Upgrade WSM client


src/main/java/bio/terra/cli/service/WorkspaceManagerServiceAws.java
```diff
@@ -50,6 +50,56 @@ public static WorkspaceManagerServiceAws fromContext() {
...
+        new DeleteControlledAwsResourceRequestBody().jobControl(new JobControl().id(asyncJobId));
+
+    handleClientExceptions(
+        () -> {
+          ControlledAwsResourceApi controlledAwsResourceApi =
+              new ControlledAwsResourceApi(apiClient);
+          // make the initial delete request
+          HttpUtils.callWithRetries(
+              () ->
+                  controlledAwsResourceApi.deleteAwsS3StorageFolder(
```
rogerwangcs, line 65: Should this be checking resource type before calling this? This common method will only work with s3 storage folders as is right? Or are we adding cases for other resources as we add them?

dexamundsen: It will be used with sagemaker notebooks as well part of https://github.com/DataBiosphere/terra-cli/pull/474

dexamundsen: Reverted this change. 
initial plan was to use a switch case and compare the resource type (similar to implementation in WSM). 
removed it since it added additional look ups and provided minimal benefit

rogerwangcs (approved):

zloery (approved):

src/main/java/bio/terra/cli/businessobject/resource/GcsBucket.java
```diff
@@ -30,6 +30,7 @@ public class GcsBucket extends Resource {
   /** Deserialize an instance of the disk format to the internal object. */
   public GcsBucket(PDGcsBucket configFromDisk) {
     super(configFromDisk);
+    this.resourceType = Type.GCS_BUCKET;
```
yuhuyoyo, line 33: is this change related?

dexamundsen: This was added to the description, not the title since the title was already long earlier. Adding it now.

dexamundsen:
Errors in PR tests are related to WSM issues. All tests are connected, hence a successful test requires stable WSM env. WSM fixes are merged, but not deployed to test env. Hence force merging this, needed for subsequent PR

PR #491, opened 2023-05-24
Mb/shell wrap faster

mbookman:
Addressing https://broadworkbench.atlassian.net/browse/PF-2804.

With this change, commands execute about 3 seconds faster.

Previous:

```
$ time terra app execute gcloud compute instances list
Setting the gcloud project to the workspace project: terra-vdevel-<snip>
Updated property [core/project].
<snip>
Restoring the original gcloud project configuration: <snip>
Updated property [core/project].

real	0m8.986s
user	0m12.100s
sys	0m1.140s
```

with this change:

```
$ time terra app execute gcloud compute instances list
<snip>
real	0m5.926s
user	0m9.062s
sys	0m0.693s
```

### DIFF
--- a/src/main/java/bio/terra/cli/app/LocalProcessCommandRunner.java
+++ b/src/main/java/bio/terra/cli/app/LocalProcessCommandRunner.java
@@ -36,15 +36,36 @@ public class LocalProcessCommandRunner extends CommandRunner {
     boolean isGcpWorkspace =
         CloudPlatform.GCP.equals(Context.requireWorkspace().getCloudPlatform());
 
+    String cmdPre = "";
+    String cmdPost = "";
+
     if (isGcpWorkspace) {
-      // For the wrapped command, set the CLOUDSDK_CORE_PROJECT environment
-      // variable, which is respected by gcloud, gsutil, and bq.
-      // See https://cloud.google.com/sdk/docs/configurations
-      bashCommands.add(String.format(
-        "export CLOUDSDK_CORE_PROJECT='%s'", 
-        Context.requireWorkspace().getRequiredGoogleProjectId()));
+      cmdPre = String.format("""
+        TERRA_GCLOUD_PROJECT='%s'
+        """, Context.requireWorkspace().getRequiredGoogleProjectId()) +
+        """
+        TERRA_PREV_GCLOUD_PROJECT="$(gcloud config get-value project)"
+        if [[ "${TERRA_PREV_GCLOUD_PROJECT}" != "${TERRA_GCLOUD_PROJECT}" ]]; then
+          echo "Setting the gcloud project to the workspace project: ${TERRA_GCLOUD_PROJECT}"
+          gcloud config set project "${TERRA_GCLOUD_PROJECT}"
+        fi
+        """;
+
+      cmdPost = """
+
+        if [[ -z "${TERRA_PREV_GCLOUD_PROJECT}" ]]; then
+          echo "Restoring the original gcloud project configuration: (unset)"
+          gcloud config unset project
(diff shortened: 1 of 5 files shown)


src/main/java/bio/terra/cli/app/LocalProcessCommandRunner.java
```diff
@@ -35,20 +35,39 @@ protected String wrapCommandInSetupCleanup(List<String> command) {
...
-      bashCommands.add(
-          "gcloud config set project " + Context.requireWorkspace().getRequiredGoogleProjectId());
+      cmdPre = String.format("""
+        TERRA_GCLOUD_PROJECT='%s'
+        """, Context.requireWorkspace().getRequiredGoogleProjectId()) +
+        """
+        TERRA_PREV_GCLOUD_PROJECT="$(gcloud config get-value project)"
+        if [[ "${TERRA_PREV_GCLOUD_PROJECT}" != "${TERRA_GCLOUD_PROJECT}" ]]; then
+          echo "Setting the gcloud project to the workspace project: ${TERRA_GCLOUD_PROJECT}"
+          gcloud config set project "${TERRA_GCLOUD_PROJECT}"
```
cbookg, line 50: this looks fine.  but instead of modifying the config, can we not set this project via environment variable or gcloud flag?

i dislike how we mess with the user's config.

mbookman: @cbookg this is a great point. Have never liked this but had assumed this had been previously well researched. It turns out that we have a few choices as noted here:

https://cloud.google.com/sdk/docs/configurations

A broad solution we could look at is to set up and manage one or more terra-specific `gcloud` configuration directories. But in the short-term, we can simply use the environment variable `CLOUDSDK_CORE_PROJECT`. I just did some testing and validated that `gcloud` and `gsutil` respect this value.

I think this would be a pretty robust solution (and obviously better than what the cli current does). I was a bit concerned about the case where a tool that someone is calling in there `terra` subcommand picks up the project *from gcloud*, but this case looks consistently handled:

```
$ CLOUDSDK_CORE_PROJECT=terra-luminous-olive-1234 gcloud config list
[core]
account = <snip>
disable_usage_reporting = True
project = terra-luminous-olive-1234
```

```
$ CLOUDSDK_CORE_PROJECT=terra-luminous-olive-1234 gcloud config get project
terra-luminous-olive-1234
```
```
$ CLOUDSDK_CORE_PROJECT=terra-vdevel-windy-radish-1234 gcloud config get project
terra-vdevel-windy-radish-1234
```

Will change to using the environment variable.

Thanks for the prod!

src/main/resources/broad_secret.json
```diff
@@ -7,6 +7,8 @@
     "redirect_uris": [
       "urn:ietf:wg:oauth:2.0:oob",
       "http://localhost"
-    ]
+    ],
+    "client_id": "436986732241-nvjfkarirkhmcq7m3r1uemdau518uvdo.apps.googleusercontent.com",
```
cbookg, line 11: i assume you didn't mean to add these?

src/main/java/bio/terra/cli/app/LocalProcessCommandRunner.java
cbookg: amazing, so much better!

src/main/resources/broad_secret.json
dexamundsen: +1 
please revert changes to broad_secret & verily_secret files
I have https://verily.atlassian.net/browse/BENCH-530 to address this

cbookg (approved):

src/main/resources/broad_secret.json
mbookman: Thanks! Reverted and updated my branch with the merge from main.

dexamundsen (approved):

PR #499, opened 2023-06-01
Fix the render_config.sh

yuhuyoyo:
$3 base 64 is optional.


dexamundsen:
fixed in https://github.com/DataBiosphere/terra-cli/pull/497

PR #513, opened 2023-06-19
[PF-2837] Remove access and cloning flags for gcp-notebook creation.

mbookman:
`--cloning` and `--access` are available in the CLI for gcp-notebook creation, but are (or should be) ignored.

This PR removes them from the CLI.

Also note that this changes the default cloning instruction passed for notebook instances from `RESOURCE` to `NOTHING`. If we want consistency, I can change this to `RESOURCE`, but `NOTHING` seems to be the "correct" value.


dexamundsen (approved):
This is the case for AWS notebooks as well, I will update it in a separate PR

ddietterich (approved):

PR #516, opened 2023-06-26
[PF-2783] Add read only file permission bits

rogerwangcs:
Currently, when a user mounts a bucket as read only and runs `ls -l` inside the mounted bucket, the files will have incorrect permission bits set. The default 644 (rx-r--r--) set by `gscfuse`. @mbookman proposed using the `--file-mode` and `--dir-mode` flags in `gcsfuse` to set the correct read only file permissions on files and directories. This PR does just that!


zloery (approved):
I'm not familiar with gcsfuse so I'll defer to Matt, but this seems like a reasonable change

mbookman (approved):
LGTM.

Thanks Roger!

rogerwangcs:
Test failures are related to the rbs workspace creation issue. Merging as it is unrelated to the changes of this PR.

PR #522, opened 2023-07-20
[PF-2924] update gcloud version in terra-cli dockerfile

amygdala:
see https://broadworkbench.atlassian.net/browse/PF-2924.  
The older pinned version broke (at least) the `terra gcloud artifacts repositories create`  command.

(We may want to consider just grabbing the 'latest' but I didn't do so in this PR).

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -17,7 +17,7 @@ RUN apt-get update \
 # other examples of installing gcloud in a docker image:
 #   - AoU (https://github.com/all-of-us/workbench/blob/master/api/src/dev/server/Dockerfile#L34)
 #   - Terra Jupyter (https://github.com/DataBiosphere/terra-docker/blob/master/terra-jupyter-base/Dockerfile#L74)
-ENV CLOUD_SDK_VERSION 411.0.0
+ENV CLOUD_SDK_VERSION 439.0.0
 ENV PATH /usr/local/google-cloud-sdk/bin:$PATH
 RUN curl -O https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-${CLOUD_SDK_VERSION}-linux-x86_64.tar.gz && \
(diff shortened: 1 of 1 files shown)


amygdala:
I believe the build errors, which are: `Error: Input required and not supplied: token`
are unrelated to this small change.

docker/Dockerfile
```diff
@@ -17,7 +17,7 @@ RUN apt-get update \
 # other examples of installing gcloud in a docker image:
 #   - AoU (https://github.com/all-of-us/workbench/blob/master/api/src/dev/server/Dockerfile#L34)
 #   - Terra Jupyter (https://github.com/DataBiosphere/terra-docker/blob/master/terra-jupyter-base/Dockerfile#L74)
-ENV CLOUD_SDK_VERSION 411.0.0
+ENV CLOUD_SDK_VERSION 439.0.0
```
yuhuyoyo, line 20: thanks for getting to it so quickly Amy. I vote we just grab the latest.

amygdala: @yuhuyoyo now that I suggested that, I can't see how to do it.  AFAICT there is no download tagged 'latest'. Which probably explains why the version was pinned to start with.  (Do you know?).

yuhuyoyo: ah bummer. it does seem that it needs to be specified according to https://cloud.google.com/sdk/docs/install.

yuhuyoyo (approved):

docker/Dockerfile
mbookman: I prefer an approach of pulling a specific version and then filing a Jira ticket to update every <N> months.
We had issues around CloudSDK in particular periodically breaking dsub/tests until we stopped getting latest.

mbookman (approved):
LGTM

amygdala:
closing this PR as I think the test failures are due to the tests trying to run in my fork.  Instead created https://github.com/DataBiosphere/terra-cli/pull/523.

PR #526, opened 2023-08-01
[BENCH-780] Change resource creation time out to 60 min

dexamundsen:
Similar change was made in WSM - https://github.com/DataBiosphere/terra-workspace-manager/pull/1390

Without this change, AWS notebooks that take 30+ mins for creation will fail in CLI tests (creation + cleanup) and remain as stale resources


mikenev (approved):
Wow, 60 min. That's a long time for tests

dexamundsen:
> Wow, 60 min. That's a long time for tests

Agree. Unfortunately, AWS notebook creation takes 45+ mins at a time. This is a short-term fix to avoid stale resources.

zloery (approved):

PR #529, opened 2023-08-15
[PF-2865] Add dataproc cluster resource

rogerwangcs:
Adds dataproc cluster resource create, delete, resolve. Also adds top level command `cluster` with the following subcommands `start`, `stop`, and `launch`. See README additions for details on the new commands

The unit test in this change can only be executed in the downstream verily environment due to the added dependency on axon server. I added the test tag `unit-verily-gcp` to the test. The idea is that It won't run in this repo's workflows as it sets the test platform to `gcp` but will run in the verily test environment as it runs all unit tests tagged with `unit`.

Followup: resource update.


src/main/java/bio/terra/cli/command/cluster/Launch.java
```diff
@@ -0,0 +1,57 @@
...
+import bio.terra.cli.command.shared.options.WorkspaceOverride;
+import bio.terra.cli.exception.UserActionableException;
+import bio.terra.cli.service.AxonServerService;
+import bio.terra.workspace.model.CloudPlatform;
+import java.util.UUID;
+import org.json.JSONObject;
+import picocli.CommandLine;
+
+/** This class corresponds to the third-level "terra cluster launch" command. */
+@CommandLine.Command(
```
yuhuyoyo, line 18: is dataproc launched?
if not, flag-guard these commands thanks.

i added a feature class [here](https://github.com/DataBiosphere/terra-cli/blob/d0473814b032b475f8f4cc0183f12da8b276c042/src/main/java/bio/terra/cli/service/FeatureService.java#L18).

src/main/java/bio/terra/cli/service/AxonServerService.java
```diff
@@ -0,0 +1,205 @@
...
+  }
+
+  /**
+   * Call the Axon Server GET
+   * /api/workspaces/v1/{workspaceId}/resources/{resourceId}/gcp/dataproccluster/start:
+   *
+   * @param workspaceId the workspace that contains the resource
+   * @param resourceId the cluster resource id
+   */
+  public void putStartCluster(UUID workspaceId, UUID resourceId) {
```
yuhuyoyo, line 94: super nit: can they just be called startCluster and stopCluster

src/main/java/bio/terra/cli/service/WorkspaceManagerServiceGcp.java
```diff
@@ -145,6 +157,58 @@ private static GcpAiNotebookInstanceCreationParameters fromCLIObject(
...
   }
 
+  /**
+   * This method converts this CLI-defined POJO class into the WSM client library-defined request
+   * object.
+   *
+   * @return GCP Dataproc cluster creation parameters in the format expected by the WSM client
+   *     library
+   */
+  private static GcpDataprocClusterCreationParameters fromCLIObject(
```
yuhuyoyo, line 167: can you put this method in the CreateGcpDataprocClusterParams class and make it a toWsmCreationParams

src/main/java/bio/terra/cli/service/WorkspaceManagerServiceGcp.java
```diff
@@ -145,6 +157,58 @@ private static GcpAiNotebookInstanceCreationParameters fromCLIObject(
...
+        .autoscalingPolicy(createParams.autoscalingPolicy)
+        .metadata(createParams.metadata)
+        .managerNodeConfig(buildInstanceGroupConfig(createParams.managerConfig))
+        .primaryWorkerConfig(buildInstanceGroupConfig(createParams.workerConfig))
+        .secondaryWorkerConfig(buildInstanceGroupConfig(createParams.secondaryWorkerConfig))
+        .lifecycleConfig(
+            new GcpDataprocClusterLifecycleConfig().idleDeleteTtl(createParams.idleDeleteTtl));
+  }
+
+  private static GcpDataprocClusterInstanceGroupConfig buildInstanceGroupConfig(
```
yuhuyoyo, line 188: ditto

src/test/java/unit/GcpDataprocClusterControlled.java
```diff
@@ -0,0 +1,213 @@
...
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+/** Tests for the `terra resource` commands that handle controlled GCP clusters. */
+/**
+ * This test only executes on verily environments. Add a tag prefix to prevent it from running in
+ * the main repo
+ */
+@Tag("unit-verily-gcp")
+public class GcpDataprocClusterControlled extends SingleWorkspaceUnitGcp {
```
yuhuyoyo, line 34: can you test launch or no?

src/test/java/unit/GcpDataprocClusterControlled.java
```diff
@@ -0,0 +1,213 @@
...
+            "--metadata=foo=bar",
+            "--idle-delete-ttl=1800s");
+
+    GcpDataprocClusterUtils.pollDescribeForClusterState(name, "RUNNING");
+  }
+
+  @Test
+  @DisplayName("list and describe reflect creating a controlled cluster")
+  void listDescribeReflectCreate() throws IOException {
+    workspaceCreator.login();
```
yuhuyoyo, line 89: i think you should already be logged in since you are extending the singleworkspaceunit

src/main/java/bio/terra/cli/command/resource/create/GcpDataprocCluster.java
```diff
@@ -0,0 +1,454 @@
...
+      names = "--image-version",
+      description =
+          "The dataproc cluster image version containing versions of its software components. See https://cloud.google.com/dataproc/docs/concepts/versioning/dataproc-version-clusters for the full list of image versions and their bundled software components.")
+  private String imageVersion;
+
+  @CommandLine.Option(
+      names = "--initialization-actions",
+      split = ",",
+      description =
+          "A comma separated list of initialization scripts to run during cluster creation."
```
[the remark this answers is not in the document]
rogerwangcs: Yup! They run serially in the order that they are passed in.

src/main/java/bio/terra/cli/command/resource/create/GcpDataprocCluster.java
```diff
@@ -0,0 +1,454 @@
...
+import java.util.Map;
+import java.util.Optional;
+import java.util.UUID;
+import picocli.CommandLine;
+
+/**
+ * This class corresponds to the fourth-level "terra resource create gcp-dataproc-cluster" command.
+ */
+@CommandLine.Command(
+    name = "gcp-dataproc-cluster",
```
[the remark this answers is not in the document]
rogerwangcs: It was because vertex ai notebooks were `gcp-notebook`. I can remove the prefix since dataproc kind of already is the prefix. We miight have to support a similar managed spark cluster resource in aws.

src/main/java/bio/terra/cli/command/cluster/Launch.java
```diff
@@ -0,0 +1,57 @@
...
+    name = "launch",
+    description = "Launch a running Dataproc cluster within your workspace.",
```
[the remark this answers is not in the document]
rogerwangcs: Will make this clearer that it launches the proxy view to a dataproc optional component, including jupyter. Fyi: These command descriptions will go through a UXW pass and will be updated with their suggestions.

src/main/java/bio/terra/cli/service/AxonServerService.java
rogerwangcs: Done.

src/main/java/bio/terra/cli/service/WorkspaceManagerServiceGcp.java
rogerwangcs: Yeah good idea. Also am curious, why we have this pattern of making an intermediary create/update params obj instead of directly mapping the command class attributes to a wsm create/request body?

src/test/java/unit/GcpDataprocClusterControlled.java
rogerwangcs: Done

yuhuyoyo (approved):

src/main/java/bio/terra/cli/command/resource/Delete.java
```diff
@@ -22,6 +23,10 @@ protected void execute() {
 
     Resource resourceToDelete = Context.requireWorkspace().getResource(resourceNameOption.name);
 
+    if (resourceToDelete.getResourceType().equals(Resource.Type.DATAPROC_CLUSTER)) {
+      CommandUtils.checkDataprocSupport();
```
dexamundsen, line 27: Flag - vwb__cli_token_refresh_enabled seems to be targetted only for CLI. Is it possible to create a data proc cluster on UI, and then be unable to access it on CLI? 
This might not be a blocker if behavior is expected and informed to solutions team. 

 further -
This check prevents all operations on the cluster, but does not hide it. 
Delete is not allowed even if the cluster was created on UI. 
what is the expected behavior when user tries to delete the workspace? will it fail?

src/main/java/bio/terra/cli/command/resource/CheckAccess.java
```diff
@@ -25,6 +26,9 @@ public class CheckAccess extends WsmBaseCommand {
   protected void execute() {
     workspaceOption.overrideIfSpecified();
     Resource resource = Context.requireWorkspace().getResource(resourceNameOption.name);
+    if (resource.getResourceType().equals(Resource.Type.DATAPROC_CLUSTER)) {
```
dexamundsen, line 29: Do we need this check here? The resource is visible in display / list resources. 
check access might only read sam. 

overall behavior is a little inconsistent wrt flags. 
Are the flags intended to be removed soon? if so, please add a ticket to track it. 
If the flag check is meant to be retained for medium - long term, this needs to be updated. 

overall --
1) if data proc is not supported,. should list resources display them? 
2) is flag=false, operations are not allowed. Some like check-access may not need this feature-flagcheck. However the same operation may be allowed via Workspace operation (example delete)

I would be open to handling it in a separate PR if the above needs discussion / complex changes

src/main/java/bio/terra/cli/command/resource/Describe.java
```diff
@@ -21,6 +22,9 @@ public class Describe extends WsmBaseCommand {
   protected void execute() {
     workspaceOption.overrideIfSpecified();
     Resource resource = Context.requireWorkspace().getResource(resourceNameOption.name);
+    if (resource.getResourceType().equals(Resource.Type.DATAPROC_CLUSTER)) {
```
dexamundsen, line 25: is it the same behavior with workspace list resources?

src/main/java/bio/terra/cli/service/utils/CrlUtils.java
```diff
@@ -37,6 +38,15 @@ public static AIPlatformNotebooksCow createNotebooksCow(GoogleCredentials google
     }
   }
 
+  /** This method is used for testing only. */
+  public static DataprocCow createDataprocCow(GoogleCredentials googleCredentials) {
```
dexamundsen, line 42: if its test-only code. can this be moved to src/test?

src/main/java/bio/terra/cli/service/AxonServerService.java
```diff
@@ -0,0 +1,205 @@
...
+  public static AxonServerService fromContext() {
+    return new AxonServerService(Context.requireUser().getTerraToken(), Context.getServer());
+  }
+
+  /**
+   * Call the Axon Server GET
+   * /api/workspaces/v1/{workspaceId}/resources/{resourceId}/gcp/dataproccluster/metadata:
+   *
+   * @param workspaceId the workspace that contains the resource
+   * @param resourceId the cluster resource id
```
dexamundsen, line 48: Minor: for sake of completeness - please add param (already added) as well as return in the comment docs

applies to all functions

src/main/java/bio/terra/cli/service/AxonServerService.java
```diff
@@ -0,0 +1,205 @@
...
+        "Failed to stop cluster");
+  }
+
+  /**
+   * Utility method that checks if an exception thrown by the Axon Server client is retryable.
+   *
+   * @param ex exception to test
+   * @return true if the exception is retryable
+   */
+  private static boolean isRetryable(Exception ex) {
```
dexamundsen, line 121: outside the scope of this pr - functions following this are similar across services and can be shared code. 
could you please add a backlog ticket under 
https://verily.atlassian.net/browse/BENCH-411

src/main/java/bio/terra/cli/service/WorkspaceManagerServiceGcp.java
```diff
@@ -331,6 +339,51 @@ public GcpAiNotebookInstanceResource createControlledGcpNotebookInstance(
...
+              () -> controlledGcpResourceApi.createDataprocCluster(createRequest, workspaceId),
+              WorkspaceManagerService::isRetryable);
+
+          // Poll the result endpoint until the job is no longer RUNNING.
+          CreatedControlledGcpDataprocClusterResult createResult =
+              HttpUtils.pollWithRetries(
+                  () -> controlledGcpResourceApi.getCreateDataprocClusterResult(workspaceId, jobId),
+                  (result) -> isDone(result.getJobReport()),
+                  WorkspaceManagerService::isRetryable,
+                  // Creating a GCP Dataproc cluster should take less than ~15 minutes.
```
dexamundsen, line 377: 15 mins is a considerable time. Can we return immediately and expect the user to check status later?
same approach is used in AWS notebooks (which can table 5-10 mins as well usually)

src/main/java/bio/terra/cli/service/WorkspaceManagerServiceGcp.java
```diff
@@ -692,6 +745,46 @@ public void deleteControlledGcpNotebookInstance(UUID workspaceId, UUID resourceI
...
+        () -> {
+          // make the initial delete request
+          HttpUtils.callWithRetries(
+              () ->
+                  controlledGcpResourceApi.deleteDataprocCluster(
+                      deleteRequest, workspaceId, resourceId),
+              WorkspaceManagerService::isRetryable);
+
+          // poll the result endpoint until the job is no longer RUNNING
+          DeleteControlledGcpDataprocClusterResult deleteResult =
```
dexamundsen, line 773: How long does a delete take? is it a candidate for async op as well?

src/main/java/bio/terra/cli/businessobject/resource/GcpDataprocCluster.java
```diff
@@ -0,0 +1,192 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Internal representation of a GCP Dataproc cluster workspace resource. Instances of this class are
+ * part of the current context or state.
+ */
+public class GcpDataprocCluster extends Resource {
+  public static final String JUPYTER_LAB_COMPONENT_KEY = "JupyterLab";
+  private static final Logger logger = LoggerFactory.getLogger(GcpDataprocCluster.class);
```
dexamundsen, line 26: nit: logger as the first field

src/main/java/bio/terra/cli/businessobject/resource/GcpDataprocCluster.java
```diff
@@ -0,0 +1,192 @@
...
+  private final ClusterName clusterName;
+
+  /** Deserialize a cluster of the disk format to the internal object. */
+  public GcpDataprocCluster(PDGcpDataprocCluster configFromDisk) {
+    super(configFromDisk);
+    this.resourceType = Type.DATAPROC_CLUSTER;
+    this.clusterName = configFromDisk.clusterName;
+    this.region = configFromDisk.region;
```
dexamundsen, line 34: Region is already part of the base class (resource), please remove

src/main/java/bio/terra/cli/businessobject/resource/GcpDataprocCluster.java
```diff
@@ -0,0 +1,192 @@
...
+  protected void deleteControlled() {
+    // call WSM to delete the resource
+    WorkspaceManagerServiceGcp.fromContext()
+        .deleteControlledGcpDataprocCluster(Context.requireWorkspace().getUuid(), id);
+  }
+
+  /** Retrieve cluster status from axon server */
+  public Optional<ClusterStatus> getClusterStatus() {
+    try {
+      return Optional.of(
```
dexamundsen, line 114: Are these values nullable? Assuming its a no. 

Applies to other get functions below

src/main/java/bio/terra/cli/command/resource/CheckAccess.java
dexamundsen: One option - check support only on create, not on read. This is often used for aws

src/main/java/bio/terra/cli/command/resource/create/GcpDataprocCluster.java
```diff
@@ -0,0 +1,455 @@
...
+
+    // build the dataproc creation parameters
+    CreateGcpDataprocClusterParams.Builder createParams =
+        new CreateGcpDataprocClusterParams.Builder()
+            .resourceFields(createResourceParams.build())
+            .clusterId(clusterId)
+            .region(region)
+            .imageVersion(imageVersion)
+            .initializationActions(initializationActions)
+            .components(List.of("JUPYTER"))
```
dexamundsen, line 170: Should this be the components option? it is currently unused

src/main/java/bio/terra/cli/command/resource/create/GcpDataprocCluster.java
```diff
@@ -0,0 +1,455 @@
...
+      private String type;
+
+      @CommandLine.Option(
+          names = "--manager-accelerator-count",
+          description = "The count of accelerators for the manager.")
+      private int count;
+    }
+
+    static class ManagerDiskConfig {
+
```
dexamundsen, line 292: nit: remove new line

src/main/java/bio/terra/cli/command/resource/create/GcpDataprocCluster.java
```diff
@@ -0,0 +1,455 @@
...
+    private String imageUri;
+
+    @CommandLine.ArgGroup(exclusive = false)
+    private ManagerAcceleratorConfig acceleratorConfig = new ManagerAcceleratorConfig();
+
+    @CommandLine.ArgGroup(exclusive = false)
+    private ManagerDiskConfig diskConfig = new ManagerDiskConfig();
+
+    static class ManagerAcceleratorConfig {
+
```
dexamundsen, line 279: nit: remove new line

src/main/java/bio/terra/cli/serialization/userfacing/resource/UFGcpDataprocCluster.java
```diff
@@ -0,0 +1,179 @@
...
+ * External representation of a workspace GCP Dataproc cluster resource for command input/output.
+ *
+ * <p>This is a POJO class intended for serialization. This JSON format is user-facing.
+ *
+ * <p>See the {@link GcpDataprocCluster} class for a Dataproc cluster's internal representation.
+ */
+@JsonDeserialize(builder = UFGcpDataprocCluster.Builder.class)
+public class UFGcpDataprocCluster extends UFResource {
+  public final String projectId;
+  public final String region;
```
dexamundsen, line 28: region is already part of parent class UFResource, please remove this from entire class

src/test/java/harness/utils/GcpDataprocClusterUtils.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class GcpDataprocClusterUtils {
+  /**
+   * Helper method to poll `terra resources describe` until the cluster state equals that specified.
+   * Uses the current workspace.
+   */
+  public static void pollDescribeForClusterState(String resourceName, String clusterState)
+      throws InterruptedException, JsonProcessingException {
```
dexamundsen, line 20: Why not use ResourceUtils.pollDescribeForResourceField()  ? The timeout can be made into a param if needed. 
Its pretty much the same functionality

src/test/java/unit/GcpDataprocClusterControlled.java
```diff
@@ -0,0 +1,244 @@
...
+/**
+ * Tests for the `terra resource` commands that handle controlled GCP clusters.
+ *
+ * <p>These tests share a single controlled cluster resource created before the tests run.
+ * Additional tests must be added before {@link #listDescribeReflectDelete()}
+ *
+ * <p>This test only executes on verily environments. Add a tag prefix to prevent it from running in
+ * the main repo
+ */
+@Tag("unit-verily-gcp")
```
dexamundsen, line 39: Why do we need a separate tag ?
in Broad the feature flag will return false, so using the same tag 'unit-gcp' would be better

src/test/java/unit/GcpDataprocClusterControlled.java
```diff
@@ -0,0 +1,244 @@
...
+public class GcpDataprocClusterControlled extends SingleWorkspaceUnitGcp {
+
+  private UFGcpDataprocCluster testCluster;
+  private final String name =
+      "cliTestUserDataprocCluster" + UUID.randomUUID().toString().substring(0, 8);
```
dexamundsen, line 44: minor - taking a substring might defeat the uniqueness (?)

src/test/java/unit/GcpDataprocClusterControlled.java
```diff
@@ -0,0 +1,244 @@
...
+ * Additional tests must be added before {@link #listDescribeReflectDelete()}
+ *
+ * <p>This test only executes on verily environments. Add a tag prefix to prevent it from running in
+ * the main repo
+ */
+@Tag("unit-verily-gcp")
+public class GcpDataprocClusterControlled extends SingleWorkspaceUnitGcp {
+
+  private UFGcpDataprocCluster testCluster;
+  private final String name =
```
dexamundsen, line 43: minor - rename 'name' to 'clusterName' - descriptive

src/test/java/unit/GcpDataprocClusterControlled.java
```diff
@@ -0,0 +1,244 @@
...
+      "cliTestUserDataprocCluster" + UUID.randomUUID().toString().substring(0, 8);
+  private final String stagingBucketName =
```
dexamundsen, line 45: minor: use the same random suffix for all three fields

private final String randomUuid8 = UUID.randomUUID().toString().substring(0, 8);
private final String clusterName = "cliTestUserDataprocCluster" + randomUuid8;
private final String stagingBucketName = "staging-bucket-" + randomUuid8;
private final String tempBucketName = "temp-bucket-" + randomUuid8;

src/test/java/unit/GcpDataprocClusterControlled.java
```diff
@@ -0,0 +1,244 @@
...
+    GcpDataprocClusterUtils.pollDescribeForClusterState(name, "RUNNING");
+  }
+
+  @Test
+  @DisplayName("list and describe reflect creating a controlled cluster")
+  void listDescribeReflectCreate() throws IOException {
+    workspaceCreator.login();
+
+    // `terra workspace set --id=$id`
+    TestCommand.runCommandExpectSuccess("workspace", "set", "--id=" + getUserFacingId());
```
dexamundsen, line 103: minor: not needed due to the setUpCluster function. Applies to all tests

src/test/java/unit/GcpDataprocClusterControlled.java
```diff
@@ -0,0 +1,244 @@
...
+    tempBucket =
+        TestCommand.runAndParseCommandExpectSuccess(
+            UFGcsBucket.class,
+            "resource",
+            "create",
+            "gcs-bucket",
+            "--name=" + tempBucketName,
+            "--bucket-name=" + tempBucketName);
+
+    testCluster =
```
dexamundsen, line 82: recommend renaming testCluster to createdCluster

src/test/java/unit/GcpDataprocClusterControlled.java
```diff
@@ -0,0 +1,244 @@
...
+        TestCommand.runAndGetJsonObjectExpectSuccess(
+            "cluster", "launch", "--name=" + name, "--proxy-view=JUPYTER_LAB");
+    String jupyterProxyViewUrl = proxyUrl.getString("JupyterLab");
+    assertNotNull(jupyterProxyViewUrl, "launch cluster jupyterlab proxy url is not null");
+  }
+
+  @Test
+  @DisplayName("list and describe reflect a deleting cluster")
+  void listDescribeReflectDelete() throws IOException {
+    workspaceCreator.login();
```
dexamundsen, line 219: can login be done in beforeAll ?

src/main/java/bio/terra/cli/businessobject/resource/GcpDataprocCluster.java
rogerwangcs: The underlying service class can throw an exception, we intentionally catch them here and then return an empty optional. Use case is `resource list` or `resource describe`, we don't want one of the api calls to cause the whole command to fail. We still wrap the non nullable objs with optional to match return type.

src/test/java/unit/GcpDataprocClusterControlled.java
rogerwangcs: Thats a good point, I feature flagged the test after I made this change, I should be able to remove this separate tag now.

src/main/java/bio/terra/cli/service/WorkspaceManagerServiceGcp.java
rogerwangcs: delete is around the same as vertex notebook delete, I leave this one as sync

src/main/java/bio/terra/cli/command/resource/Delete.java
rogerwangcs: Flag - yes that's possible, we have separate flags for UI and CLI, but they should be synced and toggled together.

dexamundsen (approved):

PR #532, opened 2023-08-21
BENCH-968: Fix failing gcp notebook test

yuhuyoyo:
it stays in the provision state for a while so pull till it becomes active

https://scans.gradle.com/s/wgq4l5xlscdlk/tests/task/:runTestsWithTag/details/unit.GcpNotebookControlled/startStop()?top-execution=1

### DIFF
--- a/src/test/java/unit/GcpNotebookControlled.java
+++ b/src/test/java/unit/GcpNotebookControlled.java
@@ -261,7 +261,6 @@ public class GcpNotebookControlled extends SingleWorkspaceUnitGcp {
 
     // `terra notebook start --name=$name`
     TestCommand.runCommandExpectSuccessWithRetries("notebook", "start", "--name=" + name);
-    GcpNotebookUtils.assertNotebookState(name, "PROVISIONING");
     GcpNotebookUtils.pollDescribeForNotebookState(name, "ACTIVE");
   }
 }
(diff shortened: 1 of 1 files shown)


src/test/java/unit/GcpNotebookControlled.java
```diff
@@ -261,6 +261,7 @@ void startStop() throws IOException, InterruptedException {
 
     // `terra notebook start --name=$name`
     TestCommand.runCommandExpectSuccessWithRetries("notebook", "start", "--name=" + name);
-    GcpNotebookUtils.assertNotebookState(name, "ACTIVE");
+    GcpNotebookUtils.assertNotebookState(name, "PROVISIONING");
```
dexamundsen, line 264: Do we need to assert this? 
Reason for ask - if the notebook is provisioned immediately the state will be active and this might fail. 
Unless we need to verify this state, we can directly poll for active state

dexamundsen (approved):

src/test/java/unit/GcpNotebookControlled.java
mikenev: Agree, this probably doesn't need to be here. But it would also be surprising to get through provisioning almost instantaneously.

yuhuyoyo: sure.

PR #535, opened 2023-08-29
Enable checking email address

yuhuyoyo:
@rogerwangcs might want to consider passing in the user email if we want to have some user to have early access to dataproc in prod.


rogerwangcs:
Is this for supporting identity overrides? I didn't need to make additional change to UI flagsmith logic to get it to pick up email_address segment overrides

yuhuyoyo:
> Is this for supporting identity overrides? I didn't need to make additional change to UI flagsmith logic to get it to pick up email_address segment overrides

ya because I only implemented the trait things in the ui :p

rogerwangcs (approved):

PR #542, opened 2023-09-07
[BENCH-1067]: Add email address to dataproc flag check

rogerwangcs:
Also pass email at flag check in `CommandUtils` instead at the caller level. We should always provide email address for feature flag checks.


rogerwangcs:
@amygdala fyi

dexamundsen (approved):

yuhuyoyo (approved):

PR #545, opened 2023-09-12
[BENCH-1086]: Add terra resource update dataproc-cluster command

rogerwangcs:
Adds resource update command for dataproc clusters.

```
❯ terras resource update dataproc-cluster --help
Missing required option: '--name=<name>'
Usage: terra resource update dataproc-cluster
       [--autoscaling-policy=<autoscalingPolicy>] [--format=<format>]
       [--graceful-decommission-timeout=<gracefulDecommissionTimeout>]
       [--idle-delete-ttl=<idleDeleteTtl>] --name=<name>
       [--new-description=<newDescription>] [--new-name=<newName>]
       [--num-secondary-workers=<numSecondaryWorkers>]
       [--num-workers=<numWorkers>] [--workspace=<id>]
Update a GCP Dataproc cluster.
      --autoscaling-policy=<autoscalingPolicy>
                             Autoscaling policy url to attach to the cluster.
                               Format: projects/[projectId]/locations/
                               [dataproc_region]/autoscalingPolicies/[policy_id]
      --format=<format>      Set the format for printing command output: JSON,
                               TEXT. Defaults to the config format property.
                               Default: null
      --graceful-decommission-timeout=<gracefulDecommissionTimeout>
                             The duration to wait for graceful decommissioning
                               to finish.
      --idle-delete-ttl=<idleDeleteTtl>
                             Time-to-live after which the resource becomes idle
                               and is deleted.
      --name=<name>          Name of the resource, scoped to the workspace.
                               Only use letters, numbers, dashes, and
                               underscores.
      --new-description=<newDescription>
                             New description of the resource.
      --new-name=<newName>   New name of the resource. Only alphanumeric and
                               underscore characters are permitted.
      --num-secondary-workers=<numSecondaryWorkers>
                             The number of secondary worker nodes.
      --num-workers=<numWorkers>
                             The number of worker nodes.
      --workspace=<id>       Workspace id to use for this command only.
```

### DIFF
--- a/src/main/java/bio/terra/cli/command/resource/update/GcpDataprocCluster.java
+++ b/src/main/java/bio/terra/cli/command/resource/update/GcpDataprocCluster.java
@@ -1,0 +1,100 @@
+package bio.terra.cli.command.resource.update;
+
+import bio.terra.cli.businessobject.Context;
+import bio.terra.cli.businessobject.Resource.Type;
+import bio.terra.cli.command.shared.WsmBaseCommand;
+import bio.terra.cli.command.shared.options.Format;
+import bio.terra.cli.command.shared.options.ResourceUpdate;
+import bio.terra.cli.command.shared.options.WorkspaceOverride;
+import bio.terra.cli.exception.UserActionableException;
+import bio.terra.cli.serialization.userfacing.input.UpdateControlledGcpDataprocClusterParams;
+import bio.terra.cli.serialization.userfacing.resource.UFGcpDataprocCluster;
+import picocli.CommandLine;
+
+@CommandLine.Command(
+    name = "dataproc-cluster",
+    description = "Update a GCP Dataproc cluster.",
+    showDefaultValues = true)
+public class GcpDataprocCluster extends WsmBaseCommand {
+  @CommandLine.Mixin ResourceUpdate resourceUpdateOptions;
+  @CommandLine.Mixin WorkspaceOverride workspaceOption;
+  @CommandLine.Mixin Format formatOption;
+
+  @CommandLine.Option(names = "--num-workers", description = "The number of worker nodes.")
+  private Integer numWorkers;
+
+  @CommandLine.Option(
+      names = "--num-secondary-workers",
+      description = "The number of secondary worker nodes.")
+  private Integer numSecondaryWorkers;
+
+  @CommandLine.Option(
+      names = "--autoscaling-policy",
+      description =
+          "Autoscaling policy url to attach to the cluster. Format: projects/[projectId]/locations/[dataproc_region]/autoscalingPolicies/[policy_id]")
+  private String autoscalingPolicy;
+
+  @CommandLine.Option(
+      names = "--graceful-decommission-timeout",
+      description = "The duration to wait for graceful decommissioning to finish.")
+  private String gracefulDecommissionTimeout;
+
+  @CommandLine.Option(
+      names = "--idle-delete-ttl",
+      description = "Time-to-live after which the resource becomes idle and is deleted.")
+  public String idleDeleteTtl;
+
+  /** Print this command's output in text format. */
+  private static void printText(UFGcpDataprocCluster returnValue) {
+    OUT.println("Successfully updated GCP Dataproc cluster.");
+    returnValue.print();
+  }
+
+  /** Update a GCP Dataproc cluster in the workspace. */
+  @Override
+  protected void execute() {
+    workspaceOption.overrideIfSpecified();
+
+    // all update parameters are optional, but make sure at least one is specified
+    if (!resourceUpdateOptions.isDefined()
+        && numWorkers == null
+        && numSecondaryWorkers == null
+        && autoscalingPolicy == null
+        && gracefulDecommissionTimeout == null
+        && idleDeleteTtl == null) {
+      throw new UserActionableException("Specify at least one property to update.");
+    }
+
+    // get the resource and make sure it's the right type
+    bio.terra.cli.businessobject.resource.GcpDataprocCluster resource =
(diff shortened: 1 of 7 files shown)


src/main/java/bio/terra/cli/service/WorkspaceManagerServiceGcp.java
```diff
@@ -558,6 +562,70 @@ public void updateControlledGcpNotebook(
...
+                  new ControlledDataprocClusterUpdateParameters()
+                      .autoscalingPolicy(updateParams.autoscalingPolicyUri));
+      callWithRetries(
+          () ->
+              gcpResourceApi.updateDataprocCluster(
+                  autoscalingPolicyUpdateRequest, workspaceId, resourceId),
+          updateErrMsg);
+    }
+
+    // Update scheduled deletion independently. Dataproc api does not allow autoscaling policy and
```
yuhuyoyo, line 611: Dataproc api does not allow updating idleDeleteTtl and other attributes together

yuhuyoyo (approved):

src/main/java/bio/terra/cli/command/resource/update/GcpDataprocCluster.java
```diff
@@ -0,0 +1,92 @@
...
+        && gracefulDecommissionTimeout == null
+        && idleDeleteTtl == null) {
+      throw new UserActionableException("Specify at least one property to update.");
+    }
+
+    // get the resource and make sure it's the right type
+    bio.terra.cli.businessobject.resource.GcpDataprocCluster resource =
+        Context.requireWorkspace()
+            .getResource(resourceUpdateOptions.resourceNameOption.name)
+            .castToType(Type.DATAPROC_CLUSTER);
```
dexamundsen, line 72: likely not needed here - should we check status or other fields before update attempt?
adding a check on status might be duplicating the logic from WSM though

src/main/java/bio/terra/cli/command/resource/update/GcpDataprocCluster.java
```diff
@@ -0,0 +1,92 @@
...
+        && numSecondaryWorkers == null
+        && autoscalingPolicy == null
+        && gracefulDecommissionTimeout == null
+        && idleDeleteTtl == null) {
+      throw new UserActionableException("Specify at least one property to update.");
+    }
+
+    // get the resource and make sure it's the right type
+    bio.terra.cli.businessobject.resource.GcpDataprocCluster resource =
+        Context.requireWorkspace()
```
dexamundsen, line 70: nit -
extract Context.requireWorkspace() to a local variable after override

src/main/java/bio/terra/cli/service/WorkspaceManagerServiceGcp.java
```diff
@@ -558,6 +562,70 @@ public void updateControlledGcpNotebook(
...
+            .description(updateParams.resourceFields.description);
+
+    updateRequest.updateParameters(
+        new ControlledDataprocClusterUpdateParameters()
+            .numPrimaryWorkers(updateParams.numWorkers)
+            .numSecondaryWorkers(updateParams.numSecondaryWorkers)
+            .gracefulDecommissionTimeout(updateParams.gracefulDecommissionTimeout));
+    callWithRetries(
+        () -> gcpResourceApi.updateDataprocCluster(updateRequest, workspaceId, resourceId),
+        updateErrMsg);
```
dexamundsen, line 594: recommend different, more relevant / specific message during each call to help debugging.

src/main/java/bio/terra/cli/command/resource/update/GcpDataprocCluster.java
```diff
@@ -0,0 +1,92 @@
...
+    returnValue.print();
+  }
+
+  /** Update a GCP Dataproc cluster in the workspace. */
+  @Override
+  protected void execute() {
+    workspaceOption.overrideIfSpecified();
+
+    // all update parameters are optional, but make sure at least one is specified
+    if (!resourceUpdateOptions.isDefined()
```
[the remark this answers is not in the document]
rogerwangcs: I think that would be nicer, but I need at least one between these update parameters AND resource name, description in `resourceUpdateOptions`, but picocli doesn't support mixins inside arg groups. Also this would change the behavior a little where the command `--help` would show the args in a subsection.

src/main/java/bio/terra/cli/service/WorkspaceManagerServiceGcp.java
rogerwangcs: The resulting error includes the more specific wsm api error:

Examples:
`Error updating controlled GCP Dataproc cluster in the workspace: Cluster must be in a RUNNING state to update attributes. Current cluster state is: UPDATING`

`Error updating controlled GCP Dataproc cluster in the workspace: AutoscalingPolicy name 'asd' must conform to .*projects/([^/]+)/(regions|locations)/([^/]+)/autoscalingPolicies/([^/]+) pattern.`

src/main/java/bio/terra/cli/command/resource/update/GcpDataprocCluster.java
rogerwangcs: Good point, UI does this check, CLI should too. Failing in client is much quicker than in the wsm update flight.

rogerwangcs: Keeping the type check to stay consistent with other resource update commands. I assume its also here for a sanity check and immediate error if user accidently passes a non cluster resource.

dexamundsen (approved):
thanks for addressing the comments

